PR #85, opened 2014-12-02
Update for Estonian Land Board WMS address

RM87:
Updated the wms server address for Estonian Land Board server. Added 3 layers (cadastre, hillshading and cir) and removed one layer that was not easily switchable to new server.


pnorman:
I had to revert the merge - the files were invalid and one of them had a strange character, so `make check` wasn't catching it before.

Please open a new PR with a rebased branch that passes `make check`

PR #91, opened 2015-03-04
Add VicMap

stevage:
As suggestede here: https://github.com/systemed/potlatch2/pull/110#issuecomment-76674003


pnorman:
please provide a license_url, and for a WMS source use [`type: wms`](https://github.com/osmlab/editor-imagery-index/blob/gh-pages/schema.json#L13) and indicate [available projections](https://github.com/osmlab/editor-imagery-index/blob/gh-pages/schema.json#L122-L127).

I don't think we're providing any TMS layers which are automatically reprojected WMS ones, so that requires a bit of thought. Is that something we should do by adding it manually, by adding automatically, or leaving up to the editor? JOSM can consume the WMS source directly.

stevage:
Thanks Paul. 

>please provide a license_url

Working on this. It may take a while - the license status is less clear than I thought. Pray to god that it is actually suitable...

>indicate available projections.

A list of EPSG's? Eg, ['EPSG;3857','EPSG:3111'] ?

>I don't think we're providing any TMS layers which are automatically reprojected WMS ones, so that requires a bit of thought. Is that something we should do by adding it manually, by adding automatically, or leaving up to the editor? JOSM can consume the WMS source directly.

Unfortunately, Potlatch2 doesn't, which is the one I care about. (And apparently Potlatch2 only takes imagery sources from here, so I'd be stuck if we couldn't use this method here...)

pnorman:
> A list of EPSG's? Eg, ['EPSG;3857','EPSG:3111'] ?

Yes. See https://github.com/osmlab/editor-imagery-index/blob/4f71262b3a44230c42ff8224792abb74b0a34a50/sources/europe/CzechUHULORTOFOTO.json#L5-L13 for an example.

I'm going to close this, because the license status needs clearing up before anything can be done.

stevage:
FWIW, I'm talking to VicMap about this. The actual licence situation is slightly odd, and sort of the opposite of services like Google:

- the raw data is Creative Commons
- the map image is not open licensed, and requires payment for commercial use

But when I raised the idea of using the map in order to trace the raw data, they were ok with that. Working on getting official wording that we can use.

PR #94, opened 2015-05-09
Add ortophoto source for Szeged, Hungary

jlevente:


pnorman:
fails `make check`

jlevente:
my bad. fixed in c78683e

PR #99, opened 2015-05-20
Added new data source

kolesar-andras:
High resolution image of four districts in Budapest, Hungary. Now it has license_url as well.


kolesar-andras:
Thanks.

PR #103, opened 2015-07-20
2.nd try to update Estonian Land board sources

RM87:
Fixed some identification. 
Created new files, unable to find any strange symbols or anything that would rise a red flag.


RM87:
found what was missing.

PR #104, opened 2015-07-20
3.nd try to update Estonian Land board sources

RM87:
better formatting and missing parentheses added.


pnorman:
Should the polygon be applied to all three sources?

RM87:
I would not apply it to other three because of the following:
CIR, ortho and hillshade are all results of aerial photos that already do cover some of the foreign territory and not jet all domestic small islets. New aerials are made yearly of approx 1/3 of territory meaning these borders may change (extend) every year.
Base map area is fairly static by covering domestic area only.

RM87:
Please merge this.

PR #107, opened 2015-09-24
Adding San Juan County WA street map tile source

npeihl:


bhousel:
Thanks!  I did test and the source does work great as a background in iD.  My only suggestion would be to rename it to something other than "vector tile source", just because "vector tiles" usually means something else.

npeihl:
Duly noted. I've changed the title of the PR. Let me know if I've missed anything else.

pnorman:
> Thanks! I did test and the source does work great as a background in iD

I was having trouble getting tiles, getting 404s for tiles in San Juan

npeihl:
Hi @pnorman,

Can you give me examples of tiles that are 404ing? Z, X and Y would be awesome for troubleshooting.

Thanks.

pnorman:
> Hi @pnorman,
> 
> Can you give me examples of tiles that are 404ing? Z, X and Y would be awesome for troubleshooting.

I had failed to reverse x and y from the normal order. This service uses `z/y/x`, normal is `z/x/y`. It's worth double-checking that they really intended this non-standard order.

sources/north-america/SanJuanCountyWAStreets.json
```diff
@@ -0,0 +1,67 @@
+{
+    "url": "http://sjcgis.org/arcgis/rest/services/Basemaps/General_Basemap_WM/MapServer/tile/{z}/{y}/{x}",
```
pnorman: this should be `{zoom}`, not `{z}`.

npeihl:
Thanks @pnormal. I've fixed the {zoom} parameter. I think I was confused because it's called {z} in the iD Editor. 
2015-09-28 14_53_50-openstreetmap

Also, yes the ```z/y/x``` order is deliberate. This is hosted by Esri's ArcGIS Server software and they have their own way of doing things. [Link to documentation](http://resources.arcgis.com/EN/HELP/REST/APIREF/tile.html)

PR #108, opened 2015-10-29
Update NLS-OSTownPlan-Edinburgh1849-1851.json

RobJN:
Tile URL updated as per https://wiki.openstreetmap.org/w/index.php?title=National_Library_of_Scotland&diff=1226496&oldid=1222548
Tested in JOSM to confirm the change is indeed correct.


RobJN:
I forgot that I had to do a make to get the imagery.xml etc file updated. Closing this and resolving in #109

PR #115, opened 2015-11-26
Add AGIV Imagery (Flanders and Brussels)

sanderd17:
This commit should add AGIV imagery (See https://github.com/openstreetmap/iD/issues/2846 for original ticket)


tyrasd:
merged as https://github.com/osmlab/editor-imagery-index/commit/1a8c1d3eac641cde9a79ae924ce8a4ad9fad208b (where I've left out the [changes](https://github.com/sanderd17/editor-imagery-index/commit/788c515bfb388d48b418f187d7fe68c33962ae82) to the makefile you accidentally commited)

bhousel:
You wanted to add "best: true" for this source, right?

tyrasd:
I've added the "*best*" flag in c3f00d6

sanderd17:
The changes to the makefile were accidental, but AFAICS, it shouldn't break any system. While the current Makefile doesn't work on distros that use Python3 as default s.a. Arch Linux. It took me a while to figure out the script only ran on Python2.

Thanks for adding the "best" flag. I thought this was derived from other data.

Now we'll have less issues with users trying to "fix" buildings and streets by aligning them to Bing imagry

PR #118, opened 2015-12-09
add australia/NSW-LPI sources + parse min/max zoom as integer

andrewharvey:


tyrasd:
thanks!

tyrasd:
Hi @andrewharvey – I noticed that at the moment all of the `LPI_NSW_Administrative_Boundaries` WMS layers don't seem to work: Instead of images, I get results that look like this from the WMS service:

```
<ServiceExceptionReport version="1.3.0"
  xmlns="http://www.opengis.net/ogc"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
  <ServiceException code="LayerNotDefined">
Parameter 'layers' contains unacceptable layer names.
  </ServiceException>
</ServiceExceptionReport>
```

Do you perhaps know if they have been renamed or moved to another server?

PR #121, opened 2016-01-07
Bulk-remove trailing whitespace

pnorman:


pnorman:
diff without whitespace changes (empty): https://github.com/osmlab/editor-imagery-index/compare/gh-pages...pnorman:whitespace?w=1

PR #124, opened 2016-01-07
Remove non-imagery from index

### LINKED ISSUE #15: Public Transport (ÖPNV) is missing bounds

### LINKED ISSUE #66: improve name for "QA No Address"
"QA No Address" is rather inscrutable. @simonpoole what would be a more descriptive name?

### LINKED ISSUE #119: Non-imagery in the list
RIght now we have layers which are purely processed OpenStreetMap data. These layers can't be used in an OpenStreetMap editor to add anything new to OSM, because all the data in them is already in OSM.

What should we do with them?

pnorman:
Fixes #119
Fixes #66
Fixes #15


bhousel:
LGTM :+1:

simonpoole:
Ahemm the layer is quite popular and further there are quite a view non-imagery layers that you missed ... in general the clean up is silly, because the editor maintainers will then simply add them back again manually which seems to make the whole project rather pointless.

pnorman:
It's not about non-imagery, it's about OSM-derived layers.

simonpoole:
Of the 26 global layers 18 are derived from OSM. While theoretically
naturally the information contained in the layers could be re-created
from OSM data, in practical terms providing it via a layer is the
most-cost effective and likely only practical solution.

Am 07.01.2016 um 17:37 schrieb Paul Norman:

tyrasd:
I think @simonpoole has a point here, and a solution like https://github.com/simonpoole/osm-layer-index/commit/5472443667f64b80444f055243baf2d536df24e1 seems to be much more elegant (iD could choose to exclude all or most non-imagery layers while another editor may want to use all of them).

jfirebaugh:
I agree with @tyrasd and @simonpoole -- the intent of this repository was to be an inclusive collection of layers useful to editors, not to be exclusively aerial imagery sources. [This](https://github.com/simonpoole/osm-layer-index) and [this](https://github.com/openstreetmap/iD/blob/aaaf830825bb8d0921c01d5fef8bc32c812664a6/data/update_imagery.js#L9-L33) change moves us back toward editors each having their own custom lists of imagery.

PR #128, opened 2016-01-09
explicitly set a locale when sorting files to prevent large diffs

tyrasd:
Sort order depends on the user's `LC_COLLATE`/`LC_ALL` environment variable. Different sort order results in large git commit diffs in the produced imagery.* files.

Explicitly setting a locale (here: `en_US.UTF-8`) makes the sorted output independent of user environment settings.


pnorman:
I used C instead of en_US, since C should be available everywhere, and we don't care about the ordering so long as it's consistent.

PR #131, opened 2016-01-09
adding imagico.de OSM images for mapping as source under misc

### LINKED ISSUE #112: source definitions for imagico.de osmim layers
I put up source definitions for the [imagico.de OSM images for mapping](http://maps.imagico.de/#osmim) on https://github.com/imagico/osmim-imagery-index.

Included there are files for the individual layers as well as a global file for the aggregate containing all the individual bounding boxes as polygons.

Not sure how this best fits into the structure here so no concrete PR.

imagico:
This adds the [imagico.de OSM images for mapping](http://maps.imagico.de/#osmim) as a single image source under *misc*.  I also replaced the bounding box polygon for Greenland with the actual cutline - the other images much more tightly match their bounding boxes.

The individual sources on https://github.com/imagico/osmim-imagery-index are also updated but would probably unnecessarily complicate things for the vast majority of users - might revisit that if i add more overlapping image sources in the future.

closes #112


pnorman:
I was planning on adding the osmim files to the appropriate directories in sources as their own files.

imagico:
That would be fine with me.

The advantages would be:

* users could select the image they want in areas where different images overlap.  This is currently the case in two places (northern Greenland and German coast).
* the individual source files contain more precise start/end date.
* the detailed description is included (which is otherwise only available in the tile metadata).

The main disadvantage would be that the user has to juggle with the individual image sources.  Note in the tiles served each of the individual sources still contains all of the images available, the layer parameter just affects the priorities of the tiles.

pnorman:
merging as-is for now

PR #134, opened 2016-01-26
Update the README to reflect a more general name of the project

iandees:
The project was never meant to be specific to imagery. I named it as such because that's what the majority of the layers were. Let's use a more general name. This updates the readme to reflect the more general name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,10 @@
-# OSM Editor Imagery Index
+# OSM Editor Layer Index
 
-The goal of this project is to maintain a canonical representation of the imagery
-overlays available to [OpenStreetMap](http://www.openstreetmap.org/) editors such as
-[iD](https://github.com/systemed/iD), [JOSM](http://josm.openstreetmap.de/), and
-[Potlatch 2](https://github.com/systemed/potlatch2).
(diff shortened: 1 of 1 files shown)


iandees:
(I had originally renamed the repo, but reverted that change until this PR gets some discussion and a merge. Either way, Github will redirect both to this repo.)

pnorman:
> The project was never meant to be specific to imagery

I've always considered this project to be for raster layers which provide data external to OSM. Not exclusively imagery, but also road overlays, etc. To me the project has never been about all possible overlays.

nicolas17:
I think imagery layers (like satellite) should be here, external raster layers (like some government's open data) should be here, and OSM-based layers (like QA) should be here. But there should be a field in the JSON saying in what "category" the layer is, to allow filtering if wanted.

jfirebaugh:
Sounds like there's a consensus:

* This repository is not just for aerial imagery layers
* We should revert #124
* Downstream users filtering layers in the way that they want to is a reasonable thing to do
* We can and should add a categorization system to assist them in doing this

tyrasd:
:+1:

PR #146, opened 2016-04-22
Add public domain imagery for San Juan County WA

npeihl:
Aerial imagery was flown in May/June 2013 and released to public
domain. Please note the service is from an Esri ArcGIS Server which
flips the x and y order in the URL. ¯\_(ツ)_/¯


bhousel:
Thanks @npeihl!

PR #147, opened 2016-04-27
New imagery from Luxembourg data.public.lu

grischard:
Thanks!


tyrasd:
Awesome, looks good!

Only minor issue I see is that you're linking to the same `licence_url` page for all imagery sets, even though AFAICS there would be individual pages for each WMTS service (e.g. https://data.public.lu/en/datasets/cartes-topographiques-services-wms-et-wmts/ for the topo map).

grischard:
You're right! I've updated the PR.

PR #154, opened 2016-05-25
Select Geodatastyrelsen as best for Denmark

Hjart:
I spend a lot of time checking edits in Denmark and have many times suggested to new or foreign (to Denmark) mappers that they use Geodatastyrelsen imagery (currently from spring 2015 and from now on updated every year) instead of older Bing imagery. Would be very nice if Geodatastyrelsen was default imagery for Denmark.

### DIFF
--- a/sources/europe/dk/Geodatastyrelsen-Denmark.json
+++ b/sources/europe/dk/Geodatastyrelsen-Denmark.json
@@ -8,6 +8,7 @@
     "type": "tms",
     "name": "Geodatastyrelsen (Denmark)",
     "country_code": "DK",
++    "best": true,
     "extent": {
         "max_zoom": 21,
         "bbox": {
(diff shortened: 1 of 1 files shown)


sources/europe/dk/Geodatastyrelsen-Denmark.json
```diff
@@ -8,6 +8,7 @@
     "type": "tms",
     "name": "Geodatastyrelsen (Denmark)",
     "country_code": "DK",
++    "best": true,
```
tyrasd: the `++` looks like a typo here (would be a syntax error otherwise)

Hjart: You're right. A typo. First time I try this. I think I got it right now.

tyrasd: looks good now, thanks!

PR #160, opened 2016-05-27
new leaflet based overview page incl. live preview

tyrasd:
This extends the gh-pages [overview](http://osmlab.github.com/editor-layer-index/) page:

* larger leaflet-based map with osm.org basemap
* map shows available layers at a coordinate on click
* *best* layers are highlighted in the list and the map
* most layers (i.e. all tms and wms layers that support EPSG:3857) can be previewed live on the map

changes to bundeled output files:
* imagery.xml now contains the *best* flag
* imagery.geojson now contains the *best* flag and the country-code field

A demo is available here: http://tyrasd.github.io/editor-imagery-index/


bhousel:
This is awesome!  Only other thing I would suggest is a link from the README so people can find it.. 🚢

imagico:
Nice but - how should i put it - your interpretation of schema.json w.r.t. polygons is at odds with the one established by practice (which assumes multiple rings means multiple independent polygons).

tyrasd:
> interpretation of schema.json w.r.t. polygons

right, good catch!

PR #162, opened 2016-06-24
Sort the entries and ditch trailing JSON whitespace

iandees:
Sort by some key that should be in the source files so that the concatenation output remains consistent. Also, remove the whitespace that's at the end of the pretty-printed JSON.


bhousel:
I was going to say https://github.com/substack/json-stable-stringify, but then realized that this project uses python!

iandees:
Yea, none of that crazy node stuff here.

iandees:
This is probably ready to go. If the concept seems OK, want to merge?

tyrasd:
looks good to me :+1:

tyrasd:
@iandees: I'm only now realizing that we already had sorting in place (#21) for a long while now :unamused:. Does this new implementation work consistently across computers with different locales (we had issues once, see #128)? Also, we could drop the sorting in https://github.com/osmlab/editor-layer-index/blob/ec59740e3eb420660be1a1eb9f26281c1a03a713/Makefile#L2 now, right?

iandees:
Hmm, yea good point. I could have sworn I saw diffs with ordering problems, but maybe it was just whitespace changes...

Perhaps the Python sorting part of this PR should be removed?

tyrasd:
I've dropped the sorting in the *.py scripts for now. (also, the JSON output in `imagery.geojson` is now also free of trailing whitespace)

PR #163, opened 2016-06-24
updating imagico.de OSM images for mapping

imagico:
This is a simple update of the images in one single layer.  I also created another branch with the individual images as separate layers:

https://github.com/imagico/editor-imagery-index/tree/osmim-individual

Not sure which is the better option.


tyrasd:
I'd actually prefer to merge it as multiple layers, given that there are now even more overlapping layers (see also https://github.com/osmlab/editor-layer-index/pull/131#issuecomment-170330460). Having it as single layers would even open up the option to set the `best` flag for some of them (if and where applicable).

@bhousel @pnorman What do you think?

bhousel:
I think multiple layers probably makes more sense too..  I don't know much about this source though.. Is `max_zoom: 15` right?  That doesn't seem like very useful imagery for tracing.

imagico:
Yes, max_zoom 15 is correct.

Regarding usefulness - most of the imagery is the most recent available in this repository and significant parts are also the highest resolution for the respective areas.  My primary motivation for having this here is because i keep seeing mappers mapping from other older and less accurate images in these areas which seems a fairly wasted effort.

Note the individual images would add quite a lot of layers (about 40).

imagico:
Closing in favor of #164.

Issue #130, opened 2016-01-09
Explore improvements to "best" flag


tyrasd:
The `best` attribute to flag imagery sources that are the _best_ layer to display by default in a given region doesn't always work well – for example when an imagery is only better than another imagery in a fraction of its coverage area (e.g. see https://github.com/osmlab/editor-imagery-index/issues/112#issuecomment-170259349 ff.).

Let's discuss possible improvements here.

imagico:
For this project a useful quality metric in addition to the capture date would be ground resolution.  This of course only can be specified for homogeneous image sources.  But together with the date it would allow applications to perform a better assessment of the image sources.  If for example a source like Bing reaches the limit as you zoom in or has an extremely old date specified in the tile metadata (like the standard *1/1/1999-12/31/2003* for the Geocover basemap) the application could look into the available other sources and suggest to the user to use whatever promises a better quality at the current location.

simonpoole:
@imagico that would not address the issue that quality can vary drastically in the same imagery source even with the same nominal resolution (naturally we don't even record that currently) and in the end quality in this case is only meaningful as a relative measure (not to mention all the issues with zoom levels and so on) and that is why I would prefer a solution as I outline here https://github.com/osmlab/editor-imagery-index/issues/112#issuecomment-170259349

pnorman:
We have an indication of approximate resolution with maxzoom for TMS sources. It's only approximate,  but imagery resolution differences are frequently several powers of 2.

imagico:
Due to the variable scale of the Mercator projection the maximum zoom level does not say much about the resolution.  In the osmim the maximum zoom currently varies between 10 and 14 for example despite all images currently offering either 10m or 15m ground resolution.

pnorman:
> Due to the variable scale of the Mercator projection the maximum zoom level does not say much about the resolution. In the osmim the maximum zoom currently varies between 10 and 14 for example despite all images currently offering either 10m or 15m ground resolution.

For imagery over a reasonably small area, like most, the variable scale doesn't matter as it impacts all imagery the same. A bigger current issue is probably the high number of sources with 22 as a max zoom (~2-3cm where it is) but the imagery is 20-50cm

bhousel:
from:  https://github.com/osmlab/editor-layer-index/pull/164#issuecomment-228856863

[...]

PR #164, opened 2016-06-27
adding OSMIM as individual layers

imagico:
replaces #163.

As individual layers.

Also added best flags to where it obviously fits (Northern Greenland, Alaska, Antarctic and some Subantarctic islands)


bhousel:
I did a quick test of Bing vs imagico on the Alaskan source, and I don't think imagico should be marked `best` here.   Can you please double check all of the areas where you're making the `best` claim?   

bing_v_imagico

imagico:
I am sorry but no, since there is no comprehensive and up-to-date info on the variable quality sources like Bing which quality level is available at what location it is impossible to check this reliably within a reasonable time frame.  Also it can easily change any day.  I added 'best' for images where i know from looking at a representative selection of points in the image they are newest and highest quality (resolution, seasonality, cloud freeness etc.) compared to other sources i know of.  This applies to the Alaska area for a huge percentage.  You can of course be very picky and allow 'best' only for sources that provide best quality by all measures for 100 percent of the coverage area but for a significant portion of the sources marked as best right now there will likely be smaller parts (especially due to clouds) where this does not apply.

Ideally having several objective quality measures (like age and ground resolution) would be better than a binary measure like 'best' - see #130.

bhousel:
Please remove the `best` flag from any sources that you don't have time to check.  

Our hope for this `best` flag is that local contributors can use it to mark imagery that they know to be better than Bing/Mapbox.  For example, "[Kanton Zürich 2015 10cm](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/KantonZuerich10cm-2015-TMS.json#L10)" is marked "best" because people from Switzerland say so, not because of the imagery's age and resolution.  Tomorrow Bing may update their imagery and then "Kanton Zürich 2015 10cm" will no longer be the best source there, and ideally somebody from the Swiss community will notice this and open an issue to remove the flag.

imagico:
OK - removed.

I only added it in the first place since @tyrasd suggested it.  As previously indicated i consider the 'best' flag unsuited for what it is used for.

bhousel:
Thanks, @imagico

PR #170, opened 2016-07-18
updates to NSW LPI layers

andrewharvey:


tyrasd:
looks good, thanks!

Klumbumbus:
Are you sure about the attribution change from 2015 to 2016?
Their website still shows "© Land and Property Information 2015"
e.g. at http://maps.six.nsw.gov.au/arcgis/rest/services/public/NSW_Base_Map/MapServer

andrewharvey:
@Klumbumbus At http://spatialservices.finance.nsw.gov.au/mapping_and_imagery/lpi_web_services/terms_and_conditions they note:

> 3.1.1 DFSI Spatial Services asserts the right to be attributed as author of the original material in the following manner:
> 
> ©Department Finance, Services & Innovation [date of extraction]

So I think the year should be kept current based on that, regardless of the metadata in their arcgis server page.

PS. The attribution on the Contributors wiki page is slightly different covered by https://wiki.openstreetmap.org/wiki/Attribution/New_South_Wales_Government_Data.

Klumbumbus:
OK, so I'll update it to 2017

PR #174, opened 2016-08-15
Updated source URL and polygon for the Canton Solothurn

simonpoole:
From https://github.com/simonpoole/osm-layer-index/commit/24704d1860c7b0f320025a4466bcf24966afa7a9


bhousel:
Thanks @simonpoole!

PR #177, opened 2016-08-20
update assets

tyrCron:
(This is an automated message and pull request.) Looks like the last commit didn't include the output of a proper `run make` to regenerate the combined sources in XML, JSON and GeoJSON formar. Let's fix that\!


tyrasd:
:tada: my cron script is actually working. Do you think it's a good idea to keep it running?

bhousel:
It's great for now!  I think it would be great to eventually move this to a CI system like Travis or Circle, but not really a priority for me..

PR #180, opened 2016-08-23
Add Bavaria 80cm imagery as in osm-layer-index

simonpoole:


simonpoole:
Noe that I didn't set the "best" flag as it is not clear if the imagery in absolute terms is the best for the whole covered area (I assume DG/bing/MB are updating at least the Munich area more often and have better nominal resolution).

bhousel:
Thanks!  can you run `make` also?

geowOSM:
>Note that I didn't set the "best" flag as it is not clear if the imagery in absolute terms is the best for the whole covered area 
Thanks, that's definitely right.

geowOSM:
I would like to suggest set the "best" flag, as these official DOPs are by far the best in Bavaria and will get updated on a 2-year-cycle from 2018. Thanks for considering.

tyrasd:
@geowOSM done

simonpoole:
Is the imagery -really- better than bing/mapbox/DG/ESRI? It may be more recent, but the difference in nonimal resolution is rarther large.

geowOSM:
In theory nominal resolution is better, but bing/mapbox/DG/ESRI are often not correctly georeferenced and are not ortho-rectified. Especially for the latest bing imagery a significant offset by several meters is not rare. I regularly have to encourage new mappers to use Bavaria DOP 80 as they redraw existing highways buildings due to bing. Bavaria DOP 80 is referenced and rectified using a DGM with 1 m spatial resolution.

Only for the Munich area bing has significant better quality (resolution 10-20 cm). Maybe someone has a footprint for this area to mark it as best.

PR #181, opened 2016-08-23
Fix Europe

simonpoole:
Essentially everything in the europe directory was either broken or in the wrong place.


simonpoole:
This should be rebased.

bhousel:
> This should be rebased.

I can help with the conflicts..  Is everything else about it good to merge?

bhousel:
I think I got it all - merged these as:
f6023f0 OpenPT map is an overlay
136563f Cleaned up europe directory, deleted stuff that doesn't exist anymore

PR #185, opened 2016-09-02
add layer with vector data from IBGE to DF-Brazil

willemarcel:
I produced a layer using data from IBGE on top of Mapbox Satellite to help users to map addresses on Distrito Federal, Brazil. The data from IBGE is released in Public Domain. I published a blog post about it https://www.openstreetmap.org/user/wille/diary/39238 (Portuguese only, sorry).

### DIFF
--- a/imagery.geojson
+++ b/imagery.geojson
@@ -10593,6 +10593,47 @@
         },
         {
             "geometry": {
+                "type": "Polygon",
+                "coordinates": [
+                    [
+                        [
+                            7.492,
+                            47.4817
+                        ],
+                        [
+                            7.492,
+                            47.6342
+                        ],
+                        [
+                            7.784,
+                            47.6342
+                        ],
+                        [
+                            7.784,
+                            47.4817
+                        ],
+                        [
+                            7.492,
+                            47.4817
+                        ]
+                    ]
+                ]
+            },
+            "type": "Feature",
+            "properties": {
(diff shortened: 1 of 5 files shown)


tyrasd:
Thanks Wille!
But why did you combine this dataset with the mapbox satellite imagery? What if someone wants to use it with a different background (e.g. better aerials, an osm layer, etc.)?

willemarcel:
I was thinking about iD users... Since we can not combine more than one layer in the iD editor, it would be more useful to have the data combined with the satellite imagery. It's really a good idea to have a layer only with the vector data. I can make it later. Thanks for the suggestion!

bhousel:
> I was thinking about iD users... Since we can not combine more than one layer in the iD editor, it would be more useful to have the data combined with the satellite imagery. It's really a good idea to have a layer only with the vector data. I can make it later. Thanks for the suggestion!

Thanks @willemarcel, the more correct way to do this is to render the data to transparent tiles and set the `'overlay'` flag to `true`.  See the [TIGER Roads Overlay](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/north-america/us/TIGER2014RoadsOverlay.json#L5) as example.

Multiple 'overlay' sources can be displayed in iD at the same time.  They have toggleable checkboxes in the Background pane:

screenshot 2016-09-06 10 33 20

willemarcel:
@bhousel That's great! I'll update the layer in Mapbox Studio and the pull request.

willemarcel:
@bhousel @tyrasd I just updated the layer and set it as overlay. Thanks for the help!

PR #187, opened 2016-09-10
add new aerial imagery: Luftbild Erlangen 2016

poempelfox:
License for the imagery is normally CC-BY 4.0, but with an extra exception for OSM use (listing on the OSM contributors list satisfies the attribution requirement for OSM).
The extent polygon was created by user klumbumbus for JOSM and according to his forum post is free for other editors to use.


bhousel:
Thanks @poempelfox!

PR #189, opened 2016-09-14
Fix CoCT 2015 imagery boundry out-of-order

Firefishy:


bhousel:
Thanks for the quick fix!

PR #191, opened 2016-09-21
Switch to a GeoJSON-based source files

### LINKED ISSUE #190: Question: Why not just geojson everything?
Question for @tmcw, @jfirebaugh following #188/#189
Is there a reason this index doesn't just use GeoJSON for everything?

GeoJSON has geometry and custom properties built into it, can be previewed directly in GitHub and constructed trivially with geojson.io, and has a wealth of tools that work with it, and can be validated with JSON schema.

All of this wasn't necessarily true a few years ago when this project started - but it seems like an obvious choice today.

We currently have a bunch of .json files containing geojson-like polygons (but not really, because holes) that we concatenate together with a python script.   This is weird and I'd like to know if there's a good reason for it before I go changing things around.

My somewhat side goal for all the projects that I currently touch is to reduce as much development friction as possible, automate repetitive tasks, and rely on bots and CI tools to make less work for me where it makes sense to do so.

This issue is really a more pointed restatement of #168, so I'm going to close that one.

iandees:
- [x] Update jsonschema
- [x] Reformat existing sources
- [x] Adjust conversion scripts to use GeoJSON format

Fixes #190.


iandees:
@bhousel what do you think?

sources/africa/cv/osmim-imagicode-LC82100502015347LGN00.json
```diff
@@ -1,26 +1,47 @@
...
-    "country_code": "CV",
-    "extent": {
-        "min_zoom": 0,
+    "type": "Feature",
```
tyrasd: It's a bit uncommon to have the type attribute not as the first property of the geojson (usually the order is `type`, `properties`, `geometry`). This could be a bit confusing.

iandees:
I probably need to adjust the documentation to reflect this new format.

sources/africa/cv/osmim-imagicode-LC82100502015347LGN00.json
iandees: Yea, this happened because by default object properties are unordered in JSON, so technically it doesn't matter. Ordering in this way can help keep future PR diffs smaller, though, so I'll make that change.

iandees:
(Thanks for the offline review, @bhousel!)

iandees:
@tyrasd I'm going to incorporate your suggestions in another PR that isn't as difficult to review as this one.

bhousel:
This looks awesome, thanks @iandees 
Agree w everything @tyrasd said too!

simonpoole:
It seems as if the net result of this PR is not backwards compatible in any form. While you could argue that maintaining both the (multi-)polygon and the bbox arguments in the source file and output was unnecessary duplication simply dropping it makes life difficult for anybody that was not using the polygons.

Which brings us back full circle to the question if the maintainers of the repo are interested in providing a stable product for any- and everybody to use, or if it is in reality iD only.

iandees:
The output formats should be backwards compatible. If they aren't, then we should fix that. Can you file a new ticket with examples, @simonpoole?

PR #194, opened 2016-10-11
Adding french aerial imagery and french style openstreetmap

sidjy:
According to http://openstreetmap.fr/bdortho, french national institute of geographic and forest information (IGN) authorizes openstreetmap to use its aeral imagery


sources/europe/fr/FR-BAN.geojson
```diff
@@ -10,6 +10,7 @@
         "url": "http://{switch:a,b,c}.layers.openstreetmap.fr/bano/{zoom}/{x}/{y}.png",
         "max_zoom": 20,
         "min_zoom": 12,
+	"overlay": true,
```
tyrasd: very minor nitpick: can you change the indention here from tab to spaces?

sources/europe/fr/FR-IGN-Ortho.geojson
```diff
@@ -0,0 +1,336 @@
+{
+    "type": "Feature",
+    "properties": {
+        "attribution": {
+            "url": "http://openstreetmap.fr/bdortho",
+            "text": "BDOrtho IGN",
+            "required": true
+        },
+        "description": "French IGN aeral imagery",
+        "url": "http://proxy-ign.openstreetmap.fr/bdortho/{zoom}/{x}/{y}.jpg",
```
tyrasd: This URL only gives me HTTP/403 errors, for example when trying to access http://proxy-ign.openstreetmap.fr/bdortho/12/2103/1461.jpg – is that just a temporary hickup or has the URL been changed?

tyrasd: Is the tile server also available via `https`?

sources/world/OpenStreetMap-FrenchStyle.geojson
```diff
@@ -0,0 +1,16 @@
+{
+    "type": "Feature",
+    "properties": {
+        "url": "http://{switch:a,b,c}.tile.openstreetmap.fr/osmfr/{zoom}/{x}/{y}.png",
+        "attribution": {
+            "url": "http://openstreetmap.org/",
+            "text": "© OpenStreetMap contributors, CC-BY-SA",
+            "required": false
```
tyrasd: Why is it not required? :confused:

sources/europe/fr/FR-BAN.geojson
tyrasd: (PS: there's two or three more places in other files of this PR where indention uses tabs instead of spaces.)

sources/europe/fr/FR-IGN-Ortho.geojson
Robou: @tyrasd yes it seems that it can be accessed both via `http` and `https`.
They are talking about this on [this openstreetmapFrance thread](http://forum.openstreetmap.fr/viewtopic.php?f=5&t=4715&start=30).

tyrasd:
related ticket: https://github.com/openstreetmap/iD/issues/3420

sources/europe/fr/FR-IGN-Ortho.geojson
tyrasd: I see. There's  also a ticket on the iD repository, apparently: https://github.com/openstreetmap/iD/issues/3197

tyrasd:
Ohh, I see that the 403's [above](https://github.com/osmlab/editor-layer-index/pull/194#discussion_r84573839) are caused by a server-side [user-agent or http-referrer check](http://forum.openstreetmap.fr/viewtopic.php?f=5&t=4715&start=15#p13593) that basically only allows tiles to be accessed by JOSM or iD. That's a bit unfortunate, since this repository is intended for more than just the iD editor (other OSM editors should be able to use the layers as well).
//cc @cquest

sources/world/OpenStreetMap-FrenchStyle.geojson
sidjy: because I did like the GermanStyle....but you're right, it should be required, I just changed it

althio:
@tyrasd Maybe not the ideal solution ultimately, but this is the best @cquest can do right now.
(JOSM then iD mappers are waiting for this in France.)

I guess server-side check can be added for other OSM editors, as OpenStreetMap France and IGN want to go for a white-list approach.

pnorman:
👎 if it requires specific user-agents or referers.

althio:
@pnorman 
> :-1: if it requires specific user-agents or referers.

I would like more details here, any explanation or guidelines about this? Is it technical, legal, ...?

To expose a bit more the context and current situation:
 - there is an agreement for imagery use with "upstream requirements", i.e. OSM editing-only see agreement/convention http://openstreetmap.fr/bdortho http://openstreetmap.fr/f/2016.05.20_OSM-FR_IGN_Ortho.pdf
 - we discussed options to control access (mentions for tokens, cookies, referers, proxies, ... https://github.com/openstreetmap/iD/issues/3197 https://github.com/openstreetmap/iD/issues/3420)
 - imagery is already accessible for JOSM users (proxy, TMS, referer) with custom link
 - imagery is already accessible for iD users with custom link
 - this PR wants to give access automatically, without the need for custom link

Any concerns/suggestions welcome.

Robou:
Hi, Can't we split this Pull Request in two parts ?

1. one for adding osmFR map rendering : "OpenStreetMap (French Style)"
1. one for the BDorthoIGN imagery automatically accessible on iD
Because the first one seems to me easier and does not cause any issue. Sory if this is irrevelant, I am pretty noob and maybe I didn't catch the true meaning of the lines of code in the PR.

bhousel:
> Hi, Can't we split this Pull Request into two parts/PR ?

There is no need to do that:
1.  iD won't be showing "French Style" mapnik tiles.  You can certainly include them in the index, but we would blacklist that source [like we do with other local osm tilesets](https://github.com/openstreetmap/iD/blob/718c02c61805b470b1f1907c0eef82bbcf3f62b8/data/update_imagery.js#L12-L15).
2. iD can't display imagery from a server that returns 403 based on a useragent/referrer check.  Again, you can include it in the index if you think it's useful to some other editor, but we'd just blacklist it on the iD side.

Neither part of this PR will make it into iD.

pnorman:
> I would like more details here, any explanation or guidelines about this? Is it technical, legal, ...?

This index is for more than just JOSM and iD.

althio:
@cquest
according to https://github.com/openstreetmap/iD/issues/3420#issuecomment-255536396 or https://github.com/openstreetmap/iD/issues/3197#issuecomment-228749029 one way to go is `access_token` or `key`, possibly with public token / public key
> general key provided for OSM editors, and therefore requires no registration. However, the accesses with this key are limited (like https://wiki.openstreetmap.org/wiki/WikiProject_Austria/geoimage.at#JOSM )

`"url": "http://gis.lebensministerium.at/wmsgw/?key=4d80de696cd562a63ce463a58a61488d&FORMAT=image/jpeg&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&Layers=Luftbild_MR,Luftbild_1m,Luftbild_8m,Satellitenbild_30m&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}"`

althio:
@bhousel 
> 1. iD won't be showing "French Style" mapnik tiles. You can certainly include them in the index, but we would blacklist that source like we do with other local osm tilesets.

Fair enough, potential access in other editors is ok.

> 2. iD can't display imagery from a server that returns 403 based on a useragent/referrer check. Again, you can include it in the index if you think it's useful to some other editor, but we'd just blacklist it on the iD side.

Well iD can, it works with custom link 
http(s)://proxy-ign.openstreetmap.fr/bdortho/{z}/{x}/{y}.jpg
So why the blacklist?

bhousel:
> Well iD can, it works with custom link 
> http(s)://proxy-ign.openstreetmap.fr/bdortho/{z}/{x}/{y}.jpg
> So why the blacklist?

It does not work:

screenshot 2016-11-03 09 56 31

althio:
Hmmm yes apparently still not right over `http` but works with `https`.

pnorman:
> Hmmm yes apparently still not right over http but works with https.

I get a 403 over https too

image

cquest:
You'll have access granted when using iD on openstreetmap.org web site.
If we can set a separate instance of iD with this layer in the preset, it may be another way to offer iD editing with this aerial imagery without having to register it in the global list.
Is this something that can be configured in iD code somewhere ? (never looked at it)

pnorman:
> You'll have access granted when using iD on openstreetmap.org web site.

Ya, that's going to make it unsuitable for here.

> If we can set a separate instance of iD with this layer in the preset, it may be another way to offer iD editing with this aerial imagery without having to register it in the global list.

You'll need to change the layers part of the iD build process which shouldn't be too hard. If you have difficulty you're best off raising an issue on the iD tracker.

bhousel:
> If we can set a separate instance of iD with this layer in the preset, it may be another way to offer iD editing with this aerial imagery without having to register it in the global list.
> Is this something that can be configured in iD code somewhere ? (never looked at it)

I agree with @pnorman that this source does not belong in the editor-layer-index.

If you want to run your own fork of iD that includes it, you can add the source to [a whitelist in `update_imagery.js`](https://github.com/openstreetmap/iD/blob/master/data/update_imagery.js#L42).

At this point, you're talking about forking the OSM tools just to support one imagery source with onerous usage requirements.  My strong recommendation is _still_ to work with the imagery provider to actually open up their imagery in a way that lets everyone use it.

tyrasd:
> My strong recommendation is still to work with the imagery provider to actually open up their imagery in a way that lets everyone use it.

:+1:

PS: @sidjy: feel free to open a new pull request that only adds the French osm tiles.

althio:
@bhousel 
> I agree with @pnorman that this source does not belong in the editor-layer-index.

Well, we have a source, not fully open, but granting the right to use for OSM editing purposes.
It seemed natural to add it in the common repo for OSM editors.

I understand that you all ( @tyrasd @bhousel @pnorman ) oppose the current implementation (which tries to provide extra steps to ensure it is only used for OSM editors, but this is not done quite right (yet?)).
However I don't understand why you think the source itself does not belong here.

> My strong recommendation is still to work with the imagery provider to actually open up their imagery in a way that lets everyone use it.

It took some years and effort from several OSM members in the French community to get there with the national mapping agency. It is a significant first step for them and for us. Not an argument, just a comment: I dare say it will not give a very good signal to open up their data if we make little use of the current agreement.

Please be sure that our final goal is free geo data and promoting actual open data. Only, with external partners, sometimes we just have to learn to walk before we can run.

@cquest 
> If we can set a separate instance of iD with this layer in the preset, it may be another way [...]

@bhousel ;
> If you want to run your own fork of iD that includes it, you can add [...]
> At this point, you're talking about forking the OSM tools just to support one imagery source with onerous usage requirements.

I cannot speak for @cquest, but I think he is just trying to find proposals/solutions, and the easiest if possible.
If every other way is blocked, he is standing and looking at the forking route. Not sure he will walk the path, just an option to look at. Anyway, thanks for the pointers.


To reiterate: currently iD and JOSM users already have access, only with custom URL. We are just trying to make it easier and transparent for beginner contributors. We would also like the imagery to be available for other OSM editors.

bhousel:
> I understand that you all ( @tyrasd @bhousel @pnorman ) oppose the current implementation (which tries to provide extra steps to ensure it is only used for OSM editors, but this is not done quite right (yet?)).
> However I don't understand why you think the source itself does not belong here.

None of us have been able to get the source to work in testing.  I think (speaking for myself) if proxy-ign returned images, I would have no issue with including it in the index.

Even if the proxy returned images that all said "visit http://whatever.fr to accept the terms", and then the user does that step and the proxy starts serving images, I'd be ok with that too.

Dominik-K:
@bhousel Did you also try the TLS connection? [Your screenshot](https://github.com/osmlab/editor-layer-index/pull/194#issuecomment-258149231) just shows the HTTP test. In the [OSM France blog](http://openstreetmap.fr/bdortho), they explicitly mention the `https` channel for `iD`.

It worked fine for me with `iD` on the official `openstreetmap.org` website using `https://proxy-ign.openstreetmap.fr/bdortho/{zoom}/{x}/{y}.jpg` as a custom background layer.

@pnorman Did you try it on the OSM website, as well?

pnorman:
> @pnorman Did you try it on the OSM website, as well?

No - nor do I feel the need to. osm.org is not the only referrer used for people accessing imagery that is in this index.

bhousel:
> @bhousel Did you also try the TLS connection? Your screenshot just shows the HTTP test. In the OSM France blog, they explicitly mention the https channel for iD.

At the time, I only tested with the `http` source string that was submitted in the PR.

It looks like even with `https` I get 403 Forbidden:

screenshot 2017-04-06 14 26 16

I'm running iD locally though, not on openstreetmap.org, so that's probably why.

grischard:
Please test #418 for a fix that adds a token to use the French IGN imagery in ID.

PR #197, opened 2016-11-01
Updated imagery for Stadt Bern

simonpoole:


bhousel:
Thanks @simonpoole - I was able to load a few tiles, but most of them are erroring with 500 internal server error.  Is there something wrong with your mapproxy instance?  let me know if you'd like me to try this again later..

 screenshot 2016-11-03 21 58 06

simonpoole:
Seemed to be hickup with the mapproxy instance, the WMS itself seems to
be rather slow.

I would suggest including it regardless, and I take it up with the folks
in Bern when I get back from Berlin.

Simon

bhousel:
> I would suggest including it regardless, and I take it up with the folks
> in Bern when I get back from Berlin.

sounds good to me.. thanks!

tyrasd:
@simonpoole the WMS service here seems to be down at the moment. For example, [this request](https://map.bern.ch/wms/OpenData/proxy.php?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&FORMAT=image%2Fjpeg&TRANSPARENT=true&LAYERS=0&WIDTH=256&HEIGHT=256&SRS=EPSG%3A4326&STYLES=&BBOX=7.3828125%2C46.93359375%2C7.40478515625%2C46.95556640625) just returns a HTTP/400 error. The mapproxy.osm.ch based TMS source still returns some tiles, but that might be due to some local caching on that side.

simonpoole:
https://map.bern.ch/wms/OpenData/proxy.php?request=GetCapabilities works for me.... (given that the place is not that large our mapproxy instance likely has a cache of the full coverage area in any case).

tyrasd:
aah, it looks like the layer names have been changed, it's not called `0` anymore but `luftbild_2016`, then it loads fine:



Probably this has to be updated on the mapproxy instance also – do you know who to contact for that? (btw: if I remember correctly, mapproxy.osm.ch misses at least some high-ish zoom level tiles in the south-west of the city)

simonpoole:
No idea who to contact .... but somehow it has magically been fixed as you wrote those words :-)

Klumbumbus:
It is https now too.

grischard:
The mapproxy is https now too. I'll update them.

Issue #168, opened 2016-07-16
Just make this JavaScript already


bhousel:
Does anyone have any serious objections if I replace the (very few) Python scripts in this repository with JavaScript?  It would make things so much easier with testing, publishing, dependency management, etc?

JavaScript ate the world already, lets just use it.  This whole repo is JSON files anyway.

iandees:
Heh, maybe it ate *your* world... My world is still squarely snake-like.

Yup, switching to node is fine. Let me know if I can help.

simon04:
@bhousel: is there a (public) branch regarding your script-to-JavaScript migration?

bhousel:
> is there a (public) branch regarding your script-to-JavaScript migration?

No, I haven't started.. I've been focused mainly on iD v2, which I'd like to release in time for osmgeoweek.  

It really helps me out a lot to be able to include your improvements to editor-layer-index.  👍   Just wanted to say thank you!

rbuffat:
ELI evolved to have more advanced CI checks implemented with Python. Imho Python is an adequate language for this use case as it is heavily used for data analysis/data science and has excellent support for spatial data processing. That being said, I'm more experienced with Python than Javascript and there is this saying that if you only have a hammer everything looks like a nail. 

Probably most things could also be implemented with Javascript, but at least one thing I'm aware of that there is currently no good library available for coordinate transformations (Proj4js seems to have troubles with some projections).

As this issue is quite old I'm closing it for now.

bhousel:
I have no idea why a project that smooshes together a few hundred json files needs to perform projections, but ok.

rbuffat:
> I have no idea why a project that smooshes together a few hundred json files needs to perform projections, but ok.

That's the reason, it's now not just smooshing a few hundred JSON files together anymore.

E.g. there is a bot now that weekly checks all WMS server-based sources, which checks advertised and not advertised projections. This is done by downloading test images for each projection (which requires projections) and comparing them using ImageHashes.  If you are curious: https://github.com/osmlab/editor-layer-index/blob/gh-pages/scripts/sync_wms.py

PR #201, opened 2016-11-09
Scripts require Python 2

simon04:
This allows running `make` on Arch Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,20 @@
 ALL = imagery.geojson imagery.json imagery.xml
 SOURCES = $(shell find sources -type f -name '*.geojson' | LC_ALL="C" sort)
+PYTHON = python2
 
 all: $(ALL)
 
 check:
-	@python scripts/check.py $(SOURCES)
+	@$(PYTHON) scripts/check.py $(SOURCES)
 
 clean:
 	rm $(ALL)
 
 imagery.xml: $(SOURCES)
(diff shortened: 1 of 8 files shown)


iandees:
I'd rather see us fix the code so that it runs on both python2 and python3.

Can you detail what problems you had running it on python3?

simon04:
```
$ python3 scripts/check.py 
  File "scripts/check.py", line 11
    print file
             ^
SyntaxError: Missing parentheses in call to 'print'
```

```
$ python3 scripts/check.py
Traceback (most recent call last):
  File "scripts/check.py", line 4, in <module>
    schema = json.load(open('schema.json', 'rb'))
  File "/usr/lib/python3.5/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```

…

See my PR #202.

bhousel:
These issues are exactly why we should move the project to node.js.  see #168 
I don't have time to deal with python dependency hell and version mismatches.

iandees:
@bhousel good thing you don't have to! Each language has its own dependency hell and version mismatches.

bhousel:
Sorry for my frustration before, we should just merge these improvements..

PR #204, opened 2016-11-10
update hike bike map

Klumbumbus:
wmflabs url is newer. hikebikemap.org also uses this. see also https://wiki.openstreetmap.org/wiki/Hike_%26_Bike_Map

updated also attribution

synchronized with josm


tyrasd:
thanks!

PR #207, opened 2016-11-11
Update OSMIMultipolygon.geojson

Klumbumbus:
multipolygon changed to area
fix max zoom and add in zoom
synced with josm


bhousel:
Looks good to me, thanks!

PR #209, opened 2016-11-11
+maxzoom Bavaria-2m

Klumbumbus:


simonpoole:
Is stating a max zoom actually necessary for a WMS server? If this is a JOSM requirement we need to have a look at all WMS defs.

Klumbumbus:
I don't know. With this layer and without maxzoom I see grey tiles in JOSM. While e.g. VoGIS: Echtfarbenbild 2012 (12cm) works fine in JOSM even without maxzoom.

bhousel:
Max zoom is at least useful to iD - it tells the code at which zoom to stop requesting images and switch to overzooming.  (e.g. there is no reason to ask the server for 4 z19 tiles, when we know that the z18 tile we already have loaded is as good as it gets).

I know iD don't support wms yet, but even if we did, fetching raster tiles would work the same way.

simonpoole:
@bhousel I might be mistaken but IMHO WMS servers return which scales they support so this might simply be a misconfiguration of the WMS server in question.

bhousel:
> @bhousel I might be mistaken but IMHO WMS servers return which scales they support so this might simply be a misconfiguration of the WMS server in question.

I'm thinking of the very simplest case where iD would send bounding box requests to EPSG:3857 supporting servers, per https://github.com/openstreetmap/iD/issues/1141#issuecomment-241298005

Specifying a `maxZoom` would help iD here.  I'm not planning to do `GetCapabilities` calls anytime soon.

simon04:
Merged: 576198ed33d4334676aaad484479fe0a68c6002b

PR #212, opened 2016-11-11
Add id to all sources

simon04:


bhousel:
This is great!  I've wanted this for a while.. I have a few questions, related to how iD uses the `id` property:

* Have you checked to see if the `id`s are unique?   They _probably_ are, but would be nice to run a test to prove it. 
* Can we ensure that all the `id`s are suitable for use as url parameters?  i.e. no [reserved characters](https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters) like `&`, `#`, etc.

iandees:
Also, can we move it out of the `properties` object and onto the `Feature` object? GeoJSON [has an `id` attribute on the feature](https://tools.ietf.org/html/rfc7946#page-11) and this would be a perfect spot for it.

tyrasd:
> but would be nice to run a test to prove it

yep, that would be nice to have for the automatic PR tests.

> Also, can we move it out of the properties object and onto the Feature object? GeoJSON has an id attribute on the feature and this would be a perfect spot for it.

+1

simon04:
* unique id: check added in b0572b4
* id as URL param: characters restricted in a0ed468
* GeoJSON id attribute on the feature → #216 (since it is a breaking change)

Klumbumbus:
Please add the id attribute also to the XML output.

PR #221, opened 2016-11-14
Mark BenincotonouPleiade2016 as best

sommerluk:
Mark BenincotonouPleiade2016 as best image available at the polygon that it covers.

This is my first PR here; I’m following the documentation but couldn’t test the change. So it should be sensibly reviewed.


bhousel:
Thanks @sommerluk, it looks good to me!

PR #223, opened 2016-11-14
Update KantonSolothurnInfrarot125cm-SOGIS2011.geojson

Klumbumbus:


simonpoole:
The file should be renamed ,,,,, SOGIS2014.geojson or whatever.

PR #225, opened 2016-11-14
Add City of Cape Town 2013 imagery

Firefishy:


bhousel:
Thanks Grant!

Klumbumbus:
Is `"start_date": "2013" "end_date": "2016"` correct? Sounds unlikely since there is an own entry for 2015.

Firefishy:
Well spotted. `end_date` should be `2014`.

Klumbumbus:
Were images actually taken in 2014 or is `end_date: "2014"` only due to the date definition at https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md ?

Firefishy:
Yes, potentially. The majority of the image was collected in Summer 2013 (note Southern Hemisphere) and was process in 2014. If there were gaps, they may have been captured in early 2014.

Klumbumbus:
OK, thanks. I'll fix it.

PR #229, opened 2016-11-15
Add some processing options to convert_geojson_to_legacyjson

simonpoole:
This adds

- an option to generate bounding boxes from the polygons (as the original boxes got zapped in the conversion to geojson)
- an option to remove the polygons (if you don't need them it doesn't really make sense to have to parse them)
- and option to ignore WMS sources

### DIFF
--- a/scripts/convert_geojson_to_legacyjson.py
+++ b/scripts/convert_geojson_to_legacyjson.py
@@ -8,6 +8,27 @@ def convert_json_source(source):
     polygon_coords = geometry.get('coordinates') or []
     if polygon_coords:
         extent_obj['polygon'] = polygon_coords
+        # generate bbox from polygon coordinates as a stop gap         
+        min_lon = 180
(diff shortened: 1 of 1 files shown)


scripts/convert_geojson_to_legacyjson.py
```diff
@@ -27,8 +54,28 @@ def convert_json_source(source):
 
     return converted
 
+try:
+    opts, args = getopt.getopt(sys.argv[1:], "btr",["bbox","tmsonly","removepolygons"])
```
iandees: Can we use Python's [`argparse` library](https://docs.python.org/3/library/argparse.html) instead?

simonpoole:
Removed after 4 weeks due to lack of interest in upstream repo.

iandees:
Or you could just say "Hey, I haven't seen any work here" instead of being passive aggressive?

PR #238, opened 2016-11-17
improve ORT10LT source

ramunasd:


ramunasd:
This source without any doubts is best in Lithuania. I've added additional flag `best`.

Issue #217, opened 2016-11-13
Support GeoJSON MultiPolygon


simon04:
@tyrasd in https://github.com/osmlab/editor-layer-index/pull/206#issuecomment-260005802:
> Well, actually, there are already a [couple](https://gist.github.com/tyrasd/048ec49ed52ed37ba56a25fcab198c2f) of "multipolygon" sources in the repo (for example the *South Tyrol 2014*, or the *Tiris &ast;* imagery), but which we forgot about in the geojsonification process in #191. Instead, they continued to use the slightly odd (and incompatible to GeoJSON) convention where the rings in the *coordinates* array are all outer rings and inner rings are not allowed.
>
> It's probably best to fix those all in one go and to apply an additional fix (very similar to [yours](https://github.com/osmlab/editor-layer-index/pull/206/commits/b7dca247bc643af20b3c0c261d0a12db07bfcffb#diff-e83c497604fe1a8262599e63cb85ea09L71) also to fix the [legacy json output](https://github.com/osmlab/editor-layer-index/blob/gh-pages/scripts/convert_geojson_to_legacyjson.py)).
>
> For this particular layer this means that we can either wait until this is all fixed in the data processing before adding the source, or that we merge it using the old&incorrect coordinate scheme (because the PR as-is will probably break things for users of our legacy-json output).

simon04:
Change in `convert_xml.py` to support MultiPolygon:
```diff
diff --git a/scripts/convert_xml.py b/scripts/convert_xml.py
index f2a0b34..47ea57a 100644
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -68,7 +68,7 @@ for source in sources:
     if geometry:
         bounds = ET.SubElement(entry, "bounds")
 
-        for ring in geometry['coordinates']:
+        for ring in geometry['coordinates'] if geometry['type'] != 'MultiPolygon' else [ring for rings in geometry['coordinates'] for ring in rings]:
             shape = ET.SubElement(bounds, "shape")
             for p in ring:
                 point = ET.SubElement(shape, "point")
```

rbuffat:
@andrewharvey @bhousel 
MultiPolygons would be handy for many sources. Is there a reason that MultiPolygons are not allowed? Are there editors that would break? It looks as some sources use interior rings of Polygons to "emulate" MultiPolygons, which leads to invalid geometries.

bhousel:
[...]

PR #241, opened 2016-11-23
Update SanJuanCountyWA2013Imagery.geojson

Klumbumbus:
improve shape, sync with josm


simon04:
MultiPolygons are not yet supported, see #217.

Klumbumbus:
I changed it to Polygon for now like all the other entries do it.
Once #217 is done this can be changed back.

PR #242, opened 2016-11-23
Update PhiladelphiaImagery2015.geojson

Klumbumbus:
better shape, sync with josm


simon04:
Fixed indentation and merged manually. Thanks.

PR #250, opened 2016-11-30
delete FR CRAIG layers (Le Puy, Montluon, Moulins, Vichy)

Klumbumbus:
they don't work anymore.
see also
- https://wiki.openstreetmap.org/w/index.php?title=WikiProject_France/CRAIG&diff=next&oldid=1142126
- https://josm.openstreetmap.de/wiki/Maps/France?sfp_email=&sfph_mail=&action=diff&version=25&old_version=24&sfp_email=&sfph_mail=


bhousel:
cool, thanks!  are you going to add the 2013 source?

Klumbumbus:
I just compared 2010 and 2013 Auvergne. These are the same pictures at z1-14 and 17-20. So only at z15-16 there is a difference in some areas. But thats not important. 2013 also has more covering in the south. So I'll not add the 2013 source but replace the 2010 source with the 2013 source.

bhousel:
For what it's worth iD only allows tracing at z16 and higher, so that's kind of my threshold for whether an imagery source really makes sense for this index.  It sounds like upgrading to the 2013 source is worthwhile.

PR #254, opened 2016-12-10
Delete Budapest-Interspect-2014.geojson

tacsipacsi:
http://geoserver.infobex.hu/ seems to be dead; it's irritating that I have to change to Bing every time I edit something as Interspect shows a black background.


simonpoole:
"seems to be" shouldn't you check with the provider is something has changed or at least verify that the service is not being offered any more?

tacsipacsi:
I tried to check it, but I found nothing on http://interspect.hu/ (attribution URL). Maybe @kolesar-andras can tell us more as he added it in 35b8e719d47033d84e7340652b6ba8194de6b044.

kolesar-andras:
Server is down, you may delete this data source. I have asked provider to restart service, they promised but there are technical difficulties that may take unpredictable amount of time. I will add this data source again when it will start working.

tacsipacsi:
Thanks, I hope it will work soon. It looks much better than any other data source (based on their website).

Klumbumbus:
@kolesar-andras could you please have a look at https://josm.openstreetmap.de/ticket/14086 There is another non working imagery entry from budapest.

PR #255, opened 2016-12-13
Create TW NLSC EMAP layers.

chihchun:
Add two layers from Taiwan National Land Surveying and Mapping Center, which are license under Taiwan Government Open Data license[1].

[1] http://data.gov.tw/license#eng

PR #256, opened 2016-12-13
TW NLSC: Fix URL.

chihchun:
s/z/zoom/


tyrasd:
already fixed via 4ce978c948dcc92fe4900db9faf1a0891c9fbca1, see https://github.com/osmlab/editor-layer-index/pull/255#issuecomment-266824952

PR #265, opened 2016-12-30
Delete NLS-OS6Inch1888-1913.geojson

Klumbumbus:
This source no longer works because it requires now an subscription api key. More information see http://maps.nls.uk/projects/subscription-api/
>The Subscription API layers are more detailed, and have involved more work to create. Subscriptions are necessary to fund these additional costs, as well as the costs of hosting, licensing, and 24/7 availability.

An access would cost now 311,05 € per year. see http://sites.fastspring.com/klokantech/product/osgb10k1888

So this seems no longer available for osm.

Thanks to user blackadder for giving information on this topic.


Klumbumbus:
Merged as nobody raised doubts.

PR #267, opened 2016-12-30
Delete Budapest-XII-ortophoto-2013.geojson

Klumbumbus:
This source doesn't deliver tiles for a while.
Maybe @kolesar-andras knows more as his name is part of the URL.


Klumbumbus:
Merged as nobody raised doubts.

PR #269, opened 2017-01-05
imagico.de OSMIM update 2017-01

imagico:
This updates the osmim layers to the current state and also provides more precise footprints.

See also http://maps.imagico.de/#osmim


Klumbumbus:
What does the part with the channels in the descriptions mean? e.g. `channels 234 (true color)`

imagico:
That are the numbers of the spectral bands in the source imagery that are mapped to RGB.

Band numbering systems can be found here:

https://pbs.twimg.com/media/C6AXwkmU0AEct5P.jpg

Usually i use 234 for Landsat8 and Sentinel-2 which approximates natural colors, hence the `true color`.

Klumbumbus:
OK, thanks for the explanation.

Klumbumbus:
There are some sources with `channels 12x`. Is the "x" a typo?

Also I think it makes sense to seperate the channels for easier understanding, like e.g. `channels 2,3,4 (true color)` or similar

imagico:
`12x` is for ASTER images with an estimated blue band (based on empirical relationships between bands under common circumstances since ASTER has no native blue band).

I admit this is all kind of cryptic and probably of little use for most users.  I have this info in my metadata mostly as a reference for me.  If you prefer to remove it or reformat it in some way i can adjust the script processing the data.

Klumbumbus:
The description is presented to the end user in iD and JOSM (I don't know about the other editors). So I think it should not contain "cryptic" information.
So I would reduce 
`S2A_OPER_PRD_MSIL1C_PDMC_20160813T183324_R094_V20160812T105622_20160812T105622, 2016-08-12, channels 234 (true color), Up-to-date late summer imagery with few clouds`
to
`Up-to-date late summer imagery with few clouds (true color).`

Klumbumbus:
Is there a typo "Cental Alps in late September 2016" instead of "Cent_**r**_al Alps in late September 2016"?

imagico:
Yes, thanks for pointing out.

PR #274, opened 2017-01-11
Canton Basel-Landschaft 2015 10 cm imagery

simonpoole:
The original WMS server this is proxied from seems to have an issue with lower zoom levels, 18 for min-zoom seems to work.


Klumbumbus:
This layer seems to be the same as https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/KantonBaselStadt-2015-TMS.geojson

(I noticed a slightly different color postprocessing and a little offset, but the source material and the bounds are the same.)

Klumbumbus:
Ah sorry, the shape is not the same, but the shape from this layer (Basel-Landschaft) is wrong.

simonpoole:
@Klumbumbus why do you believe that the shape is wrong?

Klumbumbus:
It delivers tiles for e.g. Adelhausen http://www.openstreetmap.org/node/74268455 10km outside of the shape od nearly the whole part of Basel-Stadt

simonpoole:
Sure, it is very likely that the imagery is from the same flight as the imagery for Basel Stadt from 2015 (though this is higher resolution) which went a substantial way in to Germany. But while there is AFAIK no competing imagery in Germany or France, we have more recent imagery for the Kanton Aargau (mid 2016 which will be made available rsn) that is "better", and imagery for the Canton Solothurn which is from the same time frame, so it made sense to use the "official" canton boundary as a border.

This is a general problem with the "best" flag that it applies to the whole coverage polygon.

Issue #151, opened 2016-05-07
NDOP and USGS National Map


bhousel:
NDOP supports several statewide aerial photography and orthoimagery programs.
This page has a list of several statewide programs
http://www.ndop.gov/data.html

USGS National Map looks interesting too:
http://nationalmap.gov/ortho.html#data

bhousel:
Update on USGS National Map:  https://www.usgs.gov/news/usgs-national-map-orthoimagery-map-services-transition-and-other-map-service-changes

See discussion here:
https://lists.openstreetmap.org/pipermail/talk-us/2016-December/016887.html

I think this means we need to change something in the imagery index, but I really can't follow the thread.
@iandees can you open an issue for what needs to change?

iandees:
My understanding is that the "large scale imagery" layer as we know it is gone and won't be replaced. I'm trying to find someone who can give me more information.

iandees:
It looks like [the USGS National Map imagery-only layer](https://basemap.nationalmap.gov/arcgis/rest/services/USGSImageryOnly/MapServer) has pre-tiled NAIP and higher-res imagery down to z19. The server doesn't render anything with better resolution even when it's clear that the imagery supports it (I was looking in Chicago, for example). The description on that layer says to use [the USGSImageOnlyLarge](http://services.nationalmap.gov/arcgis/rest/services/USGSImageOnlyLarge/MapServer) layer for requests beyond z19, but I can't find anywhere in the US where that service gives back better imagery.

So the confusion for me is what happened to the higher-res imagery. I wouldn't have set the USGS Large Scale imagery layer to go to zoom 22 without a reason 😄 .

bhousel:
Yeah I'm kind of confused too.  NDOP seems to be gone now, but my remembering of it was that:
* All the individual states are supposed to have their own orthoimagery programs
* There was some minimum resolution that they were supposed to support by some date
* NAIP was somewhere around 5m resolution (which is not great for tracing) but getting better (1m now?)
* Some of the states already well exceed this minimum (e.g. NJ, MD, which we've added this index)
* USGS was going to merge all the state imagery into a single national map (is that what `NAIP-Plus` layer is?)

I might be misremembering some of it..

iandees:
Hi everyone listening to this ticket: I discovered that this layer will still work, it's just that the server I'm caching from starts at z16. I've adjusted the tile server so that it should return a 404 if you request a tile < z16 instead of proxying the white images they send us.

[...]

PR #290, opened 2017-03-06
Use Caltopo's scanned topographic maps

iandees:
Related to #151.

Matt Jacobs from Caltopo gave permission for us to use their scanned topographic maps layer as the USGS removed the server that OSM US was caching/tiling.

### DIFF
--- a/sources/north-america/us/USGSTopographicMaps.geojson
+++ b/sources/north-america/us/USGSTopographicMaps.geojson
@@ -2,10 +2,15 @@
     "type": "Feature",
     "properties": {
         "id": "USGS-Scanned_Topographic",
-        "url": "http://{switch:a,b,c}.tile.openstreetmap.us/usgs_scanned_topos/{zoom}/{x}/{y}.png",
+        "url": "http://s3-us-west-1.amazonaws.com/caltopo/topo/{zoom}/{x}/{y}.png",
         "type": "tms",
         "name": "USGS Topographic Maps",
-        "country_code": "US"
+        "country_code": "US",
+        "attribution": {
+            "url": "https://caltopo.com/",
+            "text": "© Caltopo",
+            "required": true
+        }
     },
     "geometry": {
         "type": "Polygon",


Klumbumbus:
Is the permission documented somewhere?

iandees:
@Klumbumbus I have permission in my email inbox from Matt. I've asked Matt to comment here rather than copy/paste the conversation.

mtjacobs:
Permission given.  Would prefer that you use http://caltopo.s3.amazonaws.com/topo/ instead.  It's the same bucket but the caltopo.s3 domain name lets me add a robots.txt file that tells search engines not to try to mass-spider the tiles.

iandees:
Thanks, @mtjacobs. I'll switch the URL now.

Klumbumbus:
>Permission given
@mtjacobs: The permission counts for JOSM and other editors too?

mtjacobs:
Yes.  If it's not too much work I'd ask desktop apps to set either the referrer or user-agent headers to something that identifies them.  Didn't bring it up earlier because it happens automatically with web requests.

If it requires an intrusive code change then don't worry about it.

Klumbumbus:
Thanks. I updated the JOSM entry.
The JOSM user-agent should be e.g. `JOSM/1.5 (11732 de) Java/1.8.0_121`
However this can easily be changed for this imagery entry if you wish.

mtjacobs:
I'm fine with that, no need to change.

Klumbumbus:
The shape of the imagery is a bit different than what we had before with the old source. Do you already have a shape polygon of your imagery?
bereich

mtjacobs:
No

talllguy:
Currently seeing a 403 when trying to access via JOSM. Try [this tile](http://caltopo.s3.amazonaws.com/topo/18/75259/99910.png).

image

iandees:
It looks like the max zoom should be z16.

talllguy:
OK.

@Klumbumbus can you make modify the JOSM entry to include the max zoom restriction?

Klumbumbus:
Can you give the exact location? For me JOSM automatically detects the max zoom, enlarges the tiles when zooming further in and doesn't display a 403 error.

(In general it is risky to set the maxzoom explicit, as we maybe don't notice when the imagery supplier increases the maxzoom and then we don't present the best tiles to the user. However particulary for this source it seems safe to set it as an even higher zoom level is not to be expected.)

talllguy:
@Klumbumbus Thanks for the notification about that zoom enlarge feature. I realize now that it only works if JOSM is able to find some tiles to load (at a lower z#) first. See below.

[Here is the location](https://www.openstreetmap.org/#map=18/39.29331/-76.647808) where I was getting the error.

iandees:
@Klumbumbus this source is unlikely to add any extra zoom levels as the tiles are all pre-rendered and the source maps that were scanned won't be getting re-scanned anytime soon.

Klumbumbus:
I added the maxzoom to josms list and created a ticket for the 403 error https://josm.openstreetmap.de/ticket/14562

JesseCrocker:
It might be good to check with @mtjacobs if the recent announcements from caltop effect permission for OSM editors to use this layer.

iandees:
To be clear, the "recent announcements" in question: https://blog.caltopo.com/2020/09/02/ending-layer-licensing/

PR #299, opened 2017-03-30
Add 2016 aerial imagery from San Juan County WA

nickpeihl:
Please note, the {x} and {y} parameters in the URL are switched
because Esri map services are just that way.   🤷‍


tyrasd:
Awesome, thanks! :+1:

PR #300, opened 2017-04-01
Czechia - IPR ortofoto

zbycz:
Hi,

I am adding Prague ortophoto imagery from IPR, who agreed to derive data for OSM.

see https://wiki.openstreetmap.org/wiki/Cs:WikiProjekt_%C4%8Cesko/freemap#Ortofotomapy_IPR_Praha

Thanks!

Issue #211, opened 2016-11-11
use GeoJSON linter to test geojson input


tyrasd:
Additionally to checking against our custom json schema, we should use a GeoJSON linter (like [geojsonhint](https://github.com/mapbox/geojsonhint), optimally one that supports [RFC 7946 GeoJSON](https://tools.ietf.org/html/rfc7946)) to check for basic GeoJSON compatibility of the source layers (e.g.validity of geometries, winding order of polygon rings, etc.).

Maybe this could be done as part of js-ification #168?

iandees:
For what it's worth, the jsonschema check does a pretty good job of verifying the GeoJSON source files.

bhousel:
> Maybe this could be done as part of js-ification #168?

yes 💯

tyrasd:
> For what it's worth, the jsonschema check does a pretty good job of verifying the GeoJSON source files.

As far as I can see, it can't find stuff like unclosed polygon rings, wrong winding order of rings, self-intersecting geometries or mismatched inner/outer rings, can it?

iandees:
> As far as I can see, it can't find stuff like unclosed polygon rings, wrong winding order of rings, self-intersecting geometries or mismatched inner/outer rings, can it?

That's true. I guess I'm hoping that whoever generates the GeoJSON for their submission would generate valid GeoJSON, but it's no problem to add the geojsonlint check so let's do it.

grischard:
Maybe we can use https://github.com/JasonSanford/geojsonlint.com/blob/master/geojsonlint/utils.py

rbuffat:
We currently do not use a geojson conform encoding of MultiPolygons. There is https://github.com/osmlab/editor-layer-index/pull/857 to fix this, but it is stale for quite a while. With the changes in this PR the validity of geometries are tested with the shapely python library (which uses geos) https://shapely.readthedocs.io/en/stable/manual.html#diagnostics

rbuffat:
With https://github.com/osmlab/editor-layer-index/pull/1450 there is now a validation for geometries. Detected errors include e.g..: 

```
geometry is not valid: Hole lies outside shell[120.8268 15.3658]
```

```
geometry is not valid: Self-intersection[-70.5619471801745 -4.17766449833781]
```

In combination with the jsonschema checks, there is now a nearly complete check for GeoJSON validity. Thus I'm closing this issue.

PR #306, opened 2017-04-14
Update IPR-orotofoto

Klumbumbus:
improve min/max zoom and shape, sync with josm, see #300


tyrasd:
Hmm, here you broke the geojson polygon geometries. :cry:

Should be fixed in 463a759d1c0579e3705f3e0672656c9a4903b32d.

(see also #211)

Klumbumbus:
Sorry.
When I created this PR, the github geojson map preview didn't work at this time (I checked with another geojson source), so this was some kind of unfortunate circumstances that I didn't detect this error before merging :)

PR #302, opened 2017-04-07
24 Brazilian WMS layers

Skippern:

PR #307, opened 2017-04-14
Updates

Skippern:
Closing comments from @Klumbumbus on #302 and adding IBGE-BC250 highways


Klumbumbus:
IBGE-BC250-Trecho Rodoviario is also free to use in OSM?

Skippern:
All IBGE data are released free to import. Same also goes for INPE and a few other government agencies.

Aun Johnsen

Klumbumbus:
OK, thanks.

PR #308, opened 2017-04-17
2016 AGIS imagery

simonpoole:
A bit late for reasons beyond our control, but here it is.

### DIFF
--- a/imagery.json
+++ b/imagery.json
@@ -17749,10 +17749,10 @@
             "required": false,
             "text": "AGIS OF2014"
         },
-        "best": true,
+        "best": false,
         "country_code": "CH",
         "extent": {
-            "max_zoom": 19,
+            "max_zoom": 20,
             "min_zoom": 8,
             "polygon": [
                 [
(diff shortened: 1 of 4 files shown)


tyrasd:
Thanks Simon!

Klumbumbus:
You changed max zoom from 19 to 20. However it seems z20 are only enlarged z19 tiles. I think it would make sense to change it back to z19 to avoid unecessary traffic!?

simonpoole:
The original geo-tifs were tiled for zoom level 20 the other zoom levels were generated from them, so: no zoom 20 is not simply an overlooked version of zoom 19.

On 19. April 2017 18:33:30 GMT+01:00, Klumbumbus <notifications@github.com> wrote:
>You changed max zoom from 19 to 20. However it seems z20 are only
>enlarged z19 tiles. I think it would make sense to change it back to
>z19 to avoid unecessary traffic!?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/osmlab/editor-layer-index/pull/308#issuecomment-295358305

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit Kaiten Mail gesendet.

simonpoole:
Back from the UK: just to clarify the tiling at level 20 is for hysterical reasons :-). From a resolution pov at the latitude the Canton Aargau is 19 is sufficient. When I get around to it I'll likely re-tile both the 2014 and 2016 imagery at level 19. Currently I'm running that for 2016 NIR channel first.

Klumbumbus:
(see https://github.com/osmlab/editor-layer-index/commit/08e962ff1cac480d0e78a34c58d08b0eca4f574f for zoom change)

PR #309, opened 2017-04-18
imagico.de OSMIM update 2017-02

imagico:
adding newest images.


tyrasd:
Thanks Christoph!

Klumbumbus:
Enderby Land and Kemp Coast `"start_date": "2013",` is that correct? (Descricption says only 2017)

imagico:
No, that's a bug in my script.  Fixed in #310.

grischard:
For future PRs, you don't need to run `make` anymore. Travis does it automagically, and it avoids likely merge conflicts.

imagico:
@tyrasd - probably not.  I can't say it off hand because the larger high latitude image layers (Greenland, Svalbard) have a dynamic maximum zoom level, i.e. the maximum zoom level available depends on latitude.  In case of Svalbard the resolution threshold is 7m - if z13 has a pixel size of less than 7m at the southern end of the image no z13 tiles are generated.

tyrasd:
Ok, thanks. I noticed the latitude-dependent maxzoom on the greenland layer – it was confusing my script to test the maxzoom values of the tms layers in here quite a bit :sweat_smile: .

PR #311, opened 2017-04-19
Added MultiPolygon support in convert_xml

BathoryPeter:
I've tried to add an orthophoto consists of multiple pieces [1]. Created the geojson, but make failed, because of XML converter only supports Polygon geojson features.

This commit adds the MultiPolygon support.

[1] https://josm.openstreetmap.de/wiki/Maps/Hungary#SsktPusztazmorTrnokDisdorthophoto2017

### DIFF
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -76,20 +76,25 @@ def add_source(source):
     if geometry:
         def coord_str(coord):
             return "{0:.6f}".format(coord)
+
+        isMultiPolygon = geometry['type'] == 'MultiPolygon'
+        polygons = geometry['coordinates'] if isMultiPolygon else [geometry['coordinates']]
+
         bounds = ET.SubElement(entry, "bounds")
-        lons = [p[0] for ring in geometry['coordinates'] for p in ring]
-        lats = [p[1] for ring in geometry['coordinates'] for p in ring]
+        lons = [p[0] for polygon in polygons for ring in polygon for p in ring]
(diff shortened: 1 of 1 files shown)


tyrasd:
see #217

BathoryPeter:
Ok, I will add that othophoto with an invalid Polygon (outer inner ring).

But what is next step to support MuliPolygons? Should we write some legacy converter, or talk to iD editor team or something else?

Btw we should make a note in CONTRIBUTING.md to warn contributors for this problem.

bhousel:
Just seeing this @BathoryPeter and @tyrasd. I'm ok with switching the editor-layer-index sources over to proper multipolygons - it actually makes things a little easier on the iD side.

grischard:
@systemed @simonpoole would Potlatch and Vespucci be able to digest multipolygons? They still use the legacy format, correct?

systemed:
P2 uses the plain old json format, yes, which doesn't appear to have any support for inner rings, just multiple polygons. It seems like such a niche thing that I'd suggest just not writing inner rings to the plain old json and P2 users can live without holes.

grischard:
@BathoryPeter would it please be possible to make sure the legacy converter doesn't choke on this, add the documentation, and ping me so I can merge this? Thanks!

grischard:
If this ever happens, it should now be PR'd to the `master` branch.

rbuffat:
There is a follow-up PR for the same issue https://github.com/osmlab/editor-layer-index/pull/857
I'm closing this as there seems to be no progress for a long time.  If somebody objects I'm happy to open this PR again.

PR #320, opened 2017-05-03
updated HU/Sóskút ortophoto 2017 URL (WMS -> TMS)

BathoryPeter:
The WMS service is still exists but we have an own hosted mirror with Mercator TMS tiles.


tyrasd:
looks good, thanks!

Klumbumbus:
min/max-zoom is 11/20, which could be added too

PR #322, opened 2017-05-07
shortening description strings of OSMIM entries

imagico:
As requested in #269.

### DIFF
--- a/imagery.json
+++ b/imagery.json
@@ -580,7 +580,7 @@
             "url": "http://maps.imagico.de/#osmim"
         },
         "country_code": "BJ",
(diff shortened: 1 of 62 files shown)


Klumbumbus:
"true color" got lost in a few cases when it was not written in backets before. E.g. in `Icebridge DMS image of the settlement - alignment might be poor`

imagico:
Yes, that's collateral losses for the moment.  I consider better separating the verbal description from the technical specifications in my metadata layout but that is something for another day.

Klumbumbus:
alright

PR #323, opened 2017-05-09
Add DigitalGlobe-Premium and DigitalGlobe-Standard worldwide imagery layers

bhousel:
I'm very excited to be able to merge 2 new worldwide imagery layers into the editor-layer-index today!
These layers are made available by DigitalGlobe specifically for OpenStreetMap use.

* **DigitalGlobe-Premium** is a mosaic composed of DigitalGlobe basemap with select regions filled with _+Vivid_ or custom area of interest imagery, 50cm resolution or better, and refreshed more frequently with ongoing updates
* **DigitalGlobe-Standard** is a curated set of imagery covering 86% of the earth's landmass, with 30-60cm or resolution where available, backfilled by Landsat.  Average age is 2.31 years, with some areas updated 2x year.

These new sources are a great opportunity for OpenStreetMap, as they cover many parts of the world that previously had no usable imagery!   Thanks, DigitalGlobe & partners!  ❤️  🛰


iandees:
👏

imagico:
Since there has not been an official announcement yet apparently i suppose license details are to come.  The attribution link currently provided to here: https://wiki.openstreetmap.org/wiki/DigitalGlobe refers to the Mapbox Satellite layer which should probably be clearly distinguished from these new image sources - even if they are also served through Mapbox infrastructure.

Given the rather generic names of the layers it might be a good idea to add a description field detailing the differences (i.e. why should a mapper want to use 'Standard' and not 'Premium').

kevinbullock:
Official announcement from DG: http://blog.digitalglobe.com/news/digitalglobe-satellite-imagery-launch-for-openstreetmap/ 

And I've updated the wiki with the EULA: https://wiki.openstreetmap.org/wiki/DigitalGlobe 

I have a specification for Standard vs Premium, will post, just need to determine best location for the spec. 

Thanks everyone!

PR #324, opened 2017-05-09
Add orthophoto layers from Tyrol/Austria

simon04:


sources/europe/at/TirisOrthofotoCIR.geojson
```diff
@@ -0,0 +1,954 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "tirol.gv.at-orthofoto",
```
tyrasd: could you adjust this id so that it's unique (e.g. `tirol.gv.at-orthofoto-cir`)?

sources/europe/at/TirisOrthofotoCIR.geojson
```diff
@@ -0,0 +1,954 @@
...
+        "attribution": {
+            "text": "tiris.tirol.gv.at",
+            "required": false
+        },
+        "name": "Tiris: orthophoto CIR",
```
tyrasd: I'm not sure everyone knows what *CIR* stands for. Maybe just call it "infrared" or so would be better?

simon04:
Thanks for your comments. I addressed them and amended the PR.

imagico:
Since we just had the discussion regarding spectral band specifications with satellite images in #269 it should be noted that probably all _CIR_ aerial images are green-red-nir images with red vegetation and dark blue water while many false color infrared satellite images (not only here but especially also what you can see elsewhere, for example on https://landsatlook.usgs.gov/viewer.html or what you can see on openaerialmap) are red-nir-swir with bright green vegetation.  It might make sense to allow for an additional attribute specifying that so applications could display specific instructions on interpreting imagery.

PR #333, opened 2017-05-15
Add start/end date to AGIS imagery, reduce max zoom and add polygon for 2011

simonpoole:


tyrasd:
would you like to add the start_date/end_date information to the *AGIS2016* as well?

simonpoole:
@tyrasd well the 2016 has the best flag set so it doesn't really make a difference at this point in time.

tyrasd:
I don't see what the best flag should have to do with the date metadata field, but ok. :man_shrugging:

simonpoole:
Sorting. Assuming younger is better and best is best for an area, adding the dates doesn't make a difference in sorting. I was intending to add them in the future when we remove the best flag, but I can do it it now if you want :-).

tyrasd:
Ah, I see. I've added the dates now in 8a843367d2736baa4bc0a943f8575a252af7fac4, so no one can forget about it later. ;)

PR #334, opened 2017-05-19
add LINZ LDS Basemaps

### LINKED ISSUE #70: Add LINZ (Land information new zealand) imagery layers
News item here:
http://www.scoop.co.nz/stories/PA1404/S00416/aerial-imagery-now-online-for-reuse.htm

imagery and licences here (CC):
https://data.linz.govt.nz/x/vuozYk

There are lots of layers, I think 42. I'm not entirely conversant with json, but if someone could explain the format clearly, I should be able to take part in making the json files myself.

andrewharvey:
Closes #70.

* Uses a dedicated API Key
* http://www.linz.govt.nz/data/linz-data-service/lds-basemaps
* https://wiki.openstreetmap.org/wiki/Contributors#LINZ
* https://wiki.openstreetmap.org/wiki/File:OSM_waiver_-_LINZ.pdf

I've followed https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#building-the-combined-files however since some previous commit's didn't run `make` there are changes from those previous commits in this commit.


Klumbumbus:
Is the api key limited in any way? Can it be used in JOSM too?

andrewharvey:
@Klumbumbus Reviewing the TOS at https://data.linz.govt.nz/terms-of-use/

Assuming attribution is met and requests are as a result of direct user actions (which they are in iD and JOSM) then I see no issue with using this in JOSM.

> 3.2.4.	You shall not use the LDS APIs in a manner that adversely impacts the stability of the LDS service or adversely impacts other users of the LDS APIs. Furthermore, LINZ reserves the right to rate limit or block applications that make a large number of calls to the APIs that are not primarily in response to direct user actions.

> 5.2.1.	Comply with attribution requirements as given in respective metadata and/or respective licences; and

> 5.2.2.	Apply LINZ data in accordance with the restrictions in respective metadata and/or licence such as retraction of personal data, no-use for navigation of SOLAS vessels or defining rights of access.

Klumbumbus:
Ok. Thanks.

Klumbumbus:
I thought that the key is application specific

andrewharvey:
> I thought that the key is application specific

The API key used here was a new API in my account on LDS for the editor-layer-index repository. I don't see any reason not to also use it in the JOSM imagery list. At any time LINZ could revoke access, :crossed_fingers: the api key isn't abused and is only used in the OSM editors for contributing to OSM.

PR #335, opened 2017-05-24
Update url of Lithuania-ORT10LT

ramunasd:


Klumbumbus:
The "old" URL still works. Will it keep working in the future or will it be shut down soon? Do you have more information about the URL change?

ramunasd:
@Klumbumbus actually it is same server and mapproxy instance, it's just some rearrangements in url schemes. Old url will keep working for some time, until it will be completely safe to disable it. And even then, we gonna leave redirect to new url.

Issue #17, opened 2013-08-27
i18n


jfirebaugh:
We should use Transifex for translating imagery names and descriptions. Follow what iD does for presets:
- Assign each source an id
- `make` builds an imagery.yml source for Transifex
- `make translations` pulls down translations into locale directory, split into separate files
- Downstream consumers merge translations into their own

tyrasd:
Is this approach really practical for this use case? There's quite a large (and ever growing) amount of layers by now and most of them only have local or regional importance. While it certainly makes sense to translate the description of the aerial imagery layer a city like Brussels into the locally used languages (here: at least French and Dutch), it may not be necessary to translate it into every other language as well.

What if we'd just let the contributors of imagery add multiple descriptions in the locally relevant languages? Idk, but why not just use OSM's *multilingual tags* pattern here as well, e.g. `"description:fr"`, etc.?

bhousel:
I almost closed this as stale, but I think that:

- Each source should have a unique id 
- It would be useful to localize the globally relevant source names and attribution text
- It's totally ok to not translate something in Transifex that is not relevant

tyrasd:
>  It's totally ok to not translate something in Transifex that is not relevant

yeah, but that's a bit hard to convey that in transifex's UI (with all it's 0-100% progress bars and rankings that suggest one should translate each and every string).

manfredbrandl:
I would like to translate the interface strings and not the layer names. Is it possible to separate these into extra transifex resources? Then the interface part could be added as an extra resource to the iD-editor-team.

manfredbrandl:
There are untranslatable parts in the iD-Editor which are not layer-names:
![image](https://user-images.githubusercontent.com/5360273/26874162-daa34734-4b7c-11e7-8832-27001f4908c8.png)

bhousel:
This has been done for a while

PR #343, opened 2017-06-07
feat(i18n): enable i18n of selected imageries

simon04:
For https://github.com/osmlab/editor-layer-index/issues/17#issuecomment-306755589.

The generated JSON format is supported by Transifex: https://docs.transifex.com/formats/json/

cc: @tyrasd, @manfredbrandl


tyrasd:
Ok, but how would it work from here onward? Suppose we set up a transifex project for this: how/where would we retrieve and store the translated strings?

And is it really worth the effort to set this up for only a handful of strings to be translated? I personally would still prefer a simpler solution where anything is handled via PRs to this repo (as outlined here: https://github.com/osmlab/editor-layer-index/issues/17#issuecomment-230157833).

//cc @bhousel

simon04:
Either a new transifex project, or a separate string bundle within the iD editor.

Transifex supports automatically updating source files: https://docs.transifex.com/projects/updating-content/#automatically-updating-source-files

tyrasd:
sure, but:

> how/where would we retrieve and store the translated strings [from transifex]?

simon04:
Ah, I overlooked this question. One way would be to use the [Transifex Client](https://docs.transifex.com/client/client-configuration), fetch the changes from time to time and commit them to `i18n/`.

bhousel:
> And is it really worth the effort to set this up for only a handful of strings to be translated? I personally would still prefer a simpler solution where anything is handled via PRs to this repo (as outlined here: #17 (comment)).

The nice thing about this is that it's opt-in - only the imagery sources that should be translated would have the i18n flag set.  I think we'd want to set it for everything under `sources/world/`, so it would be more than just the few included in this PR.

We'd also want to get the descriptions too, not just the names.  They would need to have a unique id generated too like `layer_id + '_description'`.

Also, I think I prefer having the translation resources separate instead of as part of the source itself (like `name:fr`, `description:fr`) .  Mostly because if we do start translating it on the iD side, there are around 70 languages, so potentially 140 additional strings per source, which would make the files considerably larger.

> how would it work from here onward? Suppose we set up a transifex project for this: how/where would we retrieve and store the translated strings?

I can definitely push this to the iD Transifex project as a separate resource, and happy to grant admin access to anyone who needs it.

We could have a script like the iD one that connects to Transifex and pulls down the latest sources into this repo.  Anyone who is maintaining this repo would need to do `make translations` sometimes, which isn't too different than how we already have to regenerate the target files sometimes.  

(Yes, in the back of my mind is moving this project to having more proper versions and releases)

simonpoole:
Please do not include translated strings in the data itself, having them in separate files enables them

- to be easily transformed to a different format (assuming that .po format wont happen here in a month of Sundays)
- avoids having to parse lots of stuff that you are not interested in when reading the imagery configs

Naturally I should point out that there is an OpenStreetMap organisation on transifex that can be used for this project.. but I'm not holding my breath,

bhousel:
> Naturally I should point out that there is an OpenStreetMap organisation on transifex that can be used for this project.. but I'm not holding my breath,

@simonpoole I think this has changed since we last looked at it 2 years ago - it looks like now we can transfer iD to the openstreetmap organization and keep all of our translators and teams intact 🤞 

https://docs.transifex.com/projects/transferring-a-project-to-another-organization

If that is the case now, would it be ok for this `editor-layer-index` to be a resource under the iD project, but under the openstreetmap organization?  I'd also send out an announcement to let the iD teams know about the change and encourage them to translate the other openstreetmap projects too!

bhousel:
> @simonpoole I think this has changed since we last looked at it 2 years ago - it looks like now we can transfer iD to the openstreetmap organization and keep all of our translators and teams intact 🤞

Followup: It worked and iD is now moved 🎉 
https://www.transifex.com/openstreetmap/id-editor/announcements/7227/

simonpoole:
@bhousel thanks.

simon04:
How to proceed?
* Add new editor-layer-index project to https://www.transifex.com/openstreetmap/public/?
* Add new resource to https://www.transifex.com/openstreetmap/id-editor/?
* Do not translate at all and close this PR and #17 (wont-fix)

bhousel:
> Add new resource to https://www.transifex.com/openstreetmap/id-editor/?

I'll do this, since the imagery strings all need to be translated to the iD languages anyway.   I can setup the integration with Transifex using the client and write up some instructions and add something to the project `Makefile` to have it run.  (Will probably get to this tomorrow - I'm on vacation today.)

bhousel:
Just merged this.. I also added code to localize the attribution text if present.
TODO: Next steps for integration with Transifex are on me.

bhousel:
> Next steps for integration with Transifex are on me.

This was done!
see https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#translations

PR #344, opened 2017-06-07
feat: add license to schema.json

simon04:
This allows other users of this index to determine layers with compatible licenses.

The license names are taken from Wikimedia Commons: https://commons.wikimedia.org/wiki/Category:Free_Creative_Commons_licenses

"OSM-only" refers to a special permission for usage only in the OSM project.


simonpoole:
Putting on a different hat.

The problem with this change is that the semantics are not clear. Is the assumption that all the layers can be used for tracing/digitization in OSM, and the licence field is simply to indicate the terms for other use? Or is the licence field intended to indicate the licence for all uses? If it is the later then all CC licences with the exception of CC0 are not suitable and such sources should probably not be in the index to start with.

simon04:
> Is the assumption that all the layers can be used for tracing/digitization in OSM, and the licence field is simply to indicate the terms for other use?

Yes. The semantics definitively have to be state more clearly somewhere to make the index useful for outside OSM projects.

simon04:
I would like this index for my project https://github.com/simon04/locator-tool/. So there's at least one practical use-case 😄.

simonpoole:
I still don't quite understand how you want to differentiate

1. imagery only allowed for use for editing with OSM editors (for example bing)
2. imagery with explicit permission for editing with OSM editors and a licence that allows other public use in some form
3. imagery with a public licence that meets our criteria for use in OSM editors and can be used for other public use too

The other dimension of this, is that we typically have not actually investigated the "other public use" scenario and as a consequence will not be able to provide that information without checking.

imagico:
You could add two fields - `license` for indicating the type of license of the data that applies to any use and `permission_osm` with value either `explicit` indicating and explicit permission by the owner for use in OSM (like Bing) or `implicit` indicating the license implicitly allows use in OSM (like USGS topos).  `permission_osm=no` could (in theory) be used for sources that cannot be used for OSM purposes.

What should however be kept in mind is that for use for OSM a compatible license of the data is not sufficient, the TOS of the tile service also need to allow this - although i would assume for anything added in this repository this would be the case.

Klumbumbus:
(Currently readme.md says "This list is purely targeted at OpenStreetMap and does not include layers only useful for other projects such as Open Historical Map if the layers are not also useful for OpenStreetMap.")

bhousel:
Just catching up on this now.  I think this is a reasonable request.  If somebody wants to just pull out all the CC0 sources from this index, we should make it easy for them to do so.

However I think we should:
- Prefer the license to be specified as a standard spdx identifier:  see https://spdx.org/licenses/
- Not use a fixed enum list of licenses  (it implies that anything not in the list is invalid)
Specifying the `license` field as `type: string` seems good enough.

simon04:
Thanks for the many comments. I reworked this PR to use the SPDX License List. Specified `license` attributes are validated using the Python package [`spdx-lookup`](https://pypi.python.org/pypi/spdx-lookup).

To be clarified/discussed:
* some imageries in Austria, Germany specify a national variant of CC BY, such as https://creativecommons.org/licenses/by/3.0/at/deed.de (e.g. layer `basemap.at`) or https://creativecommons.org/licenses/by/3.0/de/ (e.g. layer `bavaria-2m`). Those variants are not listed in SPDX. How to handle those? Specify the international variant from the SPDX index?
* some imageries in Germany use the [Data licence Germany – attribution – Version 2.0](https://www.govdata.de/dl-de/by-2-0) license (e.g. http://suche.transparenz.hamburg.de/dataset/digitale-orthophotos-20cm-hamburg6). This license is not listed in SPDX. Should we [Request New License](https://spdx.org/spdx-license-list/request-new-license)?

simonpoole:
@simon04 we still need to address the permission to use in OSM vs general licence issue, something like @imagico  suggested would work fine.

PS: do we actually have explicit permission for all those CC BY / BY-SA layers? In D + A we need to err on the side of caution with respect to creating derivatives via tracing from such works.

simon04:
I added `permission_osm` as suggested in https://github.com/osmlab/editor-layer-index/pull/344#issuecomment-309689877.

For Austria, the governments agreed to the usage of their images in OSM and accepted attribution on https://wiki.openstreetmap.org/wiki/Contributors#Austria and/or http://www.openstreetmap.org/copyright (a few commits for `/copyright`: https://github.com/openstreetmap/openstreetmap-website/commit/76b568b9f0b1e8dcc19c4c093901fa9f6cf8a9fc, https://github.com/openstreetmap/openstreetmap-website/commit/aa3011dcb7a497796ecb0fc256520da55b3a43cc, https://github.com/openstreetmap/openstreetmap-website/commit/4d64b178b8968421ff30e188a2105e7560efc7f0)

grischard:
@simon04 can you please add documentation and rebase, and then ping me so I can merge this? Thanks!

grischard:
If this ever happens, it should now be PR'd to the `master` branch.

simon04:
I rebased onto the recent changes and resolved various conflicts.

Where are the fields documented [other than in `schema.json`](https://github.com/osmlab/editor-layer-index/pull/344/files#diff-76d5ffeac090c76ed5f4c581f02f3c53)?

grischard:
Fantastic, let's get this merged.

I don't think they're documented anywhere else, which isn't great...

grischard:
Vielen Dank!

simon04:
Awesome, thank you, @grischard!

grischard:
No one has used these attributes since, and it makes synchronization with JOSM more difficult, so unfortunately I'm removing them in e84572e.

andrewharvey:
> No one has used these attributes since, and it makes synchronization with JOSM more difficult, so unfortunately I'm removing them in e84572e.

@grischard Hang on a minute, that kind of change should go through a PR where it can be first discussed. We should protect the master/gh-pages branch and force everything to go through PR, so they have a chance for comment and review by the maintainers here.

It's necessary for this index to document that imagery is allowed to be used in OSM, and by providing the SPDX license it can help document cases where it can be used without a waiver CC0, and cases where it can with a waiver CC BY, etc.

grischard:
I've checked - no one uses it. It was used by very few sources. No one has ever done QA on it.

The license_url property is a lot more useful and used.

andrewharvey:
> I've checked - no one uses it. It was used by very few sources. No one has ever done QA on it.

I use it, it's there for us human maintainers.

I rely on it when reviewing PRs to ensure the source is compatible with OSM licensing and it get's used for auditing existing sources to ensure they are compatible.

The license_url is only part of it, it won't always tell you what license the imagery is under.

simonpoole:
I would have to agree with @andrewharvey on this, there was never an expectation that editors would use this information, but that it could be used for QA and for non-editor consumers ( for example if you are using the layers in a non-OSM-editing application).

Issue #360, opened 2017-09-27
Clarify if only imagery is allowed


joostschouppe:
The documentation isn't entirely clear if you are aiming to collect only imagery services, or also allow (as you seem to do in practice) services with rendered maps when they are cleared for use with OSM.

bhousel:
Sure - we currently welcome all raster sources that are useful for mapping in OSM and licensed appropriately.

The project used to be named `editor-imagery-index` a while ago, but is now named `editor-layer-index` to reflect that it's more than just aerial imagery.  We can add some language to the README and FAQ to make this more clear.

PR #361, opened 2017-09-27
Explicitely include non-imagery

joostschouppe:
As clarified here: https://github.com/osmlab/editor-layer-index/issues/360


bhousel:
Looks good, thanks @joostschouppe!

PR #341, opened 2017-05-30
Switch several tile sources to https

gappleto97:
This is related to https://github.com/openstreetmap/iD/issues/4070 and 02f20f6e6cab39badc2f1e438efbbb2960e7c0fb

PR #362, opened 2017-09-30
HTTPS mapbox's locator overlay layer

paulmach:
Similar to https://github.com/osmlab/editor-layer-index/pull/341

switch the "locator overlay" to https.  Particularly this causes a Mixed-Content warning when viewing ID under https. like here https://www.openstreetmap.org/edit#map=15/37.8714/-122.4563

I wonder if it would be worth the time to programmatically check which layers support https and auto switch them.


paulmach:
added another commit updating the reset of the tiles.mapbox.com sources to https.

bhousel:
Thanks @paulmach!

PR #363, opened 2017-10-02
Create BE_OrthoPhoto2017_Brussels_WMS.geojson

jbelien:
Add 2017 orthophotos from Brussels


sources/europe/be/BE_OrthoPhoto2017_Brussels_WMS.geojson
```diff
@@ -0,0 +1,279 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "UrbISOrtho2016",
```
tyrasd: shouldn't this have `2017` instead of 2016?

jbelien:
Fixed! 
Thanks .

tyrasd:
Nice, thanks @jbelien !

> The Flanders aerial imagery covers Brussels too and is also defined as `best: true` (which is also correct because it is the best imagery for Flanders) ; I don't know how this is handled though ...

I guess that's up to the editor. I think iD currently just chooses the first layer with the "best" flag, but shows the little star symbol for all "best" imagery layers. In principle, an editor might also choose the layer with the smallest extent if it finds multiple covering a particular region (assuming that imagery specifically produced for a small region is of better quality than imagery for a larger region) or the one with the more recent start/end dates…

jbelien:
> //edit: it looks like iD already does the sorting by covered area: openstreetmap/iD@643b4ad

Awesome ! It should work then :)
Do you know at what rate iD updates its baselayer ?

bhousel:
> Awesome ! It should work then :)
> Do you know at what rate iD updates its baselayer ?

Hi @jbelien roughly once or twice a month, and I'd like to push an update soonish, maybe this week..

jbelien:
Hello @bhousel ,

I saw that there was an update of the imagery (https://github.com/openstreetmap/iD/commit/b257e09daadf9323bf3335f74053f41a3fda7b4f) but the new aerial imagery for Brussels is not in it.

As mentioned here :
> The Flanders aerial imagery covers Brussels too and is also defined as `best: true` (which is also correct because it is the best imagery for Flanders) ; I don't know how this is handled though ...

The imagery for [Brussels](https://github.com/jbelien/editor-layer-index/blob/211d6e5f3b35fbabb07dadadd10832ca0037a406/sources/europe/be/BE_OrthoPhoto2017_Brussels_WMS.geojson) is **much** better than the imagery from [Flanders](https://github.com/jbelien/editor-layer-index/blob/211d6e5f3b35fbabb07dadadd10832ca0037a406/sources/europe/be/BE_OrthoPhoto_Flanders_TMS.geojson) but Brussels is indeed geographically "included" into Flanders. Is there a way to make sure that Brussels imagery is the default one in *iD editor* ?

Thanks a lot !

bhousel:
> The imagery for Brussels is much better than the imagery from Flanders but Brussels is indeed geographically "included" into Flanders. Is there a way to make sure that Brussels imagery is the default one in iD editor ?

ohhh I just realized this is a `wms` source, which iD does not currently support.   

If this wms server supports "EPSG:3857", you should be able to create a new tms source for that proxies it through the mapwarper tile proxy like this:
  `http://whoots.mapwarper.net/tms/{zoom}/{x}/{y}/Ortho2017/https://geoservices-urbis.irisnet.be/geoserver/ows/`

I tried this in iD and it didn't work 😢  
Looking at [that server's capabilities](https://geoservices-urbis.irisnet.be/geoserver/ows/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities), it seems like EPSG:3857 is not actually supported on that layer, so this may be the reason.

sources/europe/be/BE_OrthoPhoto2017_Brussels_WMS.geojson
```diff
@@ -0,0 +1,279 @@
...
+        "country_code": "BE",
+        "icon": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAIAAACQkWg2AAAACXBIWXMAAAsSAAALEgHS3X78AAACnUlEQVQoz11SXUiTYRh93ndz06YMNWcpRmWRSpgO+0EUDYwQ6sZ+oAyti8y6yCyKxOimILqQsLKbukmHSpFFLYlMqSkUmi3/pjZ/Zmv+Tec2bU237z1dfFHSc/HcPOc5HM45DAAREREgGOM217BpzDjs7PEHfNGa2PT47KxN+RFqrYDgjBMRE0JijBMRAKOlzmgxhChDN0cla1ThM0uTE65vsRFxRRkXk3TpMiMBkEQQQFPvw0LDnoYvVStBNwAhBIApj+X6m5OlT/dbnX0AJCHxkbl+zhTWub6XA4ZDqcd3JJTl1to7HV7G2E3TxOU2zaWcKm1YZKO5RlaliNknBma6Bqe7V6SVc5mVrmVFxSvLu1G3bSlwy2jVaXE6IzE0JLzV+jxJlxYTvp5X5j3Qx2ePzPcn6VJVysiUaHXX+d2zy8HqlrE8/bq3xXoiStalr1Fpxl1DRMQ547s27NWGRikVKtmuju9e34pEYcrhWV/f7E8iUinDQrjaH/AREQfAmSJGEzc6P0Qkvbd5ymt7DqfEmEr0Tre/4PFXInK4R7z+hXjtRiIi2aKPtpbi+qxue6vHj5ou+8KvAIBPP7yG3mkAd9sry18ULPrdAAgQAIJS4EZLacXrwqVlDwBJiKAkJCEB+Gz/UNyQ1TxY/8dWIiYgFFx5JPXM7JLjWe8jIgKBM3DGXT5no7lmy9rteVsLiIjJGXPGBcQ2XdqBlKJWa1PHeLOCcRAEhKH7zuKy+4T+QohCDQhGjLBqJBGsNl0teZI3OGMG0GiuKarPNI0aZTEyhq0qHxhjXv/C7bYyASkjIddoqctPOnY07ezf5hHRvwcikg+TXtv9jmsOjy0n8eCpnVeIESMiecnE/6kC0D/Vea+90h/wARAQqwG/AaEjsUbwLtXdAAAAAElFTkSuQmCC",
+        "type": "wms",
+        "available_projections": ["EPSG:31370", "EPSG:4326", "EPSG:3857"],
```
bhousel: Check 
https://geoservices-urbis.irisnet.be/geoserver/ows/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities

Looks like only EPSG:31370 is currently supported.

jbelien: Indeed, EPSG:3857 does not appear in the `GetCapabilites` but it's supported.

Check https://geoservices-urbis.irisnet.be/geoserver/ows/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=Ortho2017&WIDTH=256&HEIGHT=256&CRS=EPSG%3A3857&STYLES=&BBOX=486139.4998937212%2C6593763.807992443%2C486445.2480068619%2C6594069.556105584

jbelien:
I didn't know that `wms` was not supported in iD !
I'll have a look ASAP at http://whoots.mapwarper.net/ to try to make it work with this layer !

Thanks for your help ! Much appreciated ! 👍

jbelien:
I figured it out : #368 ! \o/

You just had to remove the ending `/` to make it work ! :)
```
http://whoots.mapwarper.net/tms/{zoom}/{x}/{y}/Ortho2017/https://geoservices-urbis.irisnet.be/geoserver/ows
```
instead of
```
http://whoots.mapwarper.net/tms/{zoom}/{x}/{y}/Ortho2017/https://geoservices-urbis.irisnet.be/geoserver/ows/
```

PR #374, opened 2017-11-23
Update Luxembourg imagery

grischard:
Add 2017 imagery for Luxembourg. The 'best' imagery now points at the 'latest' tiles, so future imagery updates will be integrated automatically, i.e. in 2018 we'll add the 2017 imagery separately.


sources/europe/lu/Luxembourg-GeoportailOrthoLatest.geojson
```diff
@@ -0,0 +1,932 @@
+{
+    "type": "Feature",
+    "properties": {
+        "attribution": {
+            "url": "https://data.public.lu/en/datasets/bd-l-ortho-webservices-wms-et-wmts/",
+            "text": "Administration du Cadastre et de la Topographie",
+            "required": false
+        },
+        "name": "Ortho latest geoportail.lu",
+        "end_date": "2017-06-22",
```
tyrasd: Would it be better to omit the `end_date` here, since this (*latest*) layer will be updated continually?

grischard: Hmm, since those dates are probably used for comparison, I thought I'd update it as new imagery becomes available. You'd simply drop that information?

tyrasd: As far as I know, these dates are mainly intended and used for displaying the "vintage" of the imagery, like in the id editor:
selection_116

They might also play a role in sorting of the layer list, but there the "best" flag should dominate.

I did just ask because the other automatically updated layers in this repo (like the global ones, but also a few regional layers) all don't seem to have the dates set… But if you'd prefer to keep the dates and send a pull request every year or so, that's fine for me, too.

grischard: Ah, thank you. Yes, not very elegant. On the other hand, don't those global and regional layers have a mix of dates? This layer will only ever be one year.

It's a pity the editors don't update the layers more quickly, then I'd just create a static 2017 layer. 

Maybe having both a '2017' layer with the date and a 'latest' best layer with no date would be best? What do you think, would the many layers create a list that's too long for editors?

grischard:
I've now split the 2017 and latest source, removing the dates from the latest source. @tyrasd does this look like it could be merged?

sources/europe/lu/Luxembourg-GeoportailOrtho2017.geojson
```diff
@@ -0,0 +1,931 @@
...
+        },
+        "name": "Ortho 2017 geoportail.lu",
+        "end_date": "2017-06-22",
+        "url": "https://{switch:wmts3,wmts4}.geoportail.lu/opendata/wmts/ortho_2016/GLOBAL_WEBMERCATOR_4_V3/{zoom}/{x}/{y}.jpeg",
+        "max_zoom": 20,
+        "start_date": "2017-06-14",
+        "license_url": "https://data.public.lu/en/datasets/bd-l-ortho-webservices-wms-et-wmts/",
+        "country_code": "LU",
+        "type": "tms",
+        "id": "lu.geoportail.opendata.ortho2016",
```
tyrasd: should be …ortho2017, right?

grischard: Indeed, well spotted! The URL too, actually.

PR #375, opened 2017-11-23
include LINZ CC BY waiver

andrewharvey:
Including LINZ waiver per instruction at https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#where-can-i-find-an-imagery-license-waiver originally from https://wiki.openstreetmap.org/wiki/File:OSM_waiver_-_LINZ.pdf


tyrasd:
Is it really necessary to include the pdf into this repository here, if it's already on the osm wiki? Probably not, especially since https://wiki.openstreetmap.org/wiki/Contributors#LINZ is already linked from here and there one can already find a link to this waiver pdf. :thinking:

andrewharvey:
I'm following the instruction at https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#where-can-i-find-an-imagery-license-waiver says

> Completed imagery waivers should be stored in this repository in the same location as the corresponding .geojson source files.

Personally I think it's a good thing to keep the waivers together in this repo, directly at the source.

tyrasd:
I see. But, interestingly, this would be the first waiver actually doing so (as far as I can see there are no pdf's currently in the layer index). //cc @bhousel

bhousel:
Yes, when I wrote the FAQ section on waivers I didn't really know what to do with them, so I just requested for people to add the waivers directly to the repo.  At least this puts the documents in a place that anybody can see them.

PR #373, opened 2017-11-08
add (new) strava gps traces overlay

tyrasd:
see https://www.reddit.com/r/openstreetmap/comments/7bcsw5/stravas_heatmap_got_updated_and_can_be_used_for/, https://medium.com/strava-engineering/the-global-heatmap-now-6x-hotter-23fc01d301de?_branch_match_id=456415818991382222

josm already has this layer: https://josm.openstreetmap.de/wiki/Maps#Stravacyclingandrunningheatmap

OSM wiki page: https://wiki.openstreetmap.org/wiki/Strava#Data_Permission_-_Allowed_for_tracing.21

fixes #294 

related: https://github.com/strava/iD/commit/104ae9ded4b51ab1b919273db30f58f3e5a577f0

PR #376, opened 2017-12-05
Add Strava Global Heatmap

zbycz:
The layer is "Allowed for tracing" - see http://wiki.openstreetmap.org/wiki/Strava#Global_Heatmap


bhousel:
We're already tracking this on #373 - just because the OSM wiki says it is "allowed for tracing" does not make it so.

zbycz:
Ok, thanks, sorry for not checking first :)

PR #377, opened 2017-12-24
Update Toulouse orthophotos links

don-vip:
OSM-FR TMS proxy is broken for Toulouse orthophotos, switch back to source WMS


don-vip:
JOSM supports max-zoom for WMS too: https://josm.openstreetmap.de/wiki/Maps#WebMapServicesWMS
I will add projections, sure!

PR #378, opened 2017-12-29
Update Luxembourg-GeoportailOrthoLatest.geojson

Klumbumbus:
switch to http, see comments at https://josm.openstreetmap.de/ticket/15706

### DIFF
--- a/sources/europe/lu/Luxembourg-GeoportailOrthoLatest.geojson
+++ b/sources/europe/lu/Luxembourg-GeoportailOrthoLatest.geojson
@@ -7,7 +7,7 @@
             "required": false
         },
         "name": "Latest available ortho geoportail.lu",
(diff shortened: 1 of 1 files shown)


grischard:
This won't give a mixed content warning?

tyrasd:
> Because of their internal infrastructure configuration, the https services of geoportail.lu don't reliably direct the clients to the same backend servers, and ssl resumption doesn't work. Since this makes the https services slower than http, I've switched to the http version.

Shouldn't this be something that – instead of introducing workarounds here and there – one would rather report to the maintainers of the service and have it fixed there? Or has that already been tried and received a wontfix reply?

grischard:
It's been tried, and has received a "we'll fix all that stuff some day" reply. Not to sound negative - they're great people who are actively improving their infrastructure, and I do believe they will actually get around to it when the higher priority stuff on their big plate is done.

bhousel:
Thanks!  Sorry for the slow merge..

PR #379, opened 2018-01-13
Add N50 topographic map for Norway

NKAmapper:
N50 topographic map from "Kartverket", the Norwegian Mapping Authority.

Explicit permission filed on Norwegian OSM mailing list: [Permission](https://lists.nuug.no/pipermail/kart/2014-August/004831.html)

OSM contribution page: [Kartverket contribution](https://wiki.openstreetmap.org/wiki/Contributors#Kartverket_.28Norwegian_Mapping_Authority.29)

JOSM imagery entry: [JOSM maps](https://josm.openstreetmap.de/wiki/Maps/Norway#KartverketN50topo)

### DIFF
--- a/sources/europe/no/Kartverket-N50-topo.geojson
+++ b/sources/europe/no/Kartverket-N50-topo.geojson
@@ -10,7 +10,6 @@
         "icon": "https://www.kartverket.no/Content/Images/logo-graphic-512.png", 
         "min_zoom": 3, 
         "max_zoom": 15,
-        "best": true,
         "attribution": {
             "required": true, 
             "text": "\u00a9 Kartverket", 
--- a/sources/europe/no/Kartverket-N50-topo.geojson
+++ b/sources/europe/no/Kartverket-N50-topo.geojson
@@ -1,5 +1,5 @@
 {
-"type": "Feature",
+    "type": "Feature",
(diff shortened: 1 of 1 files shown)


tyrasd:
Hi. Thanks for the pull request. Just one remark: Do you think it's a good idea to set the `best` flag to true on this topographic map layer? Has this been discussed in and decided by the community? IMHO, this flag should be used with aerial imagery layers, because if we would show a topographic map as the default mapping background, a (new) user might think that OSM is all about copying information from a reference map, while it should be more about [mapping what is on the ground](https://wiki.openstreetmap.org/wiki/Good_practice).

NKAmapper:
I fully understand your comment. 
Here is the situation: We currently have no country specific imagery in iD. New mappers therefore use Bing (the default choice) without correcting for any offset. This has created a massive offset problem here and it has been discussed many times in the local community that we would like to move away from Bing. Our problem is that we do not have the right to use the (much better) local ortophotos, which leaves us with the N50 topo map as the only alternative.

I do agree, however, that we should not give a new mapper the wrong impression. I can remove the best flag now and then we can first monitor how the N50 topo is being used. It will anyway be a big improvement to have our first local imagery available in iD.

Issue #345, opened 2017-06-09
Add NY State orthos ("latest" layer)


bhousel:
When **requesting the addition of an imagery layer**, please provide the following information. Otherwise, delete the text and type your issue.

## URL of the imagery
http://www.orthos.dhses.ny.gov/ArcGIS/rest/services/Latest/MapServer/

see also
https://gis.ny.gov/gateway/mg/webserv/webserv.html
https://gis.ny.gov/gateway/mg/metadata.cfm
https://gis.ny.gov/gateway/mg/nysdop_descrip.htm
https://gis.ny.gov/gateway/orthoprogram/index.cfm  (6" available in many areas now)

## Two letter country code
US - NY

## Desired layer name
any

## License
unclear - needs investigation

## Is the license compatible with OSM?
Not sure

kennykb:
Several emails to the program manager for NYS Digital Orthoimagery in the autumn went unanswered, alas. Nevertheless, at long last, I found [metadata in Open NY.](https://data.ny.gov/Public-Safety/Statewide-Digital-Orthoimagery/amsv-rmfz)  It includes the line:

> **Limitations** | No limitations on use of publicly available imagery.

(That's also in the XML metadata. I missed it the first time because I was looking for a License element, and there was none; this is in the Disclaimers element. Oh well, its not the only time that someone didn't quite follow a schema.)

I think we can take this as an indication that the data are in the public domain. I was about 95% sure of this, owing to the express written policies of Gov. Cuomo and the NYS Committee on Open Government, but was trying to get further assurance. With an explicit 'no limitations on use' policy in the metadata, I think we're good to go.

And I'm eager to get started, because the 2017 data for my county are up and will allow me to correct some questionable tracings. This is the only really good source for 'leaves down' imagery that I've been able to find. (Comparing summer and winter images is incredibly useful when tracing.)

You already have the link up above to the [description of the tile server](http://gis.ny.gov/gateway/mg/webserv/webserv.html). 

How can I help further?

PR #381, opened 2018-01-15
update sources to include New York State Orthos Online

kennykb:
An attempt at including New York State Orthos Online - issue [345](https://github.com/osmlab/editor-layer-index/issues/345)

Let me know if there's anything I need to do here! 

Reminder: the clearance can be found under 'Disclaimers' at [https://data.ny.gov/Public-Safety/Statewide-Digital-Orthoimagery/amsv-rmfz](https://data.ny.gov/Public-Safety/Statewide-Digital-Orthoimagery/amsv-rmfz): 

Limitations No limitations on use of publicly available imagery.


kennykb:
Oh, also, is there some sort of automation by which this change propagates over to [https://josm.openstreetmap.de/wiki/Maps](https://josm.openstreetmap.de/wiki/Maps) or does that have to be done separately to have the layer available to JOSM?

tyrasd:
> Oh, also, is there some sort of automation by which this change propagates over to https://josm.openstreetmap.de/wiki/Maps or does that have to be done separately to have the layer available to JOSM?

It needs to be done manually, but one can also set-up JOSM to use this imagery source: https://github.com/osmlab/editor-layer-index#using-this-index

kennykb:
Sorry, I hadn't seen available_projections in the first five or six GeoJSON files that I'd looked at. It seems that many others have made the same mistake, leading me to presume that the server's announcement of its own capabilities would govern.

I left "CRS:84" out of the list since I didn't see anything else using that namespace, and it's redundant with "EPSG:4326".

tyrasd:
You've probably looked only at tms sources, for which it's not necessary to specify the field. Or are there still wms layers in the index without `available_projections`?

The idea behind this field is the following: Let's say there's some software (e.g. an osm editor) using this index, which only supports a handful of projections (say, EPSG:3857 and its pseudonyms). Using the `available_projections`, the software can decide which layers it should show on a list of compatible layers, without having to check all `GetCapabilities` of all possible WMS layers all of the time (or alternatively, when showing an unfiltered list: without having to give the user an error message after they selected a layer which will turn out to be incompatible with the supported projections of the software). Hope this clarifies this a bit.

PR #382, opened 2018-01-16
Change AGRI (Australia) and Surrey Air Survey (UK) to also load over https

SomeoneElseOSM:
Both AGRI and SAS imagery also now redirect to https, and don't load into P2 on drag when defined as http.

As with https://github.com/osmlab/editor-layer-index/pull/380 , I've not tested the updated editor-imagery-index in an editor, but tiles from these layers do as expected load over https.  A number of other osm.org-hosted layers have not yet been changed to http because e.g. http://ooc.openstreetmap.org/npe/15/16303/10460.png does not yet redirect.  Also some layers don't seem to load at all (presumably they have been superceded).

### DIFF
--- a/sources/australia/au/AGRIblack-and-white25m.geojson
+++ b/sources/australia/au/AGRIblack-and-white25m.geojson
@@ -2,7 +2,7 @@
     "type": "Feature",
     "properties": {
         "id": "AGRI-black_and_white-2.5m",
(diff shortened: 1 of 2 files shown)


tyrasd:
Thanks, but it looks like as if @tomhughes was a bit faster with 81529b0c36f716ea67cd69cde106b8599acb67cf and 21ddf1f5f9773855b2987444549c89804eb9bea3

PR #384, opened 2018-01-18
Improve waymarkedtrails attribution

lonvia:
I prefer that waymarkedtrails attribution goes to the
site instead of a single person as there are other
contributors. Also removed reference to ODbL to make
it a bit shorter and it seems to be in line with rest
of this repo.


tyrasd:
Cool, thanks Sarah!

PR #385, opened 2018-01-18
Add a timestamp when the file was generated and a version string

simonpoole:
In lieu of versioning the file, what about simply adding a timestamp and a format version?

The geojson spec doesn't allow a properties object in a FeatureCollection so I've named it "meta".


grischard:
Should this also be part of the schema file?

simonpoole:
The schema file currently only covers the individual entries iirc. and the check is run on those and not on the concatenated version.

bhousel:
This seems fine to me, and has been here for a month with nobody raising issues, so let's just merge it..

PR #358, opened 2017-08-22
add Esri satellite imagery to list of providers

jgravois:
resolves #357 

- [x] substitute a real base64 encoded Esri logo
- [x] create [new](https://wiki.openstreetmap.org/wiki/Esri) OpenStreetMap wiki page
- [ ] ?

PR #391, opened 2018-01-30
Adding a second 'World Imagery (Clarity)' layer

jgravois:
[Esri World Imagery (Clarity)](http://www.arcgis.com/home/item.html?id=ab399b847323487dba26809bf11ea91a) - two for the price of one!

ref: https://github.com/osmlab/editor-layer-index/pull/358#issuecomment-359237579


bhousel:
Awesome, thanks @jgravois 🙌

simonpoole:
The imagery files were not regenerated in/after this PR.

bhousel:
@simonpoole I think 9b52fb1 includes the regenerated files?
In any case, I just did a clean build and git tells me nothing changed.

simonpoole:
Operator error on my behalf, sorry. The underlying issue is that we don't minimize the output of concatenating the sources in the repo, which forces an extra step for no good reason on consumers that need to be data size aware. I minimize with minimize-js (node), would it be acceptable to do that for the output files for the whole repo?

bhousel:
> would it be acceptable to do that for the output files for the whole repo?

Sure, seems fine to me..

PR #398, opened 2018-02-16
Add missing date for several source in France

Marc-marc-marc:
I have added several missing start_date and end_date for several source in France


tyrasd:
thanks!

PR #401, opened 2018-02-17
Added type=wms_endpoint to schema

grischard:
None of the editors that use ELI currently use the wms catalogs that are still quite rare. This will allow ELI contributors to create them, encouraging editors to support them.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -36,7 +36,8 @@
                         "tms",
                         "wms",
                         "bing",
-                        "scanex"
+                        "scanex",
+                        "wms_endpoint"
                     ]
                 },
                 "url": {


bhousel:
What is this for?

tyrasd:
It's for linking to a wms endpoint where all layers are potentially interesting for mapping (here's one in JOSM's index: [link](https://josm.openstreetmap.de/wiki/Maps/Luxembourg#geoportail.lualllayers)). An additional benefit (compared to supplying just a selected list of individual WMS layers) is that the end user can select exactly what layers are interesting for them (e.g. only building footprints and housenumbers, or roads and roadsnames on top of some aerial imagery). The downside is that creating intuitive and easy to use UI for this is quite hard (see JOSM's dialog below, or QGIS).

selection_117

grischard:
It's not always a mess - the layers at http://wmts1.geoportail.lu/opendata/service?REQUEST=GetCapabilities are lovingly arranged in subfolders.
image

bhousel:
Cool, SGTM 👍

PR #402, opened 2018-02-18
France : Cadastre : better shape + adapted name+id to match current practice

Marc-marc-marc:
@grischard has improved the shape to be less overflowing on the bordering countries.
I have converted his update to geojson.
I also adapted name+id to match current practice.


tyrasd:
cool, thanks!

PR #403, opened 2018-02-18
Belgium updates

grischard:
Fixes borken layers, updates shapes, add one interesting wms.


Klumbumbus:
For UrbisAdm FR and UrbisAdm NL JOSM uses these two URLs instead:

https://geoservices-urbis.irisnet.be/geoserver/ows/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image/png&TRANSPARENT=true&LAYERS=urbisFR&WIDTH={width}&HEIGHT={height}&CRS={proj}&STYLES=&BBOX={bbox}
https://geoservices-urbis.irisnet.be/geoserver/ows/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image/png&TRANSPARENT=true&LAYERS=urbisNL&WIDTH={width}&HEIGHT={height}&CRS={proj}&STYLES=&BBOX={bbox}

By a quick comparison I see some advantaged of the JOSM URLs
* there are property lines
* the house numbers are alligned to their street
* it seems the "ELI version" uses outdated data. See here http://osm.org/go/0EoY2pu48?layers=N&m= for comparison of the both variants. There are 2 new buildings, which can be seen with old and new aerial imagery (in JOSM).

PR #408, opened 2018-02-19
Updated URLs to newer (?) ELI version

grischard:
See https://github.com/osmlab/editor-layer-index/pull/403#issuecomment-366525927

### DIFF
--- a/imagery.xml
+++ b/imagery.xml
@@ -5977,7 +5977,7 @@
     <name>UrbisAdm FR</name>
     <id>UrbisAdmFR</id>
     <type>wms</type>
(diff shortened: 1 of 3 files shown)


Klumbumbus:
There is a typo in both new URLs (missing `/` after ows) and the URL encoding is not necessary too.

tyrasd:
> typo […] missing `/` after ows

Not sure about that. I believe it doesn't matter, but https://geoservices-urbis.irisnet.be/geoserver/web/ itself advertises the WMS without a slash after `ows`:

```
https://geoservices-urbis.irisnet.be/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities
```

grischard:
I've removed the / in JOSM in https://josm.openstreetmap.de/wiki/Maps/Belgium?action=diff&version=108

Klumbumbus:
Ok. I tested the version without / and it didn't work. Obviously I made a mistake at my test. Works now for me too.

PR #413, opened 2018-02-19
Add 10cm imagery from 2017 for the Canton Thurgau

simonpoole:
This imagery has a attribution only licence and we have obtained permission to attribute via the contributor page on the [wiki](https://wiki.openstreetmap.org/wiki/Contributors#Canton_Thurgau).


simonpoole:
PS: this is proxied from a single layer WMS server that doesn't support EPSG:3857 so it doesn't really make sense to add the WMS variant

PR #421, opened 2018-02-22
Minify output

### LINKED ISSUE #393: Minify the generated files
per https://github.com/osmlab/editor-layer-index/pull/391#issuecomment-362369626
We could generate minified versions of `imagery.json` and `imagery.geojson` too.

grischard:
Fixes #393


grischard:
I would be very thankful if someone could review this before merging.

PR #422, opened 2018-02-23
adjust names of Lyon sources

Klumbumbus:
remove code from name and in style of other sources, i.e. without `-`


Klumbumbus:
(without make)

grischard:
Merci beaucoup !

PR #427, opened 2018-02-26
fix NSW LPI WMS URLs + add ImageryDates layer

andrewharvey:
fixes the issue reported at https://github.com/osmlab/editor-layer-index/pull/118#issuecomment-368329013 thanks @tyrasd


grischard:
Thanks!

PR #429, opened 2018-03-02
Add two WMS overlays for Norway

NKAmapper:
Thank you for implementing WMS in iD - great feature!

Here are two very useful WMS overlays for Norway:
1) Marked hiking trails
2) Road network

Both services are provided by the Norwegian Mapping Authority (Kartverket).

[Explicit permission and discussion on the Norwegian mailing list.](https://lists.nuug.no/pipermail/kart/2014-August/004831.html)

[Overview of Norwegian import sources.](https://wiki.openstreetmap.org/wiki/Norwegian_import_sources)


bhousel:
Looks great, thanks @NKAmapper 👍

PR #430, opened 2018-03-04
add 32 missing license_url

Marc-marc-marc:


bhousel:
Thank you for tracking these down!  🙇

simonpoole:
Does the imagery/server referenced in CantondeFribourg50cm-2005.geojson still exist?

tyrasd:
@simonpoole – I asked the same question here: https://wiki.openstreetmap.org/wiki/Talk:Switzerland:Fribourg

simonpoole:
Ah, you should really know that OSM is not wikipedia :-) A talk page is roughly equivalent to /dev/null

Marc-marc-marc:
@simonpoole we are talking about it on talk-ch :-) and still waiting news to update the url.
This PR focus on and only on license_url
We need another pr/tools/guideline to check/detect/tag/remove dead/tmp-broken layers after x days/week/months

tyrasd:
> A talk page is roughly equivalent to /dev/null

:woman_shrugging: I got a response from a Latvian community member here: https://wiki.openstreetmap.org/wiki/Talk:Latvia/L%C4%A2IA_ortofoto

PR #436, opened 2018-03-12
Fix `end_date` for Brussels imagery

jbelien:
Those are yearly release of aeriel imagery. I think I wrote it wrong on initial commit and that's why it's still 2016 aerial imagery that is displayed in iD (for instance).
This PR should fix that (unless, I'm wrong).


jbelien:
My mistake ... `UrbIS-Ortho 2017` is available in iD !
But still, if I understood correctly, this PR fix the `end_date` error :)

grischard:
Looks good! Thank you!

PR #440, opened 2018-03-13
Update BE_OrthoPhoto2017_Brussels_WMS.geojson

jbelien:
Now iD supports WMS, no need to use whoots.mapwarper.net as a proxy anymore !


grischard:
Vespucci also supports WMS now, so this is excellent. Are there still a few Potlatch users in your area?

jbelien:
> Are there still a few Potlatch users in your area?

@joostschouppe is probably the last Potlatch user in Belgium 😂

grischard:
Hmmm. Joost, you can probably still use the whoots url as a custom layer then.

joostschouppe:
I use whatever I need to get things to work, so don't mind me. Worst case scenario is a Potlatch user in Brussels using the AIV (Agiv) Imagery, and that is kind of the same quality.

PR #446, opened 2018-03-14
add actmapi imagery 2017

### LINKED ISSUE #297: Add Canberra ACT, Australia imagery
Thread:
https://lists.openstreetmap.org/pipermail/talk-au/2017-March/011286.html

Will flesh out the other details required in a bit

## URL of the imagery
(required)
WMS - http://data.actmapi.act.gov.au/arcgis/rest/services/actmapi/imagery2016mga/MapServer/WMTS/1.0.0/WMTSCapabilities.xml

On the mailing list there is a suggestion:
>  it would be nice if we could find somebody to run a mapproxy instance or similar so
> that it could be used in iD and other editors that don't support WMS
> servers. We (SOSM) would likely offer, but it doesn't really make sense
> to ship the bits around half the globe and back again.

## Two letter country code
AU

## Desired layer name
ACT2016 (28355)

## License
CC-BY-4.0

We have been given explicit permission from ACTMapi to incorporate their CC-BY licensed datasets into OpenStreetMap provided that this attribution is made on contributions page

## Attribution (text, URL, required?)
https://wiki.openstreetmap.org/wiki/Contributors#Australian_Capital_Territory_Government_data

## Icon
(optional)

## Boundary polygon
(if possible, provide a boundary polygon, e.g. as a GeoJSON or OSM file.)
http://actmapi-actgov.opendata.arcgis.com/datasets/5739476ddd854f3b9f47bef679378f30_1 ? (Might have to investigate that a bit more/convert format)

andrewharvey:
closes #297


grischard:
Looks great, thanks!

Any other interesting layers on that server?

andrewharvey:
> Any other interesting layers on that server?

There are other data sources we can use, I'm still working out which combinations would be useful as a raster background layer as opposed to someone using the vector data.

PR #451, opened 2018-03-19
Create openlabs-geoportal.geojson (Tirana, Albania)

jbelien:
Add the layers from [Open Labs Geoportal](https://geoportal.openlabs.cc/).

### DIFF
--- a/sources/europe/al/openlabs-geoportal.geojson
+++ b/sources/europe/al/openlabs-geoportal.geojson
@@ -1,0 +1,38 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "openlabs-geoportal-tirana",
+        "name": "Municipality of Tirana (Open Labs GeoPortal)",
+        "type": "tms",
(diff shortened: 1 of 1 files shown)


Klumbumbus:
In JOSM you set max-zoom=21 and here you didn't set max-zoom. Was this an copy paste mishap? When I test the layer I get tiles only up to z20.

jbelien:
Thanks @Klumbumbus, I'll check and fix it tomorrow :)

Klumbumbus:
(I just changed it to 20 in JOSM for now.)

jbelien:
I just checked !
I confirm that the GeoPortal serves tiles up to level 21 included !

Here is an example : 
https://geoportal.openlabs.cc/mapcache/tms/1.0.0/tirana@GoogleMapsCompatibleExtended/21/1164024/1313434.png


I'll fix JOSM (put level 21 back) and add `max_zoom` parameter here to 21 as well :)

Klumbumbus:
Ah, you're right. I fell into the trap that JOSM defaults the max_zoom to 20 if not set. 21 works for me now too.

PR #454, opened 2018-03-21
switch craig layers to wmts

tyrasd:
By using `wmts` instead of `tms`, these layers don't have to use the "non-standard" `{zoom-1}` parameter anymore (which is supported by iD for example).

follows #417 – @Marc-marc-marc any objections against using the `wmts` variant here?


Marc-marc-marc:
@tyrasd no objections

PR #458, opened 2018-03-23
Add reference URLs for layers in Austria.

hikemaniac:
This PR adds reference URLs for several layers in Austria. It also fixes one typo where the text says "goverment" instead "government". As this is my first contribution to this repository, can you please verify its correctness?

### DIFF
--- a/sources/europe/at/Basemapat.geojson
+++ b/sources/europe/at/Basemapat.geojson
@@ -8,7 +8,7 @@
             "text": "basemap.at",
             "required": true
         },
-        "description": "Basemap of Austria, based on goverment data.",
+        "description": "Basemap of Austria, based on government data.",
         "i18n": true,
         "url": "https://maps{switch:1,2,3,4}.wien.gv.at/basemap/geolandbasemap/normal/google3857/{zoom}/{y}/{x}.png",
         "license_url": "https://basemap.at/#sec-nutzung",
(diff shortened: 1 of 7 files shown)


grischard:
Much appreciated, thank you!

Since you're interested in Austrian layers: the VoGIS and Vienna layers are missing a license property. Would you have any idea of what should go there?

hikemaniac:
@grischard I did some research on the other layers licenses and found out that all the Vienna layers are licensed under [Creative Commons Attribution 3.0 Austria](https://creativecommons.org/licenses/by/3.0/at/deed.de) according to [data.gv.at](https://data.gv.at). For the VoGIS layers the situation is not as easy. They are not published under a proper license but under certain [usage terms](http://vogis.cnv.at/nutzungsbestimmungen/nutzung.pdf) which say, that the data may be used when the government of Vorarlberg is informed about it (which is probably true for this project and OSM).

Should I submit another pull request for adding the `license_url` attribute?

tyrasd:
//cc @simon04

grischard:
The CC-BY license is not fully compatible with the ODbL, so we'll [need a tiny bit of extra paperwork](https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/).

I'm not sure they understand the CC-BY fully, as their usage terms contradict CC-BY. In any case, we can't guarantee that OSM won't be used, for example, for pornography.

I'm not sure the VoGIS email is enough either. Any thoughts, @simonpoole ?

simonpoole:
Yes, their additional usage terms contradict the terms in CC BY (any version, or actually any "open" license). They need to make up their mind.

PR #478, opened 2018-04-20
LINZ data are cc by 4.0

andrewharvey:
updating license text as per the updated information from LINZ

### DIFF
--- a/sources/australia/nz/LINZ-NZ-Aerial-Imagery.geojson
+++ b/sources/australia/nz/LINZ-NZ-Aerial-Imagery.geojson
@@ -3,7 +3,7 @@
     "properties": {
         "attribution": {
             "url": "https://www.linz.govt.nz/data/licensing-and-using-data/attributing-elevation-or-aerial-imagery-data",
-            "text": "Sourced from LINZ CC-BY 3.0",
+            "text": "Sourced from LINZ CC-BY 4.0",
             "required": true
         },
         "license_url": "https://wiki.openstreetmap.org/wiki/Contributors#LINZ",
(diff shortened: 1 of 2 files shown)


simonpoole:
Do we have a waiver from LINZ for this imagery?

andrewharvey:
@simonpoole Yes, it's sitting in this repository at https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/australia/nz/LINZ-NZ-waiver.pdf per [the contributing instructions](https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#where-can-i-find-an-imagery-license-waiver) as well as on the OSM wiki. It covers LINZ Data Service Datasets and Basemaps which includes the imagery and topo basemaps indexed here.

andrewharvey:
@simonpoole Oh sorry you mean an updated waiver for the CC BY 4.0 license, since the current waiver is for CC BY 3.0. I'll try to get it completed by LINZ next week. Thanks for picking up on that.

simonpoole:
For imagery there are two options:

- preferred something based on https://drive.google.com/file/d/0B3PN5zfbzThqLXg1TUlxalAtVE0/view
- minimum the CC BY 4.0 waiver https://drive.google.com/file/d/0B3PN5zfbzThqeTdWR1l3SzJVcTg/view

Given that NZ has a sweat of the brow doctrine similar to the UK and AUS I don't really think we can ignore the ramifications of the change to 4.0.

andrewharvey:
I agree the first is preferred, but it also means they aren't guaranteed attribution that their imagery is used to derive data from for OSM, which they may require for CC BY compatibility. I don't know but I'll ask if they're willing to agree to the imagery waiver, if not I'll seek the standard CC BY waiver updated to point to the 4.0 clauses.

Issue #439, opened 2018-03-13
Sources that do not contain a license


grischard:
202 sources don't have a license property.

- [ ] australia/au/AGRIblack-and-white25m.geojson
- [ ] asia/CambodiaLaosThailandVietnambilingual.geojson
- [ ] asia/ru/ScanExIRS.geojson
- [ ] asia/ru/ScanExSPOT.geojson
- [ ] asia/ph/PangasinnBulacan-PhilippinesHiRes.geojson
- [ ] world/QANoAddress.geojson
- [ ] world/OpenStreetMap-MapnikBlackWhite.geojson
- [ ] world/OpenStreetMap-GPS.geojson
- [ ] world/OSMIHighways.geojson
- [ ] world/WMT-hiking.geojson
- [ ] world/WMT-slopemap.geojson
- [ ] world/DigitalGlobePremiumImagery.geojson
- [ ] world/WMT-skating.geojson
- [ ] world/OSMIAddresses.geojson
- [ ] world/OSMIRouting.geojson
- [ ] world/OpenStreetMap-GermanStyle.geojson
- [ ] world/OSMITagging.geojson
- [ ] world/OpenPTMap-overlay.geojson
- [ ] world/Skobbler.geojson
- [ ] world/WMT-mtb.geojson
- [ ] world/OSMIPlaces.geojson
- [ ] world/OpenStreetMap-Mapnik.geojson
- [ ] world/DigitalGlobeStandardImagery_vintage_overlay.geojson
- [ ] world/DigitalGlobePremiumImagery_vintage_overlay.geojson
- [ ] world/StamenTerrainBackground.geojson
- [ ] world/DigitalGlobeStandardImagery.geojson
- [ ] world/WMT-cycling.geojson
- [ ] world/LocatorOverlay.geojson
- [ ] world/OSMIGeometry.geojson
- [ ] world/OSMIMultipolygon.geojson
- [ ] world/EsriImagery.geojson
- [ ] south-america/br/IBGE_DF_Adresses.geojson
- [ ] south-america/br/IBGE_Salvador_Streets.geojson
- [ ] europe/sk/SlovakiaHistoricMaps.geojson
- [ ] europe/sk/FreemapskHiking.geojson
- [ ] europe/sk/FreemapskCar.geojson
- [ ] europe/sk/FreemapskSki.geojson
- [ ] europe/sk/FreemapskCyclo.geojson
- [ ] europe/se/Sverigekartan
[...]

PR #486, opened 2018-05-11
update australia/au/AGRI metadata

andrewharvey:
This source is mostly only historically useful, but generally we now have access to more current and higher or same resolution imagery through the Bing, ESRI, DigitalGlobe and Mapbox layers, so I practically never see it used anymore.

That said, it's still available and may be useful to align other layers since it went through a process to ensure it's positionally accurate (or at least per the GDA94 datum).

Addresses part of #439.


sources/australia/au/AGRIblack-and-white25m.geojson
```diff
@@ -4,12 +4,16 @@
         "id": "AGRI-black_and_white-2.5m",
         "url": "https://{switch:a,b,c}.agri.openstreetmap.org/layer/au_ga_agri/{zoom}/{x}/{y}.png",
         "attribution": {
-            "text": "AGRI",
+            "text": "AGRI, Geoscience Australia",
+            "url": "https://data.gov.au/dataset/agri-the-australian-geographic-reference-image"
```
tyrasd: there's a comma missing at the end of this line

andrewharvey: thanks. fixed now

andrewharvey:
@tyrasd fixed.

PR #487, opened 2018-05-11
Add 2016 canton Fribourg imagery

simonpoole:
The is imagery from November 2016 created during  LIDAR data gathering flights and not mainly intended as a "pretty" orthophoto.  We have (in e-mail) permission to use this, however there are concerns about load on the origin WMS server so I would suggest only using our proxy. The other question is if this is really the best imagery for the area and how well it is aligned, @Marc-marc-marc any feeling on if we should set the "best" flag?


grischard:
Travis says your json is borken.

simonpoole:
For whatever reason my json editor didn't complain about the borkeness, :-) should be fixed now.

simonpoole:
Seems as if I simplified the coverage polygon a bit too much will replace that later today.

Marc-marc-marc:
alignment seems good if I compare it with what I have tried to align in the past (delicate operation given the absence of a valid source for survey points)
the quality of the imagery is quite variable. very good in sunny areas (the best available in this area, and also the most up-to-date), it degrade in the shade (DigitalGlobe Standard is in this case more "usable" to "understand" the ground despite its lower resolution, this new layer is still in this case the best layer to draw if you can see the object "in the dark")
I merge the PR with "best: false", feel free to comment the best tag

bryceco:
The polygon for this is still weird. There is an extra array nesting for each polygon, so instead of being an 'array of arrays of points' it is an 'array of arrays of arrays of points'

tyrasd:
yeah, multipolygons are actually not yet supported (see #217), but that might change soon with #490

Klumbumbus:
This is a good example that the admin boundaries are not usable for imagery bounds in most cases.
grafik

Klumbumbus:
min zoom is 1 for me, instead of 8

simonpoole:
> min zoom is 1 for me, instead of 8
Well at zoom 1 it looks like 
grafik
which will definitely enable highly detailed tracing for OSM.

Klumbumbus:
>which will definitely enable highly detailed tracing for OSM.

:D however the same can be said about z8.

grafik

Marc-marc-marc:
I have made some fix : 
- Polygon in stead of unsupported Multipolygon
- better shape (avoid unneeded gap and drop some nodes)
- min_zoom 1 to better match the layer capability (of course zoom 1 is unusable to tracing)

Some work can still be done if somebody want it : 
- reduce the lat&lon precision (currently 7 digits after the decimal point)
- reduce the number of node
- move some nodes for a better match of the shape

but at least it should be working right now.
Let's me known what you thing about it.

bryceco:
That works great. Thanks!

PR #491, opened 2018-05-17
LINZ switched to CC BY 4.0, updating the waiver

andrewharvey:
For some reason the changes from #478 were reverted by the travisbot in https://github.com/osmlab/editor-layer-index/commit/19c6117a80df64ec1bd3104e9d1d799bdd50330b#diff-771b451a034b340641ad635c90493556 I've re-introduced those changes here, but any idea why travis reverted the change?


tyrasd:
> the changes from #478 were reverted by the travisbot

huh, that's odd. let's try it again.

//cc @grischard

grischard:
That's very odd. No idea, maybe concurrent commits? I'll see if we can exclude `sources/` from travis.

PR #500, opened 2018-05-28
new OSMIM layers from May 2018

imagico:
New layers.

I also added a feature for self-description of the tile layers.  If you call

http://imagico.de/map/osmim_tiles.php?layer=

you get a list of layers available.  With

http://imagico.de/map/osmim_tiles.php?layer=northsea_s2_2018

you get an extended tilejson description of the layer and with 

http://imagico.de/map/osmim_tiles.php?layer=northsea_s2_2018&footprint

you get the layer footprint.

Pulling these in automatically is probably not a good idea since the main purpose of this repository is to provide a hand curated collection of layers but it might offer possibilities for more efficient updates in the future.  It might also make sense for other image layer providers to create similar descriptions.


grischard:
Thanks!

PR #501, opened 2018-05-28
Add orthophoto for Norway

### LINKED ISSUE #484: Default layer for Norway
Is it possible to set the default layer in Norway as Digital Globe Standard instead of Bing? DG has superior imagery and better offset compared to Bing.

Someone mentioned this was possible, but I don't see how it can be done in this repo. 

A polygon for Norway(no) exists already with the  editor-layer-index/sources/europe/no/Kartverket-N50-topo.geojson-file.

NKAmapper:
Permission: [Link](https://forum.openstreetmap.org/viewtopic.php?id=62083)

Fixes #484


grischard:
Thanks! Do you think we could simplify the Norwegian polygon's path over the North and Norwegian seas?

NKAmapper:
>Thanks! Do you think we could simplify the Norwegian polygon's path over the North and Norwegian seas?
I would rather keep it. That way it could be reused for any map within the boundary of Norway.

PR #504, opened 2018-05-30
add bologna orthophoto 2017 layer

tyrasd:
see https://twitter.com/WikimediaItalia/status/1001460635861995520 and http://dati.comune.bologna.it/node/3471. Partially re-used info from https://josm.openstreetmap.de/wiki/Maps/Italy?sfp_email=&sfph_mail=&action=diff&version=55&old_version=53&sfp_email=&sfph_mail=


grischard:
JOSM has 7 decimal points, which is already overkill. I wonder how you managed to get 15! :)

imagico:
15 digits is the default for OGR writing GeoJSON:

http://gdal.org/drv_geojson.html

grischard:
Ahh, that explains it then. Still not sure why it invented extra digits, but in any case the build scripts round all coordinates to 5 digits.

Issue #428, opened 2018-02-28
Linter improvements


grischard:
The QA of the current sources would be made easier if we had a better test suite than our current `make check`. #425  is an excellent addition, #211 would be great.

In addition to those, it would be good if we could easily check sources for missing attributes, and for a layer's presence in [JOSM's ImageryCompare](https://josm.openstreetmap.de/wiki/ImageryCompare).

Any other ideas?

- [x] Check for duplicate attributes
- [ ] Check whether layer is present in JOSM's [JOSM's ImageryCompare](https://josm.openstreetmap.de/wiki/ImageryCompare).
- [x] Check for missing attributes - do this in code or by tightening JSON Schema
   - [x] there should be an attribution
   - [x] for types that depend on it, there should be projections, min and max zooms
       - [x] but they shouldn't be present in places where they don't make sense, e.g. a tms with projections
   - [x] license should be present
- [ ] #211 Use GeoJSON linter to test geojson input
- [x] check for url arguments - a wms should have {proj}, a tms should have {x}, etc.
- [ ] check for useless redundant projections, like josm's ImageryCompare does
- [x] there should be a geometry and a country, unless the layer is global

**Checks that require making requests to the services**
- [ ] #425 Check for interesting EPSG:3857-like projections
- [ ] Check whether layer returns images at all - see all the borken layers #425 found
    - for wms we can just request the bbox
    - for tms [maybe this?](https://gist.github.com/tucotuco/1193577)
- [ ] Check whether http resources are also available over https
- [ ] Check whether max_zoom+1 or min_zoom-1 return images
- [ ] Check whether trailing slashes and www prefixes can be removed

Marc-marc-marc:
also check for duplicate attribute like https://github.com/osmlab/editor-layer-index/commit/02ae97a259ca6f17c4ba50565424a70d7e75cb6d

grischard:
Unfortunately, we can't use json schema to check for duplicate attributes. @handrews from the JSON schema project very kindly took the time to explain this to me.

Presumably to avoid requiring parsers to keep track of which properties they've already seen, JSON allows this redefinition, but states that the resulting behavior is undefined. The most common behaviour is to parse JSON objects into data structures with unique keys.  So by the time JSON Schema sees the data, the parser has already decided whether to raise an error, or just silently drop one or the other property definition.

JSON Schema can't tell what has happened, as JSON Schema does not consider the original input text.  So there is no way to constrain this in JSON Schema.

[...]

PR #510, opened 2018-06-13
Check.py improvements

grischard:
Partially addresses #428 

Usage improvements:

* Show progress bar instead of dots
* Add -v argument to increase verbosity
* Colour output and better logging

Check improvements:

* Report embedded icons. Disembedding all icons would save 578.54 KB - worth it on a 1.71 MB file, I think!
* Check for missing license_url - see #439
* Check for wms/tms tokens in URLs - we've had some missing `{proj}` before.
* Check for either `default` or geometry polygon. Many layers have neither, and will never be shown.


grischard:
The build is failing because of the missing geometries.

grischard:
As part of this, I'll mark all the layers in `sources/world` as `default`. @bhousel I think that you intentionally blacklist many of these layers in iD, do you want to preemptively add them?

Heads up to @bryceco @simonpoole too.

simonpoole:
@grischard we currently only use the default attribute in sorting, so likely no big impact. I'm not that convinced that it really has useful semantics, or if it is suppose to have them, then maybe they have to be spelled out more clearly.

bhousel:
> as part of this, I'll mark all the layers in sources/world as default. @bhousel I think that you intentionally blacklist many of these layers in iD, do you want to preemptively add them?

I'm not sure what you mean by this?  I just did a search and I don't think iD is using the `default` property, so feel free to change it..

grischard:
Ok, I think I'm happy with this and will merge it.

Issue #480, opened 2018-05-02
Add Berlin aerial photography provided by the city of Berlin


tordans:
The city of Berlin has images that are a lot better than what Bing and such provides. It would be aweseome to have them as a layer in iD Editor and such. Ideally as the default layer for that region.

There are two data layers, I think both are interesting. Layer one should have priority and be the default image layer when opening iD Editor in Berlin. Layout two could be an optional layer (like "Esri" vs "Esri (Clarity)").

## URL of the imagery

This is all I know ATM

Layer 1:
- City Website: WMS: https://daten.berlin.de/datensaetze/digitale-farbige-orthophotos-2017-dop20rgb-wms-1 / Atom: https://daten.berlin.de/datensaetze/digitale-farbige-orthophotos-2017-dop20rgb-atom
- City Map Tool: http://fbinter.stadt-berlin.de/fb/index.jsp?loginkey=zoomStart&mapId=k_luftbild2017_rgb@senstadt
- There is a project by [OKFN DE](https://github.com/okfde) that provides easier access to this data
  - https://codefor.de/projekte/2018-04-09-fis-broker.html
  - https://github.com/jochenklar/berlin-aerial
  - https://tiles.codefor.de/ (choose "Geoportal Berlin / Digitale farbige Orthophotos 2017 (DOP20RGB)" from the right)
     ``` https://tiles.codefor.de/berlin-2017/{z}/{x}/{y}.png ```

- This tool shows a comparison of images for Berlin: https://mc.bbbike.org/mc/?lon=13.377792&lat=52.516233&zoom=17&num=4&mt0=mapnik&mt1=berlin-historical-2016&mt2=esri-satellite&mt3=berlin-historical-2016i&marker=

Layer 2 Infrared:
- City Website: https://daten.berlin.de/datensaetze/digitale-color-infrarot-orthophotos-2016-dop20cir-wms
- City Map Tool: http://fbinter.stadt-berlin.de/fb/?loginkey=showMap&mapId=k_luftbild2016_cir@senstadt
- https://tiles.codefor.de/ (same as above)
   ``` https://tiles.codefor.de/berlin-2016i/{z}/{x}/{y}.png ```

## Two letter country code
DE

## Desired layer name and attribution

Layer 1:

```json
    "properties": {
        "attribution": {
            "required": true,
            "text": "Geoportal Berlin / Digitale farbige Orthophotos 2017 (DOP20RGB)",
[...]

PR #515, opened 2018-06-28
Change tile server for berlin images

tordans:
This is a follow up to https://github.com/osmlab/editor-layer-index/issues/480.

It is awesome to have the Berlin aerial photography right in iD now.
However, the image server by the city seems to be a bit slow.

# Speed comparison

| City | OKFN |
|---|---|
| bildschirmfoto 2018-06-28 um 19 21 06 ~6s waiting, [URL](https://fbinter.stadt-berlin.de/fb/wms/senstadt/k_luftbild2017_rgb?FORMAT=image/jpeg&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=0&STYLES=&SRS=EPSG:4326&WIDTH=256&HEIGHT=256&BBOX=13.469238281250004,52.46793278868702,13.469581604003897,52.46814194225074)  |  bildschirmfoto 2018-06-28 um 19 21 16 ~37ms waiting, [URL](https://tiles.codefor.de/berlin-2017/18/140800/85992.png)  |

The OKFN source is based on https://tiles.codefor.de/ / https://codefor.de/projekte/2018-04-09-fis-broker.html

Note: The 2011 data are not available with tiles.codefor, so I left them be.

# Todos

- [ ] Can someone validate that this changes will work?
- [ ] I wonder if we need the ` "required": true` attribute for the `attribution` part, [other datasets have it](https://github.com/osmlab/editor-layer-index/search?l=JSON&q=required); what does it do?

### DIFF
--- a/sources/europe/de/Berlinaerialphotograph2011.geojson
+++ b/sources/europe/de/Berlinaerialphotograph2011.geojson
@@ -2,7 +2,7 @@
   "type": "Feature",
   "properties": {
     "id": "Berlin-2011",
-    "name": "Berlin aerial photograph 2011",
+    "name": "Berlin aerial photography 2011",
     "type": "wms",
     "url": "https://fbinter.stadt-berlin.de/fb/wms/senstadt/k_luftbild2011_20?FORMAT=image/jpeg&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=0&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
     "start_date": "2011",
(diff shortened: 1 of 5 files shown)


bhousel:
> Can someone validate that this changes will work?

These changes look good to me - the best way to test is to open iD and paste the url into the "custom" .  I tried this for the 2017 layer and it does load much faster.

>   I wonder if we need the "required": true attribute for the attribution part, other datasets have it; what does it do?

I think it just means that an editor is required to show the attribution line.

bhousel:
I switched from `{z}` to `{zoom}`,  removed the `available_projections` (not applicable for tms sources) and merged.

PR #516, opened 2018-06-28
Add Berlin infrared

tordans:
This is a follow up to https://github.com/osmlab/editor-layer-index/pull/515 to add another feature-layer.

You can see a preview at https://tiles.codefor.de/ " Geoportal Berlin / Digitale Color-Infrarot-Orthophotos 2016 (DOP20CIR)"


bhousel:
I switched from `{z}` to `{zoom}`,  removed the `"best": true` and merged.

PR #517, opened 2018-07-01
Added an entry for ACTmapi Imagery 2018

FrakGart:
Adding new ACTmapi Imagery. Just a few points/questions:

- There is currently no entry for this data set on the ACT's Geospatial Catalogue. There is a notice as to the copyright and CC BY 4.0 licence on the map information panel here:http://app.actmapi.act.gov.au/actmapi/index.html?viewer=basic. Will have to add a link to the Geospatial Catalogue at a later stage.

- I haven't changed the best=true in ACTmapi-Imagery.geojson on the assumption that the logic would be to choose newer imagery over older.

- I've used http in the link to the WMS service. Not sure if your policy is to use https for everything or just references to web-sites.


FrakGart:
Can somebody help with squishing these two commits together? My Git foo is weak...

grischard:
Hello, thanks for the PR! Yes, we'll squish everything, no worries.

We do indeed prefer https if it is available.

http://app.actmapi.act.gov.au/terms.html is indeed CC-BY, but we need [an additional waiver](https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/) to be fully compatible. [There's an Australian example here](https://osmlab.github.io/editor-layer-index/sources/australia/au/wa/MainRoadsWA.pdf).

FrakGart:
license_url already points to the waiver on the wiki.

grischard:
Indeed, my bad! Thank you for this PR.

FrakGart:
No worries.

andrewharvey:
Thanks @FrakGart !

Klumbumbus:
FYI, 2017 and 2018 ACT Imagery disabled in JOSM, wms not working atm. https://josm.openstreetmap.de/wiki/Maps/Australia?action=diff&version=43

andrewharvey:
Thanks @Klumbumbus please see #590, as far as I tested that URL still worked in JOSM.

PR #526, opened 2018-07-29
add imagery from Geoportal DF

willemarcel:
We achieved permission to use the imagery from the [GeoPortal Distrito Federal](http://www.geoportal.segeth.df.gov.br) of Brazil: https://wiki.openstreetmap.org/wiki/Pt:GeoPortalDFBrasilia


grischard:
That's fantastic news!

grischard:
Quick review:

* `overlay: false` isn't necessary.

* The positions in the polygons can be simplified to six or less decimal points.

* The description should describe the layer, not the geoportal. 

* A website and a link to an icon would be nice.

grischard:
Hi @willemarcel ! Do you think you will have time to change these so that they can be merged? The description in particular should be corrected.

willemarcel:
@grischard done! about the geometry, every layer has a bit different coverage, so I drew the polygon of each of them. If you prefer to use a simpler polygon, you can simplify that one: https://github.com/osmlab/editor-layer-index/blob/88c852dc1d6b616076e18e62447036b365d3bd52/sources/south-america/br/GeoPortal_DF_2015_imagery.geojson

grischard:
Thanks, merged!

erickdeoliveiraleal:
It's not working anymore, maybe we need to update the url, there are new image layers too:
Capturar3

erickdeoliveiraleal:
https://www.geoservicos1.segeth.df.gov.br/arcgis/rest/services/Imagens

grischard:
@erickdeoliveiraleal would you like to try doing a pull request? Let me know if you need a hand.

erickdeoliveiraleal:
@grischard sorry, but I don't know how to do it

PR #527, opened 2018-07-29
Sweden: Fix licence_url's and add i18n for some sources

magol:


NKAmapper:
Thank you for updating the URLs.
You may want to add CC0 also for the Economic maps, including in the attribution text.

magol:
@NKAmapper 
Thank you to notified me about that. I have fixed it now.

magol:
I also notice that the file https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fi/Nordicsnowmobileoverlay.geojson is used in Sweden to, even if it is marked for use in only Finland. So I added `i18n `to it too.

It is little problematic when a source is used across several countries, it was a bit difficult to find this file. It should be a text file (or something) in the folder for each country that lists all sources that concern the country, at least those sources that are not in the country's own folder.

grischard:
If a file is used in more than one country, I'd just put it in `europe` and remove the `"country_code": "FI",` in the json.

magol:
@grischard 
The folder `Europe` may not be the best place as it does not cover the whole of Europe, but it is significantly better than the `fi` folder. So I've moved it now.

grischard:
Looks perfect! The folders aren't a strict hierarchy, just to help the maintainers and contributors find things.

Issue #522, opened 2018-07-25
Support different tile sizes


bhousel:
WMS sources have had this for a while, but we are starting to see TMS sources that support 512px instead of (or in addition to) 256px tiles.  See https://github.com/openstreetmap/iD/issues/5179

I'd like to add an optional `available_tilesizes` property that takes an array of acceptable sizes.

`"available_tilesizes": [256, 512]`

If `available_tilesizes` is not specified, the default of 256 is assumed.

(edit: renamed to `available_tilesizes` to be more like `available_projections`)

grischard:
Sounds good to me!

stoecker:
A single WMS or TMS URL can only support 1 tilesize, so an array has no sense. Why not simply copy the already existing JOSM feature to specify the size instead of inventing something new?

bhousel:
> A single WMS or TMS URL can only support 1 tilesize, so an array has no sense. 

We could do a single size.  This is what TMS services need.  In the case of WMS, it's really more of a maximum size, and it's usually set by the server to something much higher than we'd probably use in practice (like 4000px)

> Why not simply copy the already existing JOSM feature to specify the size instead of inventing something new?

Sure that sounds good - can you link to the documentation on this feature?  I'm not familiar with what JOSM does.

stoecker:
https://josm.openstreetmap.de/wiki/Maps#TileMapServicesTMS

bhousel:
> https://josm.openstreetmap.de/wiki/Maps#TileMapServicesTMS

Thanks @stoecker .. how does JOSM decide what width/height to request from WMS servers?

Issue #339, opened 2017-05-29
Add Orthophotograph of Helsinki


theel0ja:
City of Helsinki's orthophotograph.

## URL of the imagery
https://www.avoindata.fi/data/en/dataset/helsingin-ortokuva-2014

WMS server: http://kartta.hel.fi/ws/geoserver/avoindata/wms
## Two letter country code
FI

## Desired layer name
Orthophotograph of Helsinki

## License
CC BY 4.0

## Attribution (text, URL, required?)
"Helsingin ortokuva 2014 by Helsingin kaupungin kiinteistövirasto is licensed under a Creative Commons Attribution 4.0 International License."

## Icon
(optional)

## Boundary polygon
(if possible, provide a boundary polygon, e.g. as a GeoJSON or OSM file.)

theel0ja:
`waitfor - license permission` should be added, too.

NKAmapper:
@theel0ja : The orthophotos provided by the Helsinki region seems to have a higher quality than the national MML orthophotos. They are open under the CC BY 4.0 license and distribution is actively encouraged: https://hri.fi/en_gb/helsinki-region-gis-data-freely-at-your-disposal/.

Would you mind asking for confirmation in an e-mail to hri.fi that their data is used in OSM?
Your request would need to mention that OSM is using the ODbL license and that attribution will be made at the https://wiki.openstreetmap.org/wiki/Contributors#Finland page.

I can then help adding the orthophotos for the region (Helsinki, Espoo, Vantaa, Kauniainen) to ELI and JOSM.

NKAmapper:
#340

grischard:
Thanks @theel0ja! As @NKAmapper rightly said, we need a confirmation that this data can be used in OSM. Unfortunately, CC-BY isn't fully compatible with OSM's ODbL, as that blogpost you linked to explains. Once you have a confirmation, I'll be very happy to add this.

theel0ja:
@NKAmapper What should I ask in the email?

NKAmapper:
This is the format we have used in Norway for most of our requests:

"I am one of many contributors to [OpenStreetMap](https:openstreetmap.org), a free, editable map of the whole world that is being built by volunteers and released with an open-content license.

We have seen that [Helsinki Region Infoshare](https:hi.fi) has released orthophotos and other open datasets under the CC BY 4.0 license. We would like to get your confirmation that these datasets may be used by OpenStreetMap.

[...]

PR #532, opened 2018-08-08
Add orthophotos for Helsinki region

NKAmapper:
This PR adds orthophotos for the Helsinki region, including Espoo and Vantaa

In some areas they are better than the existing national orthophotos from MML, so both have been given a "best" attribute.

Please note that tile size is 512 for this wms (tile_size).

Permission: [Link](https://forum.openstreetmap.org/viewtopic.php?pid=710304#p710304)


grischard:
See #522 for the `tile_size` attribute.

grischard:
But if it's a wms, I'm not sure I understand the `'tile_size'` attribute?

We don't have the `'valid-georeference'` attribute in ELI either. Maybe we should, but that's probably for another PR.

Can you add https://storage.theel0ja.info/imgup3/Screen%20Shot%202018-08-07%20at%2019.15.29.png to the `sources/europe/fi` directory and link to it like we did on https://osmlab.github.io/editor-layer-index/sources/australia/au/act/ACTmapi-Imagery.PDF ?

I'm happy merging Espoo and Vantaa in the same PR too.

NKAmapper:
Ok, done. Please ignore tile size, my mistake.

Espoo, Vantaa and Helsinki are already included in this PR, so no additional imagery needed (one combined wms for all three cities).

grischard:
Thank you!

NKAmapper:
#339 may be closed now :)

PR #536, opened 2018-08-14
Added layer Japan GSI Kiban 2500

higa4:
Added layer Japan GSI Kiban 2500 which is already in use manually on iD and JOSM.


sources/asia/jp/JapanGSIKiban2500.geojson
```diff
@@ -0,0 +1,206 @@
...
+    "type": "tms",
+    "url": "http://www.finds.jp/ws/tmc/1.0.0/KBN2500FN-900913-L/{zoom}/{x}/{y}.png",
+    "country_code": "JP",
+    "license_url": "https://wiki.openstreetmap.org/wiki/GSI_KIBAN",
+    "attribution": {
+      "url": "http://www.finds.jp/siteinfo/c_tou.html.ja",
+      "text": "GSI KIBAN 2500"
+    },
+    "max_zoom": 18,
+    "min_zoom": 5
```
tyrasd: Is the zoom range 5 to 18 really correct? In a quick test, I only get fully transparent tiles as result for requests at zoom 14 or lower (example: http://www.finds.jp/ws/tmc/1.0.0/KBN2500FN-900913-L/14/14484/6500.png). Also I get good results for zoom 19 (example: http://www.finds.jp/ws/tmc/1.0.0/KBN2500FN-900913-L/19/463488/208060.png) and maybe even higher. Can you please double-check these values?

higa4: Sorry. I found zoom level 15-20.
http://www.finds.jp/ws/tmc/1.0.0/KBN2500FN-900913-L/
http://www.finds.jp/tmc/layers.html.ja

grischard:
Thank you!

PR #539, opened 2018-08-18
Add new images for Berlin 2018, change source for 2015 and add exact dates

tordans:
Thanks to https://github.com/jochenklar/fis-broker/issues/1#issuecomment-414062058 and the city of Berlin we have new data for 2018.

The licence for 2018 is the same as for the other years, so the terms that are linked in the json still apply.

As a side note: Most of the 2018 images are from 2018-04-07, the southern part of the data is from 2018-03-19. See image at https://github.com/jochenklar/fis-broker/issues/1#issuecomment-413660394

Also, the geojson is the same as for 2017, even though it does not fit perfectly with the available data as seen in the image in https://github.com/jochenklar/fis-broker/issues/1#issuecomment-414062539 http://fbarc.stadt-berlin.de/FIS_Broker_Atom/DOP/Blattschnitt2x2km.gif. However IMO this could be made public as is and maybe modified later.

**Merging:**

I made small commits do make documentation easier, but I guess squashing them into one would be a good idea.

### DIFF
--- a/sources/europe/de/Berlinaerialphotograph2015.geojson
+++ b/sources/europe/de/Berlinaerialphotograph2015.geojson
@@ -3,17 +3,11 @@
   "properties": {
     "id": "Berlin-2015",
     "name": "Berlin aerial photography 2015",
-    "type": "wms",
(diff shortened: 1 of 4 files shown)


grischard:
Danke!

PR #540, opened 2018-08-19
virginia wms layers

jonahadkins:
building footprints, property lines, and imagery!


grischard:
Thanks for your PR! Could you please check [why the tests are failing](https://travis-ci.org/osmlab/editor-layer-index/builds/417998642?utm_source=github_status&utm_medium=notification), and include a license?

rivermont:
Unnecessary `"type": "FeatureCollection'` wrapping each `Feature` in the json files. Once removing that it should work.

jonahadkins:
hi 👋 - i've included the license and overlay =`true` for 2 of the layers

grischard:
Great, thank you!

Klumbumbus:
Are these entries working for you? I only get 404 errors for all 3 sources at all zoom levels.

Klumbumbus:
I think the file extension is missing at the end of the URLs

grischard:
Yeah, for example https://gismaps.vita.virginia.gov/arcgis/rest/services/VA_Base_layers/VA_Building_Footprints/MapServer/tile/8/12630/9241 returns a png for me. Not for you?

Klumbumbus:
I see a png with building footprints at your URL. But why is it at z8? At z8 you would see the whole state, not buildings. Is there a offset or reversed function in the zoom parameter?
When I zoom in to Richmond I get e.g. https://gismaps.vita.virginia.gov/arcgis/rest/services/VA_Base_layers/VA_Building_Footprints/MapServer/tile/17/50771/37342

jonahadkins:
im guessing i need to update something here? i used [San Juan County](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/north-america/us/wa) as a template since it comes from AGOL as well.

grischard:
Ah, so it looks like the virginia tile servers have their own personal url scheme? Great.

grischard:
So something like https://gismaps.vita.virginia.gov/arcgis/rest/services/VA_Base_layers/VA_Building_Footprints/MapServer/WMTS/tile/1.0.0/VA_Base_layers_VA_Building_Footprints/default/GoogleMapsCompatible/17/50771/37342.png should work but it just 404s for me.

jonahadkins:
bummer. i usually use this [tilefy app](https://github.com/JasonSanford/tileify-ags-proxy) to provide the urls to add in the custom background dialog in iD - could we use [these links](https://github.com/jonahadkins/commonwealth/blob/master/agol2osm.md) instead?

grischard:
Hmm, ideally it'd do a redirect, but I guess we could do that. Can you submit a new PR?

Klumbumbus:
The original URLs were added in JOSM (with `{zoom-7}`) in case you want to replace the tilefy URLs here.
https://josm.openstreetmap.de/wiki/Maps/USA?action=diff&version=121

grischard:
Oh dear, who thought that was a good idea? I'm not even sure the editors would support that.

simonpoole:
Well it seems to be a classical JOSM one off, I don't quite see why we should complicate parsing of the placeholders for exactly one use case.

PR #543, opened 2018-08-28
New service name for OSMI multipolygon

NKAmapper:
The correct name now is "Areas"


NKAmapper:
Sorry, that was only correct for the OSMI wms, not for the tile service here.

PR #544, opened 2018-08-28
Add vgin sources (new links)

jonahadkins:


grischard:
Thanks! Looks ok to me if those conflicts are resolved.

jonahadkins:
dunzo 👍

grischard:
It now says:

`ERROR:root:Error in sources/north-america/us/va/VGINBuildingFootprints.geojson : {z} found instead of {zoom} in tile url`

jonahadkins:
re-dunzo 👍 (soooory for all the issues)

grischard:
Looks perfect!

PR #545, opened 2018-08-28
https for some sources and link update

Klumbumbus:


grischard:
I had Luxembourg as http because https was significantly slower a while ago. No idea if that's still the case.

Klumbumbus:
I just tested Ortho Latest and I don't notice a perfomance difference between http and https. Both a pretty fast for me.

PR #546, opened 2018-08-28
Added Balatonőszöd.

urbalazs:
Hi @Klumbumbus 
Here is a new area for Balatonőszöd, Hungary.


grischard:
@Klumbumbus you know about this? :)

What is this @urbalazs ? It expands the polygon?

urbalazs:
Hello @grischard 
A new town has been added to the existing ortophoto set, that's why I added the new polygon.

grischard:
Nagyon köszönöm!

Klumbumbus:
The new shape is not closed. The last point must be the same like the first point. (See red line in screenshot.)

There is another new image west of the new one on high zooms too. Can you add the shape for this too?

grafik

urbalazs:
The new part in the west side is also new for me, it was added just today. I fixed the previous shape, and added the new one in https://github.com/osmlab/editor-layer-index/pull/549

Klumbumbus:
Thanks.

PR #547, opened 2018-08-29
Added ortophotos for Lake Balaton, Hungary.

urbalazs:
1940 geo-tagged photography from [Balaton Limnological Institute](http://www.bli.okologia.mta.hu/).


grischard:
The `permission_osm` attribute can't be an url. It's documented at https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L58 .

urbalazs:
Fixed, thanks!

PR #548, opened 2018-08-29
Add apikey attribute

simonpoole:
This adds support for an "apikey" attribute (so that applications can filter these out if they don't want to bother with them and indicate how the key should be supplied) and adds a corresponding placeholder to the documentation and to the two Thunderforest layers as examples.

Note the downside of doing this is that it is not actually backwards compatible aka the Thunderforest layers will likely stop displaying for any app that doesn't support the new placeholder.


grischard:
I've replied in that other issue. Not that happy about introducing yet another url parameter.

simonpoole:
Just an implementation "idea" triggered by https://github.com/osmlab/editor-layer-index/issues/602

API keys can be stored in the users preference key-value store and in principle accessed by multiple applications (I do this for the maproulette API key in Vespucci) which a) means the user would only have to enter a key once and b) would be able to use the corresponding layer in multiple different editors.

grischard:
@bryceco @bhousel what do you think, would your editors get a key, or ask the user for the key?

@Klumbumbus any idea how JOSM handles this?

Should we find non-technical solutions to this, like asking for API keys for all of OSM?

Klumbumbus:
>@Klumbumbus any idea how JOSM handles this?

I'n not aware of any mechanism in JOSM for user defined api keys for sources in the JOSM wiki.

grischard:
CC'ing @kevinbullock.

The UX of this has to work, otherwise we'll scare newbies. The editors should be able to retrieve an anonymous token in the background.

grischard:
I've merged this - heads up @bhousel @simonpoole @bryceco, expect {apikey} to appear in URLs. I'd recommend using the layer id as a key.

bhousel:
> I've merged this - heads up @bhousel @simonpoole @bryceco, expect {apikey} to appear in URLs. I'd recommend using the layer id as a key.

Thanks! I'll make the changes on my end too.  Just to be sure, will the thunderforest layers work if the apikey is left empty, or should I request for one for iD? cc @gravitystorm 

Update:  Seems unhappy without one
Screenshot 2020-03-03 12 12 09

andrewharvey:
Looks like the feed now gives an error/warning in JOSM, looks like https://josm.openstreetmap.de/ticket/18726 may address it.

gravitystorm:
> or should I request for one for iD? cc @gravitystorm

It's best if you sign up for a (free) Thunderforest account in order to get an API Key for iD, like the other editing apps.

PR #549, opened 2018-08-29
Added Balatonszemes.

urbalazs:
A new ortophoto from Balatonszemes, Hungary.


rbuffat:
@urbalazs This source seems to be down for some time. Do you know if its URL has changed or if it should be removed?

PR #552, opened 2018-09-03
Add support for HTTP header indicating invalid tiles

simonpoole:
This adds fields that allow specifying a HTTP header and associated value that indicates an invalid tile, this is mainly useful for layers that return error tiles instead of 404s. This mirrors a similar facility of JOSM.

Note, this could naturally be modelled as a JSON object or an array instead of two string fields, no particular preference on my behalf.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -104,6 +104,14 @@
                     "type": "string",
                     "pattern": "^\\d\\d\\d\\d(-\\d\\d(-\\d\\d)?)?$"
                 },
+                "no_tile_header": {
+                    "description": "HTTP header to check for information if the tile is invalid",
+                    "type": "string"
+                },
+                "no_tile_value": {
+                    "description": "Value for the no tile header indicating that the tile is invalid",
+                    "type": "string"
+                },
                 "overlay": {
                     "description": "'true' if tiles are transparent and can be overlaid on another source",
                     "type": "boolean",
--- a/schema.json
+++ b/schema.json
@@ -106,14 +106,25 @@
                 },
                 "no_tile_header": {
                     "description": "HTTP header to check for information if the tile is invalid",
-                    "type": "string"
-                },
-                "no_tile_values": {
-                    "description": "Values for the no tile header indicating that the tile is invalid",
-                    "type": "array",
-                    "items": {
-                        "type": "string"
-                    }
+                    "type": "object",
+                    "minProperties": 1,
+                    "maxProperties": 1,
+                    "patternProperties": {
+                        "^.*$": {
+                            "anyOf": [
+                                {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string"
+                                    }
+                                },
+                                {
+                                    "type": "null"
+                                }
+                            ]
+                        }
+                    },
+                    "additionalProperties": false
                 },
                 "overlay": {
                     "description": "'true' if tiles are transparent and can be overlaid on another source",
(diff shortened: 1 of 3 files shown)


simonpoole:
JOSM allows alternatives in the value `<no-tile-header name='ETag' value='"067736a547cafe90014b4e59b6510abe"|"ee1f6802b0234046b553cbbc672ac7d9"|"9f5a2f1d7cc131e58befc2052c71c827"' />` that could be modelled as a JSON array if we wanted to (the sole example is Mapboxs sat layer).

simonpoole:
I'll refactor this to use an array for the values, seems to make more sense.

simonpoole:
I haven't changed the mapbox entry (yet), as I haven't been able to find an area for testing (aka where Mapbox is returning an invalid tile instead of a 404).

simon04:
What about using an object for `no_tile_header` mapping header keys to values. This makes the syntax more concise (and permits checking different header keys).

```json
"no_tile_header": {
    "X-VE-Tile-Info": "no-tile"
}
```

```json
"no_tile_header": {
    "ETag": ["067736a547cafe90014b4e59b6510abe", "ee1f6802b0234046b553cbbc672ac7d9", "9f5a2f1d7cc131e58befc2052c71c827"]
}
```

grischard:
Hmm, so this can't be automatically detected by requesting zoom level 404?

simonpoole:
@simon04 it is more concise but now requires that we decide if we allow null values, and allowing the extra complication of multiple headers makes the on disk representation (aka DB) more complicated for the benefit of 0 use cases.

Not against making the change, but then restrict it to one entry.

simonpoole:
@grischard the layers in question don't return 404s (which is silly, but the way it is).

grischard:
@simonpoole requesting zoom level 404, which will never exist, would return whatever a non-existing tile looks like. For example, https://a.tile.openstreetmap.org/404/134/83.png returns a http 404 error.

simonpoole:
@grischard it might return what a tile request inside of the max zoom but not available returns, it might not. The other issue is that you then have to create a hash or whatever on the tile you got back and calculate that for every tile you download, great if you have unlimited CPU resources, less if your don't. And naturally there are ways to workaround all of this, but it seems to be awfully expensive compared to simply indicating it in the configuration (for the three cases we know of).

grischard:
It would be interesting to hear what the authors of other editors think. Ping @bhousel?

bhousel:
> It would be interesting to hear what the authors of other editors think. Ping @bhousel?

I don't think iD can check the network header without doing a separate request.   We load the images directly into html `` tags and only know if the load failed if the error handler gets called.  

(I'm ok if you want to add it, but we wouldn't use it for anything.)

grischard:
I'm actually cool with merging this if we can do simon04's proposal.

simonpoole:
> 
> 
> I'm actually cool with merging this if we can do simon04's proposal.

Didn't I already do that in https://github.com/osmlab/editor-layer-index/pull/552/commits/729e5d79f3bb1d4f058aab4e9204a2a77ae8add5 ?

grischard:
You did 👍

don-vip:
We now compare the no tile headers, see https://josm.openstreetmap.de/ticket/17730
Spotted one difference for ESRI.

grischard:
I love the conclusion: "So ELI is wrong."

One more thing we need to fix then :)

grischard:
Fixed in 7450181

PR #554, opened 2018-09-08
Align HRI entry to JOSM one

don-vip:
This aligns HRI entry (#339) to [modifications made](https://josm.openstreetmap.de/wiki/Maps/Finland?action=diff&version=30) by @NKAmapper on JOSM wiki.


grischard:
Thanks!

PR #557, opened 2018-09-15
Better attribution URL for HRI-orto

NKAmapper:
Alignment with JOSM.
Attribution URL updated to avoid confusion regarding owners.


don-vip:
I already created #556

NKAmapper:
Ah, thanks

PR #559, opened 2018-09-25
Improve discoverability of PICC

tbertels:
+"best": true,
This is the second most useful source for Wallonia (house numbers, shapes and street names)

### DIFF
--- a/sources/europe/be/BE_SPW_PICC_WMS.geojson
+++ b/sources/europe/be/BE_SPW_PICC_WMS.geojson
@@ -499,6 +499,7 @@
             "EPSG:3857",
             "EPSG:4258"
         ],
+        "best": true,
         "country_code": "BE",
         "i18n": true,
         "icon": "https://osmlab.github.io/editor-layer-index/sources/europe/be/SPW(allonie)2009-2010aerialimagery.png",


bhousel:
I'm pretty sure you don't want to do this..
1. Because `SPW_ORTHO_LAST` is already marked as `"best": true` for Belgium
2. Because `SPW_PICC` is not even aerial imagery at all - yes it has some building footprints and housenumbers, but... that's not what "best" is for.

You don't want a user opening the editor for the first time to see this:

screenshot 2018-09-25 09 35 39

grischard:
Doesn't Belgium have a new mapper welcome message thing going on? That message might be a perfect spot to recommend PICC.

tbertels:
> I'm pretty sure you don't want to do this..
> 
>     1. Because `SPW_ORTHO_LAST` is already marked as `"best": true` for Belgium
> 
>     2. Because `SPW_PICC` is not even aerial imagery at all - yes it has some building footprints and housenumbers, but... that's not what "best" is for.
> 
> 
> You don't want a user opening the editor for the first time to see this:

Yes, I do.
1. Since `SPW_ORTHO_LAST` is already marked as `"best": true` for Belgium, this would only be selected if the user changes the overlay.
2. `SPW_PICC` has all (or most of) the buildings, numbers and street overlays and names. This is for me the second best choice for Wallonia.
3. Per 1. a user won't open the editor with that overlay if the default overlay is as I think chosen alphabetically (so `SPW_ORTHO_LAST` should be selected).

However, ideally `"second_best": true` would be better if supported.

Here's how the overlay panel currently looks like:
image

For some reason, in French, the PICC one is at the top of the SPW ones:
image

There's the same problem for Flanders and Brussels (except for the difference of order between languages). Here's what it looks like for Flanders:
image

The two most useful overlays are (fortunately) highlighted, but everything between them is mostly useless.

Maybe should I open an issue about the possibility to set second best overlay(s)?

tbertels:
> 
> 
> Doesn't Belgium have a new mapper welcome message thing going on? That message might be a perfect spot to recommend PICC.

I don't know, I will check thanks!

tbertels:
> > Doesn't Belgium have a new mapper welcome message thing going on? That message might be a perfect spot to recommend PICC.
> 
> I don't know, I will check thanks!

There is: https://welcome.openstreetmap.be
But talking about the overlays in this general message may be a bit too much.

grischard:
There's more than one editor using ELI, and the way they interpret the `best` flag is, unfortunately, different. If we're going to change the logic for layer selection, then we should tackle the problem of some layers being 'best' in some places but not others, and provide a service that returns the best layer for the place you're looking at - it doesn't really make sense to download all of ELI and its hundreds of polygons.

PR #561, opened 2018-09-26
add au/vic/city_of_melbourne may 2018 imagery

andrewharvey:
See permission at https://wiki.openstreetmap.org/wiki/Data.melbourne.gov.au_explicit_permission


grischard:
Thank you! Could we have that wiki page as the license_url property actually?

andrewharvey:
Sure, done.

andrewharvey:
I've also added license, icon.

grischard:
Thanks!

PR #563, opened 2018-09-27
Add information on licence and OSM usage permission

simonpoole:
This commit adds information on the licence and the nature of the
permission to use the layer in OSM to the configurations for the global
aerial imagery mosaics.

As there is no fitting SPDX licence id for non open licences I've used
"COMMERCIAL" which seems to be fit well and is reasonably diplomatic.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -65,7 +65,7 @@
                     ]
                 },
                 "license": {
-                    "description": "The license for the imagery specified using a SPDX identifier",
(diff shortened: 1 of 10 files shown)


grischard:
Thanks!

grischard:
Do we have even vaguely relevant license URLs we could have to tick those off for #439? Something like https://wiki.openstreetmap.org/w/index.php?title=File:Bing_license.pdf for Bing.

simonpoole:
I believe there is something around for the ESRI and DG mosaics. Need digging out though.,

PR #570, opened 2018-10-17
Change orthography

KikeSan:


bhousel:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/openstreetmap/id-editor/dashboard/), so you should sign up for an account there to fix issues with the translations.

PR #573, opened 2018-10-23
New URL for Offshore WMS

NKAmapper:


grischard:
Thank you again @NKAmapper :)

PR #577, opened 2018-10-27
Add Thunderforest Outdoors layer

### LINKED ISSUE #551: One more for thunderforest
You have 2 types of maps for thunderforest : landscape and cycle

For me the best one is outdoors. An example with the 3 maps are enable here :
https://mc.bbbike.org/mc/?lon=-75.406002&lat=-8.963726&zoom=14&num=3&mt0=thunderforest-outdoors&mt1=thunderforest-landscape&mt2=thunderforest-cyclemap

On outdoors you have the altitude and you can see clearly the tracks/roads.
On landscape you can see nothing (no altitude, roads very badly)
On cycle roads are very good but the altitude are not visible

I don't want you to delete the 2 thunderforest you have, just to add outdoors

## URL of the imagery
https://{switch:a,b,c}.tile.thunderforest.com/outdoors/{zoom}/{x}/{y}.png

## Two letter country code
Sorry, I don't know waht you want : I think for every country ?

## Desired layer name
Thunderforest outdoors

## License
I think you need one, myself I have a ?apikey= to use it

Yes, thunderforest is part of OSM I believe ?

hikemaniac:
This PR adds the Thunderforest Outdoors layer and therefore closes #551.


grischard:
So this doesn't need an api key?

simonpoole:
To display without the watermark: yes.

hikemaniac:
It looks like we also don't have an API key for OpenCycleMap and Thunderforest Landscape.

grischard:
So this closes #551

We can add the {apikey} url param and privacy_policy_url from the other tf layers.

grischard:
Closed in ddb2695, thanks @hikemaniac

maro-21:
Does it cover the whole world? Because it doesn't work in my region, screen is black.

hikemaniac:
@maro-21 This is probably due to the API key not being set in the client application. You can instead test the URL https://{switch:a,b,c}.tile.thunderforest.com/landscape/{zoom}/{x}/{y}.png. When the client does not fill in the `{apikey}` template, the tiles can not be displayed because the URL becomes invalid.

hikemaniac:
~~This is actually really interesting, because I just tested all of the Thunderforest layers without the apikey parameter appended to the URL in iD (hand-pasting the URLs) and they are displayed without any watermark. @gravitystorm Is that intentional? Can the apikey parameter be omitted?~~ 

The above only seems to work in iD. I tested the same URL using another tool and the watermark is displayed. Strange.

maro-21:
I use iD and there's nothing I can do with the apikey. When I type it manually it works, but why Thunderforest Outdoors doesn't?

hikemaniac:
The `apikey` is either something that the application stores secretly or that the user has to manually fill in. In iD's case, some of the built-in layers (Maxar) come with API keys, so the user doesn't have to bother, but the same doesn't apply for this layer and  it's therefore broken in iD.
Hope this helps :blush:

hikemaniac:
I created a pull request to remove the layer from iD, since it is invalid without the parameter being filled in. See openstreetmap/iD#7559

PR #584, opened 2018-11-04
Add TIGER 2018 roads overlay

iandees:
Adds a more recent TIGER roads overlay.


bhousel:
Nice!  Looks better then the old data in the few places that I've checked 👍 
 
tiger 2018

PR #585, opened 2018-11-04
Remove/update some OSM US-hosted layers

iandees:
Some of the layers hosted by OpenStreetMap US have been broken or out of date for years. This removes or adjusts those layers.


grischard:
Thanks!

PR #587, opened 2018-11-05
Remove duplicate entry. Set Estonia ortho as local best

RM87:


grischard:
Hi RM87! Thanks for the PR. We still need that `max_zoom` though.

grischard:
Aitäh RM87!

Klumbumbus:
I just tested the wms and tms sources and for me the min zoom values are 15 for wms (instead of 14) and 6 for tms (instead of 14)

grischard:
Thank you, I've corrected the minzoom then. There's a tms? Is it any faster?

I also see that the polygon could be greatly simplified.

Klumbumbus:
>There's a tms?

Yes and it is already in ELI ;)

grischard:
Ahhhhhhhhh!

PR #594, opened 2018-11-29
update City of Melbourne imagery with CC BY waiver

andrewharvey:
the original permission was from before the OSMF published the CC BY waiver template


grischard:
Excellent, thank you!

Issue #210, opened 2016-11-11
manual sync with josm database


Klumbumbus:
I just wanted to make sure if it is ok to use your database to update/add imagery sources in the josm database with the goal to bring both databases up to date?

bhousel:
Yes!  We'd love your help with this 👍

Klumbumbus:
OK thanks, then I close this issue.

PR #600, opened 2018-12-06
Fix typo on string #210

kant:
? on string 224 & 228 & 233: ¿Contribuidores o Colaboradores?


grischard:
On line 250, you still have a single quote at the start.

Issue #55, opened 2013-12-23
MapBox terms URL


pnorman:
The current link for MapBox terms sends to an About Mapbox Streets page. 

> Mapbox Streets is licensed under our Terms of Service. OpenStreetMap contributors create and improve its map data, licensed under the ODbL. DigitalGlobe acquires and processes our commercial imagery. All other imagery is derived from public sources and processed by Mapbox.

The DigitalGlobe link is probably not what is relevant for the imagery given.

PR #601, opened 2018-12-06
Fix typo on string #55

kant:


grischard:
Your attention to detail is appreciated!

PR #603, opened 2018-12-12
australia -> oceania

andrewharvey:
Although many Australians consider New Zealand domestic, it's a bit unfair to put our friends in Papua New Guinea and New Zealand in the Australia region. Oceania is a better continental group.


grischard:
Thank you Andrew. 'down_under' wouldn't work? :)

PR #604, opened 2018-12-16
Point to MassGIS cache instead of the proxy

iandees:
The cached version has address data rendered with the polygons.


grischard:
So that maxzoom in the current version is a lie? LGTM otherwise.

iandees:
The [source map](http://massgis.maps.arcgis.com/apps/OnePane/basicviewer/index.html?appid=47689963e7bb4007961676ad9fc56ae9) stops at 19 and no new data is added after, so I thought 22 was a bit of a stretch and moved it to 20.

grischard:
Cheers!

PR #608, opened 2018-12-28
Add Stuttgart aerial imagery provided by Stadtmessungsamt Stuttgart.

johsin18:
I'm new to this project, and I could not really test this change (except for pasting the URL it into a WMS layer configuration in JOSM).  So I'm asking for a thorough review.


grischard:
Hello, and congratulations on your first PR!

I see that there are also tiles available, e.g. https://gis5.stuttgart.de/arcgis/rest/services/1_Base/Luftbilder_aktuell_c_EPSG3857/MapServer/tile/20/361034/551047

All editors support tiles, only some support WMS resources. If we're allowed to use the tiles, it would be great.

If we're only allowed to use the WMS, we should list all the projections that https://gis5.stuttgart.de/arcgis/services/1_Base/WMS_Luftbilder_aktuell/MapServer/WMSServer?request=GetCapabilities&service=WMS lists.

Thank you!

johsin18:
Hi grischard,

the authority providing the aerial imagery asks us to limit ourselves to WMS on EPSG:25832, as everything else might be discontinued mid-term.
https://lists.openstreetmap.de/pipermail/stuttgart/2019-January/000905.html

grischard:
What a pity. Yeah, we can integrate it, but AFAIK none of the clients that use ELI know how to reproject from 25832. So it would be a lot more than "wohl praktischer", it would be useable.

3857, on the other hand, can be used by iD, at least. I think we should include it, even though it produces "Transformationseffekte".

grischard:
@johsin18 would you like to add 3857 to the supported projections so we can merge this?

grischard:
Hoping this passes the checks

grischard:
Thank you @johsin18 !

PR #613, opened 2019-01-17
Create JapanGSIairphotoImagery.geojson

higa4:
Widely covered and/or newer than GIS ortho Imagery.


grischard:
Hi Higa4, thank you for your PR!

The polygon can be greatly simplified since the east coast of Japan is just water. The whole line from Hokkaido to Taiwan can be simplified to just one point in the middle of the Pacific.

Let me know if you'd like to try that, otherwise I'll merge it as it is.

higa4:
@grischard Thanks for your notice. Sorry, it seems hard to me because there are many islands to include toward east-southern pacific area underneath Tokyo, Japan.

grischard:
Thank you very much, it's merged 👍

higa4:
Thanks a lot!

PR #622, opened 2019-02-17
Add Japan GSI seamlessphoto Imagery

higa4:


grischard:
Thank you!

Are the other layers from https://maps.gsi.go.jp/development/ichiran.html interesting?

higa4:
Thanks a lot!
I think it's eoough for now. But I may create a PR later if someone needs.

PR #625, opened 2019-02-27
Juno GPS tiles for NYC

Zverik:
This adds a new overlay layer that looks exactly like the OpenStreetMap's GPS tile layer, except that it covers only New York City, has much more tracks (30-40 thousand) and only tracks from a single day, always at most two days old. It is updated daily. The license is CC-BY-SA 4.0 with an explicit permission to use for OSM.

See https://gps-tile.junolab.net/ for an overview and terms of use.

In a week or two we open in New Jersey, so it will have more data. Not sure if putting it in the `ny` directory is correct, but it certainly doesn't cover the whole of US.

I have also updated the `requirements.txt` file, since version 0.13.3 of the `transifex-client` has been pulled from pypi.


grischard:
Thank you!

If we have `"permission_osm": "explicit"`, could we please have a link to where this is being made explicit?

Zverik:
It's all in the metadata: https://gps-tile.junolab.net/about.html
I used the text from https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates

grischard:
Thank you!!

PR #627, opened 2019-03-10
Update host name for polska.e-mapa.net: Buildings

gjedeer:
See https://github.com/openstreetmap/iD/issues/6031


bhousel:
Thanks @gjedeer 👍

PR #628, opened 2019-03-17
Add a short description to all sources proxied by SOSM

simonpoole:
This just adds a one-liner pointing out that the sources are proxied by SOSM, otherwise the privacy policy link is confusing.


grischard:
Cheers!

PR #629, opened 2019-03-19
Belo Horizonte base map

tsmock:
I'm going through [https://wiki.openstreetmap.org/wiki/Dados_Abertos_-_Prefeituras_do_Brasil](https://wiki.openstreetmap.org/wiki/Dados_Abertos_-_Prefeituras_do_Brasil), which indicates that the sources on the page are under ODBL compatible licenses.

This is (so far) the easiest to add, except I still need to find the actual license. It appears that most of the others have PDF's instead (is there a way to add those to the index, even if the images have to be extracted and hosted elsewhere?).

I _don't_ speak Spanish/Portuguese, although I can _barely_ read and comprehend basic sentences in it.


naoliv:
https://wiki.openstreetmap.org/wiki/Dados_Abertos_-_Prefeituras_do_Brasil had an incorrect/incomplete statement (saying that everything listed there have data compatible with ODbL)

I have included a warning on the page, asked the Brazilian community to pay more attention about this and also asked to check the license/permission of the listed sites.

PR #635, opened 2019-03-23
Add VoGIS aerial imagery 2018 and general WMS endpoint

IdealChain:


grischard:
At that license URL, it says CC-BY 4.0, not 3.0.

Did we ever get a CC waiver to use those in OSM?

IdealChain:
Well, I've just added a new layer for the already existing imagery. They seem to have changed their license to CC BY 4.0 sometime in the last few weeks.

There is [this mail](https://wiki.openstreetmap.org/wiki/WikiProject_Austria/vogis#Nutzungsbedingungen) from 2012, which was a response to being explicitely asked about the attribution, as well as a [waiver for basemap.at](https://wiki.openstreetmap.org/wiki/Contributors#Austria), which incorporates data of all federal states. So I would say there is permission, however, I don't think that there is an up-to-date waiver explicitely for this imagery, so if that is a strict requirement, then you need to remove all VoGIS* files from this repo until there is one, as well as Tiris* and Vienna*, I guess.

grischard:
Sorry this got forgotten!

PR #636, opened 2019-04-03
Remove OpenInfraMap

russss:
OpenInfraMap has used vector tiles for a while now, and the old raster
tile renderer is now offline. Sorry!


grischard:
Thanks @russss!

PR #637, opened 2019-04-04
add new aerial photo city of Erlangen 2018, change URL for Erlangen 2016

rrzefox:
Because the city of Erlangen currently has no public WMS-server, the university of Erlangen kindly provides a TMS as a replacement.


grischard:
Thank you!

PR #639, opened 2019-04-07
Update orthophoto for Lyon - correct a non-loading layer

Ecologeek:
Add the Lyon orthophoto from the OSM-FR WMS server


grischard:
Merci beaucoup!

PR #640, opened 2019-04-07
ortho, dtm and dom for Kanton Zurich (wms)

rbuffat:
This PR adds the following three background layers for the Kanton Zurich, Switzerland, Europe:

- Orthophotos for the year 2018
- Hillshade of the digital terrain model of the year 2014
- Hillshade of the digital surface model of the year 2014

Each of the layers is only available as wms and is open (no attribution needed):
* [https://opendata.swiss/en/dataset/wms-digitales-hohenmodell-ogd](https://opendata.swiss/en/dataset/wms-digitales-hohenmodell-ogd)
* [https://opendata.swiss/de/dataset/wms-orthofotos-ogd](https://opendata.swiss/de/dataset/wms-orthofotos-ogd)

As I'm not aware how to test this locally this PR is unfortunately completely untested.


rbuffat:
Replaced the tms orthophoto layer of Canton of Zurich for the year 2014/15 and 2015/2016 using mapproxy.osm.ch with the direct wms.

I tested all new layers on a local instance of the iD editor.

simonpoole:
> 
> 
> Replaced the tms orthophoto layer of Canton of Zurich for the year 2014/15 and 2015/2016 using mapproxy.osm.ch with the direct wms.

This is really an unrelated change. In any case I would suggest removing the year as the ortho imagery (if they are still doing what they originally intended) is updated in a rolling fashion in any case. So it would better be named "latest" or something similar.

rbuffat:
> > Replaced the tms orthophoto layer of Canton of Zurich for the year 2014/15 and 2015/2016 using mapproxy.osm.ch with the direct wms.
> 
> This is really an unrelated change.

- The old "Kanton Zürich Orthophotos 2014/2015 10cm" does not work (wms based, not mapproxy.osm.ch, sorry for the confusion here)
- In my opinion, it makes sense having all three orthophoto layers using the same wms service having a similar config. Except if there are technical reasons that tms layers should be preferred. However, I don't mind reverting these changes.

>In any case I would suggest removing the year as this (if they are still doing what they originally intended) is updated in a rolling fashion in any case. So it would better be named "latest" or something similar.

It looks indeed that the newest orthophoto has just "ortho" as layer name, whereas the older layers have a year in the layer name. For attribution reasons, I think it would be good to include the year. I'm new to the OSM community, but looking at different imagery, I see that often the spatial resolution is included in the name. Metadata, including, e.g., start_date, end_date would need to be changed regardless if a new dataset is released.

grischard:
Danke sehr!

don-vip:
What's the difference between [OrthoZHWMS](https://wms.zh.ch/OrthoZHWMS?FORMAT=image/jpeg&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities) (the one we use at JOSM) and [OGDOrthoZH](https://wms.zh.ch/OGDOrthoZH?FORMAT=image/jpeg&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities) (the one you use here)? Capabilities are not the same (a few more layers in OGD), layers in common have different names.

grischard:
@rbuffat any idea?

don-vip:
I got the answer on Twitter and already switched JOSM: https://twitter.com/GeoKtZH/status/1115898444483375104

grischard:
Thanks @don-vip! @rbuffat can you please submit a new PR with the OrthoZHWMS url?

rbuffat:
> What's the difference between [OrthoZHWMS](https://wms.zh.ch/OrthoZHWMS?FORMAT=image/jpeg&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities) (the one we use at JOSM) and [OGDOrthoZH](https://wms.zh.ch/OGDOrthoZH?FORMAT=image/jpeg&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities) (the one you use here)? Capabilities are not the same (a few more layers in OGD), layers in common have different names.

@don-vip   I was not aware of the differences and used the OGD version because it has OGD in its name. After your comment, I wrote an email to ask for clarification but did not yet get an answer. It seems as they are more active on twitter than answering emails. 

> Thanks @don-vip! @rbuffat can you please submit a new PR with the OrthoZHWMS url?

@grischard My PR included only the OGDOrthoZH, respectively OGDLidarZH WMS.

> In any case I would suggest removing the year as the ortho imagery (if they are still doing what they originally intended) is updated in a rolling fashion in any case. So it would better be named "latest" or something similar.

@simonpoole  I created a watchdog to check for the wms services I added using travis: https://github.com/rbuffat/imagery_watchdog If the "ortho" layer changes it's a description from "Orthofoto ZH Sommer 2018 RGB" to a newer version the corresponding test will fail.

don-vip:
Sorry my post wasn't clear: OGD is the correct site to use. We were using the older one which is now deprecated and we switched to OGD. So the url are ok here, no need to change them.

simonpoole:
The PNGs generated by the WMS are extremely large compare

https://wms.zh.ch/OGDOrthoZH?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&LAYERS=ortho&FORMAT=image/png&CRS=EPSG:3857&WIDTH=512&HEIGHT=512&BBOX=935130.6040408462,6006268.80859258,935283.4780974165,6006421.6826491505

https://wms.zh.ch/OGDOrthoZH?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&LAYERS=ortho&FORMAT=image/jpeg&CRS=EPSG:3857&WIDTH=512&HEIGHT=512&BBOX=935130.6040408462,6006268.80859258,935283.4780974165,6006421.6826491505

So while it is a shame I suspect switching to JPG would make sense in this case.

rbuffat:
@simonpoole  fixed with https://github.com/osmlab/editor-layer-index/pull/652

simonpoole:
Thanks

PR #641, opened 2019-04-09
dsm/dtm hillshade layer for canton Solothurn

rbuffat:
This PR adds the following two WMS layers of the canton of Solothurn, Switzerland, Europe:
- digital surface model hillshade
- digital terrain model hillshade


simonpoole:
A note on this PR (which applies to the ZH one too): you should be adding the privacy policy URLs as these layers will have a higher likelihood of being used by people that care about such things than elsewhere.

grischard:
Thank you, and thank you @simonpoole!

PR #642, opened 2019-04-09
dsm/dtm hillshade layer for canton Bern

rbuffat:
This PR adds the following two WMS layers of the canton of Bern, Switzerland, Europe:

digital surface model hillshade
digital terrain model hillshade


grischard:
Ah, for mapping peaks and stuff? Thank you!

rbuffat:
@grischard  Digital terrain model hillshades are e.g. useful to see paths in forests: http://maps.zh.ch/s/riya50ah

As they are derived from Lidar point clouds they should also have less distortion compared to aerial images. Thus, digital surface model hillshades could be used to map buildings etc.

PR #643, opened 2019-04-12
Add Croatia TMS server for City of Knin aerial imagery

hbogner:


grischard:
Cool! What's the licence on these? Could you please add the licence properties?

hbogner:
I got written permission to use it for OSM

hbogner:
I tried, it fails. Now you take over!!!

hbogner:
OK, what license do you mean exactly?
Original data license or license to use our TMS?

grischard:
You can copy-paste from #644. `license` is the license for the imagery specified using a [SPDX identifier](https://spdx.org/licenses/), or 'COMMERCIAL'. In your case, since you got special permission, it's a link with `license_url` that would be useful.

andrewharvey:
@hbogner Could you also include a copy of or link to the written permission?

hbogner:
Here is the law on which the data was released, description of rights, description of license and closest license on that list... please advise what exact license to use....

hbogner:
License for this data is not "commercial" but some sort of "cc-by" or "open data" as it can be read here:
Law on which data was released: https://narodne-novine.nn.hr/clanci/sluzbeni/2013_02_25_403.html
Description of rights: http://data.gov.hr/o-portalu-cijeli
Description of license: http://data.gov.hr/open-licence-republic-croatia
Additional description: http://data.gov.hr/faq#5

It doesn't match https://spdx.org/licenses/ exactly so checks fail.

Also should I create osm wiki page with all this described and link to it for "license_url"?

grischard:
Ok, if you can just drop the 'license' property I'll merge it. The license url is enough.

grischard:
Čestitam!

hbogner:
Hvala :)

And now when I got the syntax right I'll add rest of the layers available for OSM.

don-vip:
The icon is not used?

hbogner:
I forgot to include the icon.
Will fix that with next layer and next pull request: https://github.com/osmlab/editor-layer-index/pull/663

PR #649, opened 2019-04-20
Remove eox.at as best layer

rivermont:
eox.at cloudless layer (#644) has the `best` attribute.  
I haven't checked anywhere, but in many places it is certainly not the preferred imagery.  
I believe it is also practice not to offer worldwide imagery as the best layer.

This Pull removes the `best` attribute from the "eox.at cloudless 2018" layer.


grischard:
Thank you! Closed in 76cb95e5cb43682412061e23bc58bf2207f437fb

PR #652, opened 2019-04-27
Switch from png to jpeg

rbuffat:
As recommended by @simonpoole switching from png to jpeg. 

+ Updated license_url for Kanton_Bern_dsm_hillshade_wms.geojson and Kanton_Bern_dtm_hillshade_wms.geojson. Same file, but directly hosted on a be.ch server and not geodienste.ch.


grischard:
Thank you!

don-vip:
The other license file is bilingual (german/french).

PR #654, opened 2019-05-05
bilingual (de/fr) version of licence url

rbuffat:
As suggested by @don-vip in https://github.com/osmlab/editor-layer-index/pull/652 it is better to use the bilingual version for the license url.


grischard:
Merci dir and @don-vip :)

PR #655, opened 2019-05-08
Add sources and license for DigitalGlobe-Vivid

bhousel:
Hello!  I'm happy to add an update to the Digital Globe imagery in the editor layer index project.  Digital Globe/Maxar has granted access to their Vivid product to all OSM partners for use in all editing applications.  The attached pdf contains links to detailed terms and product spec.   

We will also leave the existing DigitalGlobe Premium and DigitalGlobe Standard layers in place for a few months, though they will eventually be removed.

Thanks DigitalGlobe/Maxar!  👏


imagico:
If the terms of use are different from existing layers https://wiki.openstreetmap.org/wiki/DigitalGlobe should be updated.

> We will also leave the existing DigitalGlobe Premium and DigitalGlobe Standard layers in place for a few months, though they will eventually be removed.

Who is we here?

It would be regrettable if the DigitalGlobe Standard layer is not available any more in the future because a large number of different images is always beneficial and the lack of blending between individual images and less processing in general can often be beneficial for mapping.  And also because there are areas covered in Standard but not in Premium/Vivid.

Will there also be image date information like for the other layers?

bhousel:
Good questions @imagico 

> If the terms of use are different from existing layers https://wiki.openstreetmap.org/wiki/DigitalGlobe should be updated.

Links can go to the EarthWatch API and Terms of Use:
http://www.digitalglobe.com/legal/earthwatch-license
http://www.digitalglobe.com/legal/earthwatch-maps-api-end-user-terms-of-use
Also we should link to the attached PDF which grants use to the OSM partners for use in our editing applications.

> Who is we here?

I'm working with some new folks at Maxar who reached out to me because I was the original point of contact for #323.  I'm not sure whether they are on GitHub, but I will ask them to stop by this ticket and say hi.

> It would be regrettable if the DigitalGlobe Standard layer is not available any more in the future because a large number of different images is always beneficial and the lack of blending between individual images and less processing in general can often be beneficial for mapping. And also because there are areas covered in Standard but not in Premium/Vivid.

I agree - more imagery is better.. I'm just passing on the information that was given to me.

> Will there also be image date information like for the other layers?

I don't know but I'll ask.. The new layer is on a different service.. I'll check the EarthWatch API and see whether we can use to get metadata.

imagico:
Thanks.

If the Maxar people involved in this are reading here - maintaining the DigitalGlobe Standard layer for OSM mapping purposes would be desirable because it often offers images not available in any of the other DG data based image layers we can use.  This is helpful in particular

* in areas with high cloud incidence like [here](https://mc.bbbike.org/mc/?lon=7.662362&lat=46.183533&zoom=13&num=4&mt0=digitalglobe-standard&mt1=bing-satellite&mt2=digitalglobe-premium&mt3=esri-satellite)
* in areas with incomplete coverage like [here](https://mc.bbbike.org/mc/?lon=128.778561&lat=70.27847&zoom=13&num=4&mt0=digitalglobe-standard&mt1=bing-satellite&mt2=digitalglobe-premium&mt3=esri-satellite) or [here](https://mc.bbbike.org/mc/?lon=55.855435&lat=74.239243&zoom=13&num=4&mt0=digitalglobe-standard&mt1=bing-satellite&mt2=digitalglobe-premium&mt3=esri-satellite)

PR #664, opened 2019-05-18
add wms for ortho and dtm hillshade for kanton thurgau

rbuffat:
This PR adds two sources for the Canton Thurgau, Switzerland, Europe.

The first source is a WMS with a hillshade of a terrain model. The second source is a WMS with an ortho from 2017. This source is a duplicate to https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/KantonThurgau-2017-TMS.geojson (the same wms service, but through a tms proxy by SOSM). If you don't like duplicate I can remove the ortho WMS source from this PR.

While adding these WMS I noticed the following: 
- The WMS Server would support WMS 1.3.0 and supports only EPSG:4326 (and not also e.g. EPSG:3857).
- With WMS 1.3.0 and EPSG:4326 the format of the coordinates of the bbox are min_lat, min_lon, max_lat, max_lon [[WMS 1.3.0 standard, page 18]]( http://portal.opengeospatial.org/files/?artifact_id=14416):
```
EXAMPLE 1 A <BoundingBox> metadata element for a Layer representing the entire Earth in the CRS:84 Layer CRS
would be written as
<BoundingBox CRS="CRS:84" minx="-180" miny="-90" maxx="180" maxy="90">.
A BBOX parameter requesting a map of the entire Earth would be written in this CRS as
BBOX=-180,-90,180,90.
EXAMPLE 2 A <BoundingBox> representing the entire Earth in the EPSG:4326 Layer CRS would be written as
<BoundingBox CRS="EPSG:4326" minx="-90" miny="-180" maxx="90" maxy="180">.
A BBOX parameter requesting a map of the entire Earth would be written in this CRS as
BBOX=-90,-180,90,180. 
```
- At least in the id editor the bbox seems always to be in the format min_lon,min_lat,max_lon,max_lat. This should be fine for EPSG:4326 and WMS <= 1.1.1 [[WMS 1.1.1 standard]](http://portal.opengeospatial.org/files/?artifact_id=1081&version=1&format=pdf). 
```
The value of the
BBOX parameter in a GetMap request is a list of comma-separated numbers of the form
"minx,miny,maxx,maxy". 
```
- As the wms server also supports WMS 1.1.1 I could use it. 

Is this a known issue, or just a problem with the iD editor?


grischard:
It looks like iD won't be able to use either one, so I'm a bit meh about merging this:

https://github.com/openstreetmap/iD/issues/4843

Would it please be possible to add the Thurgau hillshade layer to the SOSM tms proxy, @simonpoole?

rbuffat:
@grischard The sources should work with iD. I specified in the url explicitly wms version 1.1.1 and the server accepts that.

I'm currently travelling but I will verify when I'm back.

rbuffat:
@grischard The sources work with my local installation of the latest id.[ ID supports EPSG:4326](https://github.com/openstreetmap/iD/blob/df699b2dc1a3f16efedb066516266ee58ee526d9/data/update_imagery.js#L59). 

For ID (and probably all other editors that don't check the wms version in the url) EPSG:4326 and WMS < 1.3.0 works. To get EPSG:4326 and WMS == 1.3.0 working, the position of the x and y coordinates in the bbox must be switched.

grischard:
Aha, good to know! I should really document this somewhere.

What do you think we should do about the duplicate? Won't it be confusing to users if the same source is shown twice?

rbuffat:
@grischard  I asked the canton of Thurgau if it would possible to add EPSG:3857 to their opendata wms layers and they very kindly added them within a few days!

I updated the sources accordingly. Furthermore, I added to following by the canton of Thurgau on opendata.swiss published wms layers:
* Cadastral survey map
* Hiking routes
* Cycling routes

Regarding the orthophoto layer. The existing sources uses the mapproxy.osm.ch to access the same data. It was added by @simonpoole. As far I understand it was necessary to use the mapproxy when wms was not supported by editors.  For me, it would make sense to directly access the wms layers and not use a proxy, but there might be reasons I'm not aware of. Ultimately, I think it is up to you (@grischard, @simonpoole ) to decide which source (or both) should be included in this index. As stated above, I'm happy to remove the orhto wms source from this PR.

grischard:
Hi @rbuffat, very cool! I think we can drop the proxy entry for the orthophoto layer then.

Klumbumbus:
Travis timed out and didn't generate the combined files.

simonpoole:
>
>     * Hiking routes
> 
>     * Cycling routes
> 
I had a quick look at the later, and it includes routes from SchweizMobil which is not open data and likely the canton doesn't have the rights to publish as is (except if they surveyed the routes naturally). As the usefulness of these two layers is roughly zero in any case I would suggest removing them till such a point in time that their provenance is clarified.

In general I would suggest not using ELI as the dumpster for any old junk found on the internet to start with as doing so just overwhelms our ability to vet sources properly, not even considering long term maintenance. 

Note:  had the PR title accurately reflected what was being added in the PR I would have complained far earlier.

rbuffat:
> I had a quick look at the later, and it includes routes from SchweizMobil which is not open data and likely the canton doesn't have the rights to publish as is 

@simonpoole basically you imply that the Canton of Thurgau is incompetent? This is quite a bold statement! In case you doubt that the Cantons do not have the right to publish the datasets, I suggest that you contact them to clarify.

Probably you are not aware, but at least two other Cantons offer hiking routes for download (but only as zipped shapefiles). The Canton of Valais also on[ opendata.swiss](https://opendata.swiss/de/dataset/freizeitverkehr-wandern), and the Canton of Bern on [their own website](https://www.geo.apps.be.ch/de/geodaten/suche-nach-geodaten/sheet/de-DE/8a58ea49-d2a7-9ce4-953a-18950217c3ae/geocatalog/complete/search_list.html) (Zugangsberechtigung: A: öffentlich zugängliche Geodaten, Datenabgabe: Freier Download, weitergehende Datenabgaben durch das AGI).  

Thus at least three Cantons would be incompetent. In my eyes, this is highly unlikely. 


As probably most visitors here are not aware of the situation in Switzerland (for hiking routes), here a brief summary. Switzerland is a bit special, as hiking routes are included in our constitution. The Swiss state delegates the responsibility for the planning and maintenance of hiking routes to the Cantons (The administration level below the state). Typically, hiking routes (and I assume also cycling routes) are included in the “Richtpläne” of the Cantons.

> except if they surveyed the routes naturally

I don’t think the Cantons survey the routes. I assume it is the other way around: the routes are built after the plans of the Cantons. 


> In general I would suggest not using ELI as the dumpster for any old junk found on the internet to start with as doing so just overwhelms our ability to vet sources properly, not even considering long term maintenance.

Is this “old junk found on the internet” just general, random grumble, or specific to these datasets? Because probably the Cantons have the most up to date datasets for hiking routes. I’m not sure if the two datasets published on openata.swiss correspond to the newest data the Cantons have because I don’t have the means to validate such a claim, but at least a yearly update interval is indicated in the metadata.

> As the usefulness of these two layers is roughly zero 

I understand that cycling and hiking routes are not of interest to you. Put please respect that others can have a different opinion. The coverage of hiking routes in OSM for the Canton of Thurgau is minimal. This dataset has the potential to change this. Furthermore, it can help to keep existing routes up to date.


Regarding SchweizMobil: I assume that SchweizMobil neither plans nor surveys routes. I think what they are using are data of the Cantons. On their website you can find the following statement: 
> Routendaten von SchweizMobil auf der Webkarte
Datenherr ist das Bundesamt für Strassen (ASTRA)
https://www.schweizmobil.ch/de/copyright-datenschutzerklaerung.html

I don't think it is valid to infer from the fact that SchweizMobil does not publish the routes as open data (which they probably can't anyway because they don't own the data)  that no one in Switzerland can publish hiking routes as open data.

rbuffat:
@Klumbumbus Thanks for fixing my mistake with the description!

PR #611, opened 2019-01-14
Ohio Statewide Imagery Program

1ec5:
Added the Ohio Statewide Imagery Program’s most recent available 1-foot and 6-inch orthoimagery layers. The geometry for the 1-foot layer is based on https://github.com/osmlab/editor-layer-index/issues/593#issuecomment-452133664, and I eyeballed the geometry for the 6-inch layer based on the slippy map linked from [the REST endpoint’s landing page](https://geo.oit.ohio.gov/arcgis/rest/services/OSIP/OSIP_6in_best_avail/ImageServer). The 6-inch layer is an `exportImage` endpoint, so it might not perform quite as well as a normal WMS endpoint in clients that request individual tiles.

Fixes #593.

PR #665, opened 2019-05-21
Migrate Ohio Statewide Imagery Program to new server

1ec5:
This PR makes a few important updates to the Ohio Statewide Imagery Program (OSIP) layers. OSIP has moved to a new server endpoint. The layers [will be removed](https://ogrip.oit.ohio.gov/ServicesData/GEOhioSpatialInformationPortal/RESTServiceEndpoints.aspx) from the old server endpoint at geo.oit.ohio.gov on May 27, so all references to that domain need to be removed by then.

The new server endpoint includes 6-inch OSIP III imagery from 2017 and 2018. This is the first time 6-inch imagery is available for Cincinnati and Dayton and the first time any imagery has been made available for those areas since 2012.

Expanded the 6-inch layer’s geometry to cover all of Ohio, just like the 1-foot layer, in anticipation of rolling updates. The 6-inch ImageServer is finally WMS-enabled, so the exportImage hack from #611 is no longer necessary.

before after
_1-foot from 2012 versus 6-inch from 2018_


1ec5:
Also made [corresponding changes to the JOSM wiki](https://josm.openstreetmap.de/wiki/Maps/USA?action=diff&version=157).

don-vip:
Why do we have a decrease of max-zoom?

1ec5:
Perhaps I misunderstood what the minimum and maximum zoom levels are for. The layers only cover Ohio, so there’s no practical difference between what the server would return at z8 versus z1. But the server does return something at lower zoom levels too; it just takes a long time. If clients are unlikely to downscale the imagery, then I can change it back.

don-vip:
I'm talking about the max-zoom. You decreased it from 20 to 19 in OSIP 1ft, and from 22 to 19 in OSIP 6ft. It means the best detailed imagery is no longer provided, with this change. If Ohio state didn't remove these high-res images (I don't see why they would), this should be reverted.

1ec5:
Ah, sorry for misreading. Previously, I had specified the maximum zoom level of the 6-inch layer by picking a number (22), since the exportImage method can return an arbitrarily large image for the same bounding box. It just keeps upscaling the same image without getting any higher in resolution.

This time, now that both layers are WMS-enabled, I entered the GetCapabilities URL for each one into JOSM and noticed that JOSM wouldn’t fetch any tiles beyond z20, even after zooming into z21. However, I accidentally entered the maximum zoom levels as 19 instead of 20. I changed them back to z20 in 927e7847d42e5a7c9e92ef5da94b06f97d5779cd and [on the wiki](https://josm.openstreetmap.de/wiki/Maps/USA?action=diff&version=158).

Thanks for catching my mistake!

PR #668, opened 2019-05-24
Remove DigitalGlobe Vivid, replace with Maxar Standard and Maxar Premium

bhousel:
Hi!  I've received a request to replace the DigitalGlobe-Vivid source added in #655 /  0b7cf41 with Maxar-branded Premium and Standard sources.

This will reflect the new branding (Maxar) and state clearly that it is Beta. The plan is to address any open issues on the Beta endpoints between now and end of June  –  preparing for the removal of the legacy endpoints no later than 6/30/2019.

Tagging @peterkern and @kevinbullock who can answer any questions.
Thanks!


grischard:
All cool, but please don't delete icons when doing changes like this until you release a new version of iD; users now have an ugly broken icon for Vivid.

bhousel:
Ok sorry!  new version of iD coming very soon...

pnorman:
Does the explicit permission for OSM need updating?

Zverik:
I've done a quick check in some Russian cities and found out that some have fresher imagery in DG Premium layer, which is likely to go away. Does it mean we need to quickly trace it in a month, or will the Maxar imagery be updated to be on par with DG Premium layer? @kevinbullock

PR #669, opened 2019-05-31
Update basemap.at layers

simon04:
- Add new surface layer
- Add new terrain layer
- Enable i18n for all basemap.at layers
- Concise filenames


ppete2:
Any idea why basemap.at Terrain and Surface are not available withing iD v2.19.6 yet? 
This layers would be a great help for mapping streams, tracks, buildings which are hidden in the Orthophoto within wood.

simon04:
Because they were removed a few months later via commit https://github.com/osmlab/editor-layer-index/commit/40ca57df30c8766a155e4b3bc3d14d98f83ef952. @grischard, why?

grischard:
Borken meant it probably wasn't answering. If they're working again, the commit can be reverted for these sources.

ppete2:
They are working at least in JOSM.

simon04:
@grischard, would you test those layers and re-add them to ELI again? Thank you in advance!

ppete2:
Do you know when these layers are available again within iD-editor? They still aren't.

rbuffat:
@ppete2 I'm not in the loop, but it looks as there are currently no developers actively developing iD: https://blog.openstreetmap.org/2021/04/26/joint-statement-from-quincy-morgan-and-the-osmf-board

Nobody seems to care to update the ELI version on the OSM homepage.  There are however some iD forks that are still maintained, e.g. https://mapwith.ai. 


CC @grischard

PR #674, opened 2019-06-13
Update Mainz-Gint-latest.geojson

ghost:
Change WMS Source to MapProxy WMS Server


grischard:
Danke!

Klumbumbus:
min_zoom is 12 now and it supports EPSG:3857 too

grischard:
Thanks @Klumbumbus, added in f774fa1

PR #677, opened 2019-06-14
created MN Composite service geojson source file

DcJoeS:
created MN Composite service geojson source file


grischard:
I doubt that anything that uses ELI will natively support EPSG:26915, unfortunately. The best solution would be to run a MapProxy in front of it - maybe OSM US can help with the hosting?

PR #680, opened 2019-06-29
Deprecate digitalglobe

bhousel:
Just a followup from #668, as requested we are removing the old DigitalGlobe imagery endpoints "no later than 6/30/2019."

See https://www.openstreetmap.org/user/@kevin_bullock/diary/390153 for more info!

cc @peterkern and @kevinbullock.
Thanks!


bhousel:
BTW this pull request leaves the `DigitalGlobePremiumImagery.png` logo only, so that the logo will continue to work for the next few days.  I also took this opportunity to shrink a few other logos that were really much larger in size than they needed to be.

PR #683, opened 2019-07-02
Update aerial imagery in Capetown, South Africa

tsmock:
Add 2017 and 2018 imagery from Capetown, South Africa.

Largely copied from the 2015 aerial geojson file, with modifications for current years. I did leave a "best" tag in the 2018 imagery properties.


grischard:
Thank you!

tsmock:
It looks like something happened on the Travis CI build ( https://travis-ci.org/osmlab/editor-layer-index/builds/555263087 ).

Do you have any idea why it would have errored out?
`    lons = [p[0] for ring in geometry['coordinates'] for p in ring]`

I'm not seeing any floats that are not inside another array. It should be something like
```
    "geometry" : {
      "type" : "Polygon",
      "coordinates" : [
         [
            18.35571984388,
            -34.06347619207
         ]
      ]
    }
```
right? Or did I mess up when I converted it from a LineString?

PR #693, opened 2019-08-16
IndianaMap

### LINKED ISSUE #686: IndianaMap
## URL of the imagery
* https://maps.indiana.edu/arcgis/rest/services/Imagery/Orthos_2016/MapServer
* https://maps.indiana.edu/arcgis/rest/services/Imagery/Orthos_2017/MapServer
* https://maps.indiana.edu/arcgis/rest/services/Imagery/Orthos_2018/MapServer

## Two letter country code
US

## Desired layer name
* IndianaMap Orthoimagery 2016 (Central)
* IndianaMap Orthoimagery 2017 (Eastern)
* IndianaMap Orthoimagery 2018 (Western)

## License
[Public domain](http://gis.iu.edu/datasetInfo/statewide/in_2016.php):

> This is a public domain dataset with no use restrictions.

Yes

## Attribution (text, URL, required?)
Indiana Office of Information Technology, Indiana University Spatial Data Portal, UITS, Woolpert Inc.
(not required)

## Privacy policy (URL)
http://gis.iu.edu/privacy.php

## Icon
[](http://gis.iu.edu/files/css/resources/blockiu_white.gif)

## Bounding polygon
* [Overpass query for 2016](http://overpass-turbo.eu/s/KBb)
* [Overpass query for 2017](http://overpass-turbo.eu/s/KBc)
* [Overpass query for 2018](http://overpass-turbo.eu/s/KBd)
* [GeoJSON for each year](https://gist.github.com/1ec5/d720d47f69a46b8bed07f3ca65dd3f86) (may need to be simplified)

1ec5:
Added the three WMSes for IndianaMap orthoimagery. Note that the 2018 geometry covers more counties than the WMS currently provides; these additional counties will be added to the 2018 MapServer sometime this year.

Fixes #686.


1ec5:
[Added to JOSM](https://josm.openstreetmap.de/wiki/Maps/USA?action=diff&version=166&old_version=165).

grischard:
Hi Minh,

On one hand, it looks like I'm the one to blame for not having CI tools that caught that you were trying to add a Multipolygon, which isn't supported. I thought `check.py` was supposed to catch this.

On the other hand, you didn't wait for feedback, you merged quickly and on a weekend, and used a Multipolygon while the documentation says to use a Polygon.

Can you please fix your PR to use a Polygon geometry instead of Multipolygon? And I'll try to see why check.py didn't catch this.

1ec5:
Sorry for jumping the gun. Would it be possible to use a feature collection instead? Unfortunately, all three layers have discontiguous coverage areas. If not, I can try to draw a very thin line connecting the polygons.

1ec5:
Looks like the validation script only checks whether the `type` can be compared to the string `Polygon`, not whether it’s equal to that string:

https://github.com/osmlab/editor-layer-index/blob/e3b91eadb69cc9cc9cec8e4fb98b1eed1a85bd12/scripts/check.py#L127-L130

1ec5:
#697 fixes the bug in the validation script.

grischard:
The usual, admittedly unfamiliar, for-historical-reasons way to do it is to have a simple polygon with more than one ring. Look at the France geojsons with Corsica for example.

grischard:
Did you see #694 too?

You should play the lottery 😄

Issue #690, opened 2019-08-10
Make privacy_policy_url mandatory


simonpoole:
Currently only a very small number of sources have the privacy_policy_url set, even very new entries like the Maxar layer don't. 

This is an issue, not only, but particularly for the OSMF as using an editor on osm.org will not only transfer the IP address (which in itself is already personal data) and associated browser information, but also the viewed area to the imagery providers. Now we do disclose this in the OSMF privacy policy and point to here for more information, but that doesn't really help if nearly no layers have the information.

Best would be if the information would easily be accessible in the editor itself (something which Vespucci has had for a while), but I'm not holding my breath for that.

Naturally it will take a longish time for this to be available for all configurations, but we could start by requiring the information for any new or updated layers.

Thoughts?

grischard:
So it's a bit like #439 ? I think most sources don't even *have* a privacy policy.

A nice travis lint script would be one that checks PRs for such attributes, yes.

simonpoole:
> 
> 
> So it's a bit like #439 ? I think most sources don't even _have_ a privacy policy.
> 
> A nice travis lint script would be one that checks PRs for such attributes, yes.

Well, I was thinking more (the back log in #439 proving my point) that the build should fail and the corresponding PR not be mergable, if either of those URLs are missing. Naturally we can only change that going forward for sources that are modified or are new, the question is if we can determine the last changed data in a reasonable form. I suppose git could potentially be used for that.

1ec5:
> I think most sources don't even *have* a privacy policy.

This is often the case with state and local government agencies in the U.S. For example, the WMS added in #593 is provided by the state’s Department of Administrative Services, which has [a privacy policy](https://das.ohio.gov/Privacy) covering only the department’s homepage but [no privacy policy](https://ogrip.oit.ohio.gov/privacy.aspx) explicitly covering the WMS. Maybe ArcGIS Server doesn’t have a built-in setting for one.

When these agencies do have privacy policies, they [read very differently](https://ohio.gov/wps/portal/gov/site/home/privacy-notice-and-policies) than typical commercial privacy policies, because public records / open government laws sometimes work at cross purposes to personal data protection.

simonpoole:
> 
> 
> > I think most sources don't even _have_ a privacy policy.
> 
[...]

PR #695, opened 2019-08-17
add privacy policy url for Bing

matkoniecz:
fixes tiny part of #690

this url is linked from bottom left ('Privacy and Cookies') text
on Bing Maps, for example from https://www.bing.com/maps?FORM=Z9LH3


grischard:
One step at a time. Thank you :)

PR #697, opened 2019-08-18
Verify that layer geometry is a polygon, not that it exists

1ec5:
Fixed the part of the validation script that ensures each layer’s geometry is a polygon.

/ref https://github.com/osmlab/editor-layer-index/pull/693#issuecomment-522301116

### DIFF
--- a/scripts/check.py
+++ b/scripts/check.py
@@ -124,9 +124,7 @@ for filename in arguments.path:
         # If we're not global we must have a geometry.
         # The geometry itself is validated by jsonschema
         if 'world' not in filename:
-            try:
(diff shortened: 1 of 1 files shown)


1ec5:
At first I accidentally based this PR on #693, so you can see [the failure](https://travis-ci.org/osmlab/editor-layer-index/builds/573353299?utm_source=github_status&utm_medium=notification) that it would cause.

grischard:
Thank you! I think this would cause misery and exceptions if your geometry has no type at all. I have a fix ready to push :)

grischard:
Done in 6ac50ec0e9018d1c1309d69899fcc04ea0b48f6a

PR #701, opened 2019-08-27
Create BE_OrthoPhoto2019_Brussels_WMS.geojson

jbelien:
Add UrbIS-Ortho 2019 (Brussels, Belgium)


simonpoole:
Privacy policy url?

jbelien:
@simonpoole Is this what you mean by privacy policy : https://bric.brussels/en/our-solutions/security-solutions/information-security-consultancy-datapprivacy

If so, I'll add it to every layer from BRIC :)

simonpoole:
@jbelien not really. 

The operators of any imagery server get data that is clearly covered b the GDPR, that is IP address and related information from the client and information which geographic area they are interested in. By law they are required to provide a document indicating what they do with the data (including if they just throw it away), that is the document that should be linked to in the entry.

jbelien:
@simonpoole 

Would Chapter 5 in https://bric.brussels/en/disclaimer-conditions-of-use-of-the-bric-website fit ?
It's focus on the use of the BRIC website and not really the BRIC web services but that's the only thing I could find ...

simonpoole:
@jbelien as a stop gap measure I suppose yes, but as they need to either include their other services in the text or create a separate one for the web services, I would suggest contacting them and pointing the issue out.

grischard:
Ok, let's add this as a privacy policy url and merge this?

jbelien:
I don't have anything better...
Should I add https://bric.brussels/en/disclaimer-conditions-of-use-of-the-bric-website as privacy url ?

I'll admit I'm a bit surprised, the aerial imagery from previous years was merged without the privacy url.

simonpoole:
@jbelien while the changed requirements are motivated by the introduction of the GDPR, we probably should have always pointed out the privacy impact of using 3rd party imagery sources.

grischard:
Adding it to the previous years would be really cool.

Seeing the rate at which #439 is progressing, I doubt we'll ever have a privacy policy for every source, but we can at least move in that direction.

jbelien:
Okay, I'll add https://bric.brussels/en/disclaimer-conditions-of-use-of-the-bric-website as privacy URL for all the services coming from CIRB/BRIC! 👌

grischard:
Merci!

PR #702, opened 2019-08-28
add au/act 2019 June imagery

andrewharvey:
PS. any chance I could also be granted permissions to merge? I'd still leave it open for a few days for others to review...


simonpoole:
Any chance of a privacy policy url for these?

Klumbumbus:
Please use "Feb 2019" and "Jun 2019" instead of  "2019 Feb" and "2019 Jun" in the name to stay in sync with JOSM (https://josm.openstreetmap.de/wiki/Maps/Australia?action=diff&version=71)

andrewharvey:
> Any chance of a privacy policy url for these?

Added.

> Please use "Feb 2019" and "Jun 2019" instead of "2019 Feb" and "2019 Jun" in the name to stay in sync with JOSM (https://josm.openstreetmap.de/wiki/Maps/Australia?action=diff&version=71)

Okay. I've kept the `id` for the Feb2019 imagery unchanged to avoid a breaking change.

FrakGart:
I enquired about the Spookfish attribution because the company was bought out last year and no longer exists. I didn't get a reply. So, I've left the JOSM entry with the same attribution as the Feb 2019 version.

andrewharvey:
> I enquired about the Spookfish attribution because the company was bought out last year and no longer exists. I didn't get a reply. So, I've left the JOSM entry with the same attribution as the Feb 2019 version.

I've just copied the upstream metadata, which I think is all we should do. If that metadata gets updated we can change it, but since it matches now, I think it's fine.

FrakGart:
> I've just copied the upstream metadata, 

No you haven't. You've corrected the spelling of Government ;-)

Point was just explaining why the JOSM entry is different.

PR #707, opened 2019-09-10
Add TIGER 2019 Roads Overlay

iandees:


iandees:
This covers more than CONUS – do we have a polygon somewhere for all of USA (including Alaska and territories)?

bhousel:
> This covers more than CONUS – do we have a polygon somewhere for all of USA (including Alaska and territories)?

The previous TIGER overlays already use a polygon that includes Alaska and territories.

iandees:
Ah indeed it does. The GitHub rich diff viewer only showed CONUS, that's why I was confused :).

bhousel:
> Ah indeed it does. The GitHub rich diff viewer only showed CONUS, that's why I was confused :).

Yeah, unfortunately this project treats the polygon geometry as an array of outers, instead of arrays of proper GeoJSON geometries with outers and holes.  That makes it unnecessarily difficult to operate with other tools (like the rich diff viewer for example).

simonpoole:
Given that the tiles are hosted by mapbox, it should be possible to provide a privacy_policy_url for the layer.

1ec5:
In case it helps, Mapbox has [this company-wide privacy policy](https://www.mapbox.com/legal/privacy/). (In addition to the raw Styles API used here, it also covers services like customer accounts, Mapbox Studio, and the mobile SDKs that aren’t relevant here, but I don’t know of a privacy statement specific to the tile server.)

grischard:
Cool, let's add that as a privacy policy url and merge this.

PR #708, opened 2019-09-11
Add Luftbild 2016 Frankfurt am Main

Nakaner:


simonpoole:
@Nakaner it's privacy_policy_url :-)

Nakaner:
@simonpoole I fixed it.

grischard:
It's still failing, because you probably want 'explicit' where you have the permission url, which itself goes elsewhere.

xuiqzy:
Thanks for bringing in this source, I just wanted to add it and saw your PR! :)

Because of the age, I am not so sure that the "best" attribute should be set in comparison to other available imagery for this region.
This opens up the possibility that many enter details that are 3+ years out of date without looking at the date of the imagery used. The ofther way around, there are less wrong edits and users who want to enter more details can select the 2016 imagery knowing that it is older!

The permisson url itself would probably just go in the `license_url` field, it says license or **permission** [here](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L72).
The original license info can be maintained by putting it into `license`, but since it has no SPDX identifier, it has to be classified as `COMMERCIAL`. If you want, you can include the license in the description shortly in braces afterwards or something like that.

Since there are translation files for every language, I would just keep the name and description in English for easier translation into other languages by translators who speak only their language and English.

xuiqzy:
@Nakaner Are you still working on this? :) There is really not much left here, but if you don't want to, I can do it, too.

grischard:
@xuiqzy go ahead and do it :)

grischard:
I went ahead and did it. Thank you @Nakaner and @xuiqzy

xuiqzy:
Thanks a lot for finishing this @grischard ! :)

Considering the Maxar imagery that is now available in the region, that is a lot newer, I wouldn't put the `best` attribute in it for it to become the default for this region. This could lead to users wrongly mapping old buildings etc and remove newer ones.
Shouldn't we let the users familiar with different imagery explicitly choose this layer, as then they will clearly see the older date?

grischard:
Ah, yes, I'll remove 'best' then.

grischard:
Done in abccecaf49b302c2ac5669b4d3bc46471988e1df

Nakaner:
The German community prefers mappers to use official imagery as first choice. Commercial imagery is poorer in terms of georeferencation and impairs the existing data quality.

andrewharvey:
> The German community prefers mappers to use official imagery as first choice. Commercial imagery is poorer in terms of georeferencation and impairs the existing data quality.

Agreed, `best` doesn't necessarily mean latest, it's a subjective call, but no matter what the default is it can lead to issues (set it as the newest and then some people may re-align roads to less accurate imagery, set it as the most accurate and then they start changing things that you can see are different in the newer imagery).

So either way mappers will need to be aware of the different imagery layers and swap between them.

Klumbumbus:
The URL doesn't work as the version parameter is missing. Also to make the source work on all zooms it seems you need to add all layers of that wms together. Here is the working URL: https://josm.openstreetmap.de/wiki/Maps/Germany?action=diff&version=218

grischard:
Thank you @Klumbumbus, fixed in 4ae2a5c. I've kept the https though.

grischard:
There's also https://offenedaten.frankfurt.de/dataset/wms-luftbild-2017 if anyone feels like making another PR.

Klumbumbus:
Thanks, I missed that. The attribution URL works with https too.

PR #709, opened 2019-09-12
Add AGIS 2019 imagery, remove best flag from 2017 version

simonpoole:
The title says it all.


rbuffat:
@simonpoole  As you are in contact with AGIS, would it be possible to ask them if it would be possible to also use their hillshade digital terrain model imagery?  https://www.ag.ch/geoportal/geodatenshop/Datendokumentation.aspx?Datensatzelement=4189

The data source is available in the agisviewer (Karten -> Geländemodelle -> Hillshade - > DTM 0.5 Meter Hillshade), but I did not find an official wms/wmts layer.

simonpoole:
@rbuffat I'm on vacation right now, but will ask AGIS when I get back, as they do not provide a WMS service for the general public currently, it would boil down to either getting the data and serving it ourselves, or proxying a semi-private WMS server (as we do for this layer).

simonpoole:
@rbuffat I've got permission to proxy the layer, will set it up asap.

rbuffat:
@simonpoole That is great news! Thank you!

PR #712, opened 2019-09-29
Slovenia/RABA-KGZ: Added permission_osm=explicit

stefanb:
https://wiki.openstreetmap.org/wiki/Slovenia_Landcover_Import_-_RABA-KGZ#License


grischard:
Fantastic, thank you!

PR #716, opened 2019-10-04
Add Croatian State Geodetic Administration tophographic layers

hbogner:
The Croatian State Geodetic Administration(Državna Geodetska Uprava) changed  their Geoportal terms of use to allow derivative works, so the aerial imagery and topographic maps can now be used for OSM.
Adding topographic layers:
topo 1:25 000
topo 1:25 000-new
topo 1:100 000
topo 1:200 000

Additional
Fix to Zagreb 2018 area coverage


grischard:
Čestitamo!

hbogner:
Hvala :)

PR #720, opened 2019-10-09
Add all missing OpenRailwayMap overlays

hikemaniac:
This pull request adds the OpenRailwayMap Maxspeed overlay and the OpenRailwayMap Signalling overlay.


grischard:
Good idea, thanks!

Are we even allowed to use these actually? I couldn't find anything on the website, and there are no license attributes.

Can you please add the privacy policy at https://www.openrailwaymap.org/en/imprint ?

Would it also be possible to have better descriptions for the layers?

hikemaniac:
@grischard I just added the privacy policy URL and a better description. Thank you for pointing that out. 👍 

Since the data on the overlay comes directly from OpenStreetMap, I presume that we are allowed to use the layers. The infrastructure OpenRailwayMap layer has been in the index since 2018 and it came from the JOSM layer index, therefore I am positive that the license is OSM-compatible for all the other layers of OpenRailwayMap as well. On the website, the attribution text always remains the same regardless of the layer, so the terms should also be exactly the same.

grischard:
I wouldn't want to add it without their permission, even if just to make sure we won't hammer the servers. See https://github.com/OpenRailwayMap/OpenRailwayMap/issues/88 - it's been a know issue for a while.

grischard:
@Nakaner did you mention that you'd updated a wiki page somewhere that would fit?

grischard:
https://wiki.openstreetmap.org/wiki/OpenRailwayMap/API#Terms_of_use is the wiki page that allows usage.

grischard:
Thank you @hikemaniac !

maro-21:
Could you please correct it to make it translatable?

hikemaniac:
@maro-21 Thanks for the suggestion, I created a PR for this in #832. :+1:

maro-21:
Thank you!

PR #721, opened 2019-10-18
Add additional Croatia layers

hbogner:
Add additional Croatia layers:
osm-hr:
US Topo
Orbview3
dgu.hr:
DGU HOK
DGU Aerial 2011
DGU Aerial 2014-2016
DGU Aerial 2017
DGU Aerial 2018

PR #723, opened 2019-10-20
Croatia wms&tms bugfix

hbogner:
Fix bugs reported in https://github.com/osmlab/editor-layer-index/pull/721#issuecomment-544193232


Klumbumbus:
Thanks.

Issue #700, opened 2019-08-23
Add additional NLS maps for GB


boothym:
I got in touch with the NLS and they have updated the wiki page with some additional out of copyright imagery layers, quite detailed and mostly covering Scotland. https://wiki.openstreetmap.org/w/index.php?title=National_Library_of_Scotland&type=revision&diff=1892687&oldid=1235691

I've detailed them below:

**NLS - OS/War Office - GSGS 3906 1:25,000, 1940-43**

tms:https://geo.nls.uk/mapdata3/gsgs3906/{z}/{x}/{y}.png (Max zoom 16)

**NLS - OS 1:10,560 National Grid Maps, 1940s-1960s (This layer covers most of Great Britain).**

tms:https://geo.nls.uk/mapdata3/os/britain10knatgrid/{z}/{x}/{y}.png (Max zoom 16)

**NLS - OS 1:1,250 National Grid Maps, 1940s-1960s**

tms:https://geo.nls.uk/mapdata3/os/scotland_1250_country/{z}/{x}/{y}.png (Max zoom 20)

**NLS - OS 1:2,500 National Grid Maps, 1940s-1960s - single sheets**

tms:https://geo.nls.uk/maps/os/scotland_2500_singles/{z}/{x}/{y}.png (Max zoom 18)

**NLS - OS 1:2,500 National Grid Maps, 1940s-1960s - double sheets**

tms:https://geo.nls.uk/maps/os/scotland_2500_doubles/{z}/{x}/{y}.png (Max zoom 18)

grischard:
Thank you again!

Issue #699, opened 2019-08-23
Add end_date for some GB sources


boothym:
Would also basically fix #607 

https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/7thSeries-OS7.geojson - 1960

https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/OS125khistoric-OSM.geojson - 1961

https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/OSNewPopularEditionhistoric.geojson - 1955

https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/OSScottishPopularhistoric.geojson - unsure, but going by housing developments in my town 1955 sounds about right

Also, the url of https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/NLS-OS6InchScotland1842-82.geojson no longer works and needs to be changed to https://geo.nls.uk/mapdata3/os/6inchfirst/{z}/{x}/{y}.png (taken from the wiki)

PR #726, opened 2019-10-21
Add end_dates and new GB sources from NLS

### LINKED ISSUE #607: Remove OS125khistoric-OSM as it is a duplicate
Or at least add an end_date of 1962 so it doesn't show up in iD.

This layer: https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/OS125khistoric-OSM.geojson

is the same as: https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/NLS-OS125k1stSeries1937-61.geojson

but the first one has a fraction of the coverage (see below). Doesn't make sense to keep them both. 
id

boothym:
Decided to have a go at adding some new sources myself instead of waiting for someone else to do it 😄 

Hopefully the polygon co-ordinates are ok, some have been copied from other gb sources and some are newly created.

Resolves #607, #699, and #700


grischard:
Thank you!

PR #728, opened 2019-10-30
ASIT VD stored imagery update

yblatti:
- switch to HTTPS
- bump Lausanne imagery to 2016 flight


grischard:
Looks good!

Do you want to rename Lausanne-Orthopho-2012.geojson to 2016?

Is there any value in keeping the 2012 as a historic imagery layer?

yblatti:
> Do you want to rename Lausanne-Orthopho-2012.geojson to 2016?

Right !

> Is there any value in keeping the 2012 as a historic imagery layer?

Not sure of the value, plus we're cannot guarantee the storage of historic versions on our small server.

grischard:
Ah, and we should have https://www.asitvd.ch/tarifs-conditions.html as a `privacy_policy_url`.

grischard:
Thank you!

PR #730, opened 2019-11-08
Repairing access to USGS imagery

LucasLarson:
Access was broken because we were requesting the data through an overescaped URI (it should have been `https://ags-proxy.openstreetmap.us/tiles/{zoom}/{x}/{y}?url=https://basemap.nationalmap.gov/arcgis/rest/services/USGSImageryOnly/MapServer` instead of `https://ags-proxy.openstreetmap.us/tiles/{zoom}/{x}/{y}?url=https%3A%2F%2Fbasemap.nationalmap.gov%2Farcgis%2Frest%2Fservices%2FUSGSImageryOnly%2FMapServer`), but the proxy isn’t even necessary, so I’m removing that as well.

### DIFF
--- a/sources/north-america/us/USGSImagery.geojson
+++ b/sources/north-america/us/USGSImagery.geojson
@@ -4186,7 +4186,7 @@
         "min_zoom": 12,
         "name": "USGS Imagery",
         "type": "tms",
-        "url": "https://ags-proxy.openstreetmap.us/tiles/{zoom}/{x}/{y}?url=https%3A%2F%2Fbasemap.nationalmap.gov%2Farcgis%2Frest%2Fservices%2FUSGSImageryOnly%2FMapServer",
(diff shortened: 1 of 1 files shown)


iandees:
Last I checked the proxy was necessary because the tile root that they were using on their server was offset from zero, meaning everything would appear in the editors in the wrong spot. Have you tested the non-proxied URL in an editor?

LucasLarson:
I tested without the proxy in places I know well in a Western US state and in an Eastern US state using JOSM. In both places, the imagery appeared where I expected it.

On the other hand, I’m actually not able to get the proxied version to load at all, even when the URI is unescaped, and even accounting for `{zoom}` versus `{z}` and `{x}/{y}` versus `{y}/{x}`.

iandees:
Yea, looks like the direct URL should work just fine.

grischard:
Thank you @LucasLarson !

Issue #731, opened 2019-11-09
GRAFCAN is better in the canary islands than PNOA - add a "best" flag for it?


JonasCz:
I've noticed that "GRAFCAN - Canary Islands" is better than "PNOA Spain" in the Canary islands, since it's much sharper (higher resolution). Maybe it'd be possible to add the "best" flag to it, so that editors will use it by default? (Not entirely sure how that works, so no pull request...)

For reference, here's a comparison screenshot (click for full-res): 

GRAFCAN:

![GRAFCAB](https://github.com/JonasCz/stuff/raw/master/Grafcan.jpg)

PNOA:

![PNOA](https://github.com/JonasCz/stuff/raw/master/Pnoa.jpg)

Thanks;

grischard:
Hello! So GRAFCAN is a local layer, and  PNOA covers all of Spain? PNOA might already be tagged as best because it's better than the other choices for the rest of Spain. We could tag the two as 'best', but the behaviour then depends on the editors.

See also #691, #130, #390 and for a potential way forward #586

JonasCz:
Yep, exactly - GRAFCAN is local for the canary islands only, but it's better there.

With both tagged as "best" iD chooses the smaller area one first (i think?), so that seems like it would yield good results for at least iD users.

grischard:
Ok, let's try and see :). Do you want to PR this?

JonasCz:
PR is here: https://github.com/osmlab/editor-layer-index/pull/732

I think I got the JSON right, but please check.

grischard:
Merged in 9adac880a898ff2f6dd9acbf5893cf12e8cb5b70, thank you!

PR #732, opened 2019-11-14
Add "best" flag for GRAFCAN, since the resolution is better than PNAO spain in the Canary Islands.

JonasCz:
"GRAFCAN - Canary Islands" is better than "PNOA Spain" in the Canary islands, since it's much sharper (higher resolution). Issue here with screenshots: https://github.com/osmlab/editor-layer-index/issues/731

I think I got the JSON right, but check to make sure.


grischard:
Thank you!

Issue #136, opened 2016-01-29
Better source categorization


bhousel:
per @nicolas17 #134 

> I think imagery layers (like satellite) should be here, external raster layers (like some government's open data) should be here, and OSM-based layers (like QA) should be here. But there should be a field in the JSON saying in what "category" the layer is, to allow filtering if wanted.

grischard:
Why and when would this kind of filtering be wanted?

simonpoole:
For example a website wanting to show aerial imagery but not maps ....

Klumbumbus:
see also https://josm.openstreetmap.de/ticket/16103

Marc-marc-marc:
for the key name:
@grischard proposes `category`in #266 that look like fine.
for the value : 
@grischard proposes `osm`in #266 for render made from osm data. it would have the advantage of involving an (c) openstreetmap contributors.
@stoecker [proposes](https://josm.openstreetmap.de/ticket/16103) `map` and `photo`.
it look like a good idea to also add those 2 values (map for not-osm-based map like official one, and photo for sat and drone imagery).

did we need a value for derivative works like QA ?

Klumbumbus:
>@stoecker proposes map and photo.
The disuccion in this ticket went a bit further

Marc-marc-marc:
@Klumbumbus you're right. I forget to check it again before posting.
from josm ticket, i like map, osmbasedmap, photo, other
I dislike historicmap. should it be enough to use "date" for historic category ? I think so.
question remain for QA : is it a good idea to put them into osmbasedmap ? a layer which displays for example the error points found by Osmose, it is not really a map, but still osm-based.
I hope that the discussions on the two projects will continue to be mutually enriching in order to have a common vision that will facilitate everyone's life.

Marc-marc-marc:
so what ? what did you ting if : 
- we add a key category with value map, osmbasedmap (or osmbased ?), photo, other ? qa ?
- we add not-the-lasted=no to allow app to hide or group previous versions of a layer ?

Klumbumbus:
**`category`** was added just today to josm https://josm.openstreetmap.de/ticket/16103#comment:28 with the following supported values:
**`photo`** aerial or satellite photo, **`map`** a map, **`historicmap`** historic or otherwise outdated map, **`osmbasedmap`** map based on OSM data, **`historicphoto`** historic or otherwise outdated aerial or satellite photo, **`other`** any other type

[...]

PR #733, opened 2019-11-15
Add a category attribute like JOSM, thanks @simonpoole

simonpoole:
As discussed. 

I'll start adding the field to some of the sources in a 2nd commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ about which licenses are compatible with this index.
 The 'source' documents for this project are the .geojson files in `sources`. To add
 a new imagery source, add a new file to this directory.
 
-Each source must be a GeoJSON `Feature` and must minimally have `name`, `type`,`url` and `category`properties.
+Each source must be a GeoJSON `Feature` and must minimally have `name`, `type`,`url` and `category` properties.
 To improve readability, the keys of the GeoJSON document should be ordered consistently: `type`, `properties`, then `geometry`.
 
 We further recommend to add the licence related and `privacy_policy_url` properties.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,10 +8,10 @@ about which licenses are compatible with this index.
 The 'source' documents for this project are the .geojson files in `sources`. To add
 a new imagery source, add a new file to this directory.
 
-Each source must be a GeoJSON `Feature` and must minimally have `name`, `type`, and `url` properties.
+Each source must be a GeoJSON `Feature` and must minimally have `name`, `type`,`url` and `category`properties.
 To improve readability, the keys of the GeoJSON document should be ordered consistently: `type`, `properties`, then `geometry`.
 
-We further recommend to add `category`, the licence related and `privacy_policy_url` properties.
+We further recommend to add the licence related and `privacy_policy_url` properties.
 
 See [schema.json](schema.json) for the full list of available properties.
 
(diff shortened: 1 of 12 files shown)


andrewharvey:
:+1: Is there any guidance on when a imagery becomes historical? Is historical imagery intended for the ELI?

It would be nice to have a description of each somewhere in the documentation, if possible.

simonpoole:
https://josm.openstreetmap.de/wiki/Maps so it isn't actually "historic" the way JOSM uses it, just outdated.

simonpoole:
> 
> 
> https://josm.openstreetmap.de/wiki/Maps so it isn't actually "historic" the way JOSM uses it, just outdated.

Any opinions on if we should follow JOSM on this specific point or not @grischard & @Marc-marc-marc  ?

grischard:
We should follow JOSM on this point. Editors might of course want to show a better explanation - "newer layer available" for example.

grischard:
I'm all in favour of making the attribute mandatory, and doing a great big retro-tagging of all existing sources.

We should also adapt the conversion scripts.

Marc-marc-marc:
> > https://josm.openstreetmap.de/wiki/Maps so it isn't actually "historic" the way JOSM uses it, just outdated.
> 
> Any opinions on if we should follow JOSM on this specific point or not @grischard & @Marc-marc-marc ?

as said in the issue,
I find a very bad idea to have 2 meaning for historic (osm <> josm).
I find also a bad idea to have 2 characteristics in one (type+latest)
but if it's too complicated to use a more meaningful tag (latest, outdated,...) and split those in 2 and some people think that a common tag to both projects is better than a quality tag, then I prefer not to oppose.

simonpoole:
I went through the Swiss sources and tried to categorize those with category photo that made sense, and it clearly showed the problem with historicphoto: is a 10 year old layer historic if there is no successor or is it considered current?

The remaining question is if we should copy the attribute values from JOSM mechanically, or simply wait till people to it of their own accord?

Marc-marc-marc:
> I went through the Swiss sources and tried to categorize those with category photo that made sense, and it clearly showed the problem with historicphoto: is a 10 year old layer historic if there is no successor or is it considered current?

if it's the lasted, it'sn't historic for josm.

> The remaining question is if we should copy the attribute values from JOSM mechanically, or simply wait till people to it of their own accord?

an automatic bidirectional synchronization would obviously be the best, so that the contribution of the 2 projects would benefit everyone.
But if not, someone would have to do it on their own account.

don-vip (approved):

grischard:
See #136 - JOSM recently added 'elevation' as a type.

Klumbumbus:
I updated the JOSM database with the new category types: 19 qa and 30 elevation

don-vip:
20 qa now. I added French BANO as qa too.

simonpoole:
> 
> 
> See #136 - JOSM recently added 'elevation' as a type.

I've added it too now, even though it seems to be a slightly weird choice.

andrewharvey:
>> See #136 - JOSM recently added 'elevation' as a type.

> I've added it too now, even though it seems to be a slightly weird choice.

Seems appropriate to me. Though I noticed JOSM actually describes what these values are to be used for and I think we need to do that here too eg.

`elevation` - A map of digital terrain model or digital surface model.

Klumbumbus:
> `elevation` - A map of digital terrain model or digital surface model.

That was updated: https://josm.openstreetmap.de/changeset/15685/josm

grischard:
Looks ready to merge for me; we can update the various scripts in another step.

simonpoole:
@grischard https://github.com/osmlab/editor-layer-index/pull/760

PS: @andrewharvey I would have used "terrain"

andrewharvey:
> PS: @andrewharvey I would have used "terrain"

I think either is fine, so long as we include the description within the schema documentation.

PR #734, opened 2019-11-17
Add Mesa County 2019 Imagery (WMTS)

ghost:
This is adds the latest imagery from the County.
Permissions haven't changed.


grischard:
Looks good!

Do they have a privacy policy of some kind?

ghost:
The closest thing I could find is this: https://www.mesacounty.us/contact-us/copyrightdisclaimerprivacy/

I'll see if I can find something more specific tonight.

On November 18, 2019 6:03:02 AM MST, Guillaume Rischard <notifications@github.com> wrote:
>Looks good!
>
>Do they have a privacy policy of some kind?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/osmlab/editor-layer-index/pull/734#issuecomment-555005770

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ghost:
I didn't find anything more specific.

I've added it to the other Mesa County Imagery files.

grischard:
Thank you!

You've added a 'wmts' type endpoint, but as far as I'm aware, no editors that use ELI support it. See https://github.com/openstreetmap/iD/issues/4828 for iD for example.

The WMTS can't provide EPSG:3857? Would there be a WMS that does instead? Let's merge this as it is otherwise.

ghost:
Only WMTS and only EPSG:3857 (see https://mcgis.mesacounty.us/image/rest/services/Mosaic_Datasets/MesaCounty_2019/ImageServer -- this has the endpoints, and the WMTS capabilities document only references EPSG:3857).

I did try to take an image url (for example https://mcgis.mesacounty.us/image/rest/services/Mosaic_Datasets/MesaCounty_2019/ImageServer/WMTS/tile/1.0.0/Mosaic_Datasets_MesaCounty_2019/default/default028mm/10/83680/86095 ) and convert it to a TMS/WMS template, but that didn't end up working in JOSM. I tried https://mcgis.mesacounty.us/image/rest/services/Mosaic_Datasets/MesaCounty_2019/ImageServer/WMTS/tile/1.0.0/Mosaic_Datasets_MesaCounty_2019/default/default028mm/{zoom}/{y}/{x} , but that didn't work (the `{x}` and `{y}` coordinates were wildly different from expected, same area as above was looking for a tile with the coordinates 12/1564/812 (same zoom and everything). I couldn't create a WMS url since that needs a `{bbox}`.

That being said, it would be nice if there was a way to specify the default layer for the imagery (JOSM has this, and when the imagery is synced over, I'll check and see if I need to add one).

andrewharvey:
I'm not sure if this helps but take a look at the oceania/australia/act for an example of a less than ideal endpoint. 1

ghost:
@andrewharvey: That did provide a usable WMS service ( url is wms:https://mcgis.mesacounty.us/image/rest/services/Mosaic_Datasets/MesaCounty_2019/ImageServer/exportImage?f=image&format=jpeg&imageSR=3857&bboxSR=3857&bbox={bbox}&size={width},{height}&foo={proj} ).
I'll send the county an email to check and see if that is OK (I've been told that WMTS is more efficient, which means it costs less for the county to run, and they have been fairly consistently worried about decreasing their budget), since it isn't publicly available.

If WMS is really needed, and the county doesn't want people to use non-WMTS endpoints, I can download the imagery and run my own server, but it will probably be a while before I get around to it.

grischard:
WMTS will always use the same image sizes, which can be cached, which makes it a bit faster.

I doubt that we'll create a significant amount of traffic though. The WMS could probably run off a raspberry pi and still be reasonably fast. I've ran `mapproxy` for Luxembourg and a WMS for Kosovo on dinky little VMs that ran without breaking a sweat.

If the county prefers that we don't use the WMS, they could also let their WMTS provide EPSG:3857.

We can also, say, test WMS until March and then re-assess.

See also https://github.com/openstreetmap/iD/issues/4828 for better wms/wmts support in iD.

grischard:
Merging this for now. If anyone wants to convert this to a 'tms' or 'wms' endpoint, I'll be happy to accept another pull request.

grischard:
Thank you @tsmock

PR #735, opened 2019-11-18
Added two missing areas to LPI NSW Topographic Map

GenericError:
Lord Howe Island and Balls Pyramid were missing from the existing geoJSON. This edit adds two additional polygons, one for each of the areas.


andrewharvey (changes requested):
Thanks. Check out https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md.

"Each source must be a GeoJSON Feature"... not a FeatureCollection.

Also per https://github.com/osmlab/editor-layer-index/blob/gh-pages/.editorconfig it would be great if you could use an indent size of 4 spaces for consistency.

GenericError:
Would it be OK to duplicate the geoJSON extent of the NSW Best Imagery file, seeing as they both cover the same area?

Edit: and thank you for pointing out the error

andrewharvey:
> Would it be OK to duplicate the geoJSON extent of the NSW Best Imagery file, seeing as they both cover the same area?

No because they covered different areas, you can see the difference in the current files or at https://osmlab.github.io/editor-layer-index/, eg around the victorian border.

andrewharvey:
I'd suggest using the existing files but just amend the parts that need correcting rather than creating a whole new file.

andrewharvey:
I've implemented #743 which addresses the issues here and includes the extra areas in the topo bounds.

andrewharvey:
Oh and @GenericError thanks for spotting this and for the PR, I just went ahead and implemented this in #743 because I wasn't sure if you still had time to update this PR.

grischard:
Closing in favour of #743, thank you @GenericError!

PR #736, opened 2019-11-18
Update Torokbalint and Budapest XI. district

urbalazs:
New ortophotos are available for Törökbálint and Budapest XI. district.


grischard:
Thank you! Could you please add https://kozigazgatas.ujbuda.hu/adatvedelem - or a better link - as a `privacy_policy_url`?

urbalazs:
I used the same URL as there was added to the previous versions of the ortophotos. The link that you are paste above is the data privacy URL for visitors of the homepage, it is not applied to the ortophotos. We have special usage rights only for OpenStreetMap, and we collect all kind of special usage rights in the [OSM Wiki](https://wiki.openstreetmap.org/wiki/WikiProject_Hungary/Ortofot%C3%B3k).

urbalazs:
@grischard what are you expected from me? I can not provide other privacy policy URL than it is already has in the repository. Nothing changed, but newer images are available under the same license. Please compare the old and the new files, only the date is changed for the images.

andrewharvey:
The privacy policy is different to the license. The license grants exceptions to the copyright allowing others to republish the data. The privacy policy says what happens to your information you send to the server (like IP address, which areas you're looking at etc).

https://kozigazgatas.ujbuda.hu/adatvedelem and http://terinfo.ujbuda.hu are unavailable to me right now so I can't check, but the website privacy policy is probably good enough and can be considered to cover the mapping server too.

andrewharvey (approved):
Looks good to me, thanks.

grischard:
Thank you @urbalazs !

PR #737, opened 2019-11-20
Added topo/outdoors activities tileset for Slovakia

alexanderzatko:
Freemap.sk - made tileset combining topo map with trails for hiking, skiing, cykling and other activities.


grischard:
I really like the freemap.sk style and wish it were available in more places.

Would it please be possible to add the `license_url` and `privacy_policy_url` attributes? I wasn't able to find anything like that on the site.

PR #739, opened 2019-11-20
add Freemap.sk-Outdoor

alexanderzatko:
Based on recommendation from https://github.com/osmlab/editor-layer-index/pull/737 I moved the file to the EU directory


grischard:
Aha, good! Now we still need some kind of license and privacy policy.

alexanderzatko:
We have none of those, but can maybe produce them. However, I noticed that many (most?) sources are missing one of these or both. Has something changed and now it is impossible to publish the sources without values for these properties?

grischard:
Yeah, #439 and #690 - we're very slowly fixing this, and trying not to make the problem worse.

alexanderzatko:
Ok, it makes sense. I will try to produce that info.

grischard:
Thank you. You've also given me fresh motivation to go work on those tickets a bit.

zdila:
Hello @grischard. Can this be merged or are you still waiting for some licence? The map itself is derived from OSM and its license is CC-BY-SA.

grischard:
Hi @zdila! Is that documented on freemap.sk anywhere? Is there a privacy policy somewhere?

zdila:
Hi @grischard. It is mentioned only directly on www.freemap.sk?layers=X when you press © button:

image

Feel free to give me a hint how/where to better document it or what do you expect.

Thanks!

zdila:
Please use https://github.com/FreemapSlovakia/freemap-mapnik/blob/master/limit.geojson instead.

@grischard Please help us with the license and privacy policy. Map is rendered using Mapnik with our own opensource nodejs application and uses OpenStreetMap data and SRTM public domain data. Anybody using this map must obey ODbL 1.0 (because of OSM) and must also attribute Freemap.sk. Nothing more.

zdila:
@grischard should I create a new pull request with links to policy which I will somehow write myself as I am author of the map? Or what should I do to make this source appear in iD? I can do some small donation if it would make things faster...

andrewharvey:
> Please use https://github.com/FreemapSlovakia/freemap-mapnik/blob/master/limit.geojson instead.

If you could update this PR with the new file then I'm happy with this being merged.

> @grischard Please help us with the license and privacy policy. Map is rendered using Mapnik with our own opensource nodejs application and uses OpenStreetMap data and SRTM public domain data. Anybody using this map must obey ODbL 1.0 (because of OSM) and must also attribute Freemap.sk. Nothing more.

If it's using only OSM data and SRTM public domain data then that's fine. I assume you're the owner of freemap.sk and are okay with this being included in the index here?

> @grischard should I create a new pull request with links to policy which I will somehow write myself as I am author of the map? Or what should I do to make this source appear in iD? I can do some small donation if it would make things faster...

If the service doesn't have a privacy policy that's fine, you can just leave this out.

So as far as I'm concerned this is fine to be merged, but if you wanted to update the polygon, feel free to update this PR otherwise I can merge it.

andrewharvey (approved):
If upstream has no privacy policy that's fine we can't link to one that doesn't exist.

Confirmed license should be fine if only public domain SRTM (not one of those processed ones which has stricter usage requirements) and OSM data and the creator of the map is fine for it to be in the index.

zdila:
@andrewharvey yes, it uses public domain SRTM and OSM data and I, as a map creator, am OK with that.

@alexanderzatko will you please update the pull request?

andrewharvey:
I think it's okay for now, so I've merged it, if the boundary polygon needs to be improved can be done in another PR.

zdila:
Thank you. I will create it soon.

Issue #586, opened 2018-11-04
ELI's future


grischard:
I've been thinking about where to take this project next for a while. The current index, which was a perfect solution when we had a dozen layers, is really showing its age. 

* Everyone downloads 1.75 MB of geojson, most of which is in areas they're never going to edit
* Everyone downloads all entries, even if our clients don't support the wms or wmts catalog endpoints.
* It's impossible to specify that a layer is "best" in some places but not others.
* It's impossible to store extra local attributes like imagery offsets
* The index only stores imagery. Local communities are another, incompatible project. For things like address display rules, local language preferences or local defaults, we don't even have a centralised index, and creating one would be hard.
* The [schism with JOSM](https://josm.openstreetmap.de/wiki/ImageryCompare) is nowadays wasting volunteer time.
* The output is statically included in iD and only gets updated when a new version gets released.
* Contributing a new layer can be confusing for newbies.

The solution should use off-the-shelf when possible. This means existing software, existing APIs, code that already works.

The solution should let the clients query about any kind of local features, whether they're imagery, communities, or anything locally relevant we can think of (tagging presets? Validation rules?).

There have been a few similar solutions or attempts: the osm [Imagery Offset Database](http://offsets.textual.ru/), https://github.com/osmlab/osm-community-index/issues/80, maybe the [JOSM chat plugin](https://wiki.openstreetmap.org/wiki/JOSM/Plugins/GeoChat), geocoders if you really stretch the definition.

WFS is possible, but is slow when not [backed by elasticsearch](https://github.com/ngageoint/elasticgeo) and antiquated always. Parsing XML is a pain. Something like [python FeatureServer](https://github.com/iocast/featureserver) (which seems dead) or [JavaScript FeatureServer](https://github.com/koopjs/FeatureServer) (which seems immature) would maybe work. Maybe elasticsearch's [native geospatial support](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-shape-query.html) can do the trick. I really haven't played with any of these yet.

The data could still be integrated from ELI at first, but ideally I'd like something a bit more newbie-friendly to be made available, with a form you can fill out.

What do you think?

simonpoole:
Given that've suggested similar things I suppose I should comment.
> 
> 
> I've been thinking about where to take this project next for a while. The current index, which was a perfect solution when we had a dozen layers, is really showing its age.
> 
[...]

PR #743, opened 2019-11-28
oceania/au/nsw include LHI in TopoMap (#743), thanks @andrewharvey @GenericError

andrewharvey:
this superseeds #735

cc @GenericError


grischard:
Can we simply include the whole ocean? No mappers go there, and it simplifies the polygon.

andrewharvey:
> Can we simply include the whole ocean? No mappers go there, and it simplifies the polygon.

Err I'd rather not since it technically doesn't cover the ocean. It would be misleading to say it's covered there, but in reality it's not. Is there an issue with multiple Polygon rings?

GenericError:
Late follow up, but thank you for going ahead and modifying it. I second that it shouldn't include the ocean unless it is necessary.

grischard:
It's not 'necessary' but the geojson is 2 MB already, and the points don't add any value for our purpose since no one will ever map the ocean.

We already do this for the LPI NSW Base Map, and for many sources elsewhere - see Spain on https://osmlab.github.io/editor-layer-index/ for a good example.

If there is any land in that polygon or anything mappers would potentially want to map, we shouldn't do it of course.

Klumbumbus:
I would use the middle way. Use about 5 nodes for the coastline of australia and one (or two) rectangle(s) for the two islands. This way you have only a handful of nodes more than the current shape, while it is nearly as exact as the first version and it doesn't falsify the shape that much.

andrewharvey:
> It's not 'necessary' but the geojson is 2 MB already,

There are many options to compress this for downstream users who need smaller file sizes without sacrificing quality of the data, like Mapbox Vector Tiles or geobuf or flatgeobuf. So I still think the original one I proposed at https://github.com/osmlab/editor-layer-index/blob/e31f75bb8d9b8bc071f29c0eb71f48c197c37ff5/sources/oceania/au/nsw/NSW-LPI-WebServices-TopographicMap.geojson is the one we should use, it's not like we have the coastline traced out here, it's just a very rugged footprint polygon.

>  points don't add any value for our purpose since no one will ever map the ocean.

They show the footprint more accurately. There are many other sources where have the same level of detail and it's absolutely important for editors to have that precision if they are editing on the boundary of the coverage.

> We already do this for the LPI NSW Base Map

Because the LPI NSW Base Map does cover that area, it's an accurate representation of the footprint.

grischard:
We don't need to show the footprint accurately though - for what purpose would we do that? We need the editors to get the imagery over land, and that's it.

Very good idea for vector tiles - I was thinking it could be part of a solution to #586?

andrewharvey:
@grischard  End of the day each of these files are only 1.7K compressed, that's tiny. I took some time to reduce the coordinate precision in my version at https://github.com/osmlab/editor-layer-index/pull/743/commits/e811a01b70a08b9971fca27f3e7cf0f77852b67a, but I realised I probably still can't convince you so, I've reinstated your version.

I'd still much prefer it not covering the ocean if there is no actual content here, but it's not worth our time to argue over this so let's just merge this.

grischard:
You actually made me laugh out loud. I hope I'm not that stubborn :).

The build script actually simplifies coordinates to, IIRC, five digits after the comma.

It would actually be interesting to see what the biggest source files are.

PR #745, opened 2019-11-29
[WIP] Adding the ability to have more strict checks for files changed in a PR

rbuffat:
This PR adds the ability to apply more checks to sources changed in a PR (e.g. to require a privacy policy for modified sources). It is currently a work in progress (respectively just a proof of concept), and I suppose there will be a need for discussions.  

This index contains a vast amount of sources. If requirements are updated (e.g. to have a privacy policy), it is not feasible to require that all existing sources meet these new requirements on the spot. However, it would be feasible to enforce these new requirements for new, respectively changed, sources. 

This PR adds the following functionality:
- Extends checks.py with a "strict" check mode (Probably, there is a better name for it)
- Add tests for a missing privacy policy, which is tested in the strict mode
- Apply checks.py in strict mode to all files changed with a PR
- Only for testing purposes: adds a source with missing privacy policy (this needs to be reverted before merging)


rbuffat:
it looks like as there are differences in the behavior of "git diff" depending on if it is executed on a fork or not (https://travis-ci.org/rbuffat/editor-layer-index/builds/618399862)

grischard:
Wow, very cool. Many thanks for starting on this.

rbuffat:
@grischard  It looks like it works.  I also added a test to check if the license url actually exists when strict mode is enabled. 

I currently see the following issues:
- When we require the privacy url, then this should be documented somewhere (E.g. contributing.md) 
- What happens if a source has no privacy url? Should then "privacy_policy_url" be set to None or just be ignored?

grischard:
I finally had the chance to review this, and it's so much simpler than I thought. You did so many simple and clever things. Thank you so much.

I'll update contributing.md separately, and we can see about an `if privacy_policy_url != 'None'` when we have that problem.

PR #753, opened 2019-12-20
updated license url for 2 added 1 new resource

DcJoeS:


andrewharvey:
According to the license

>  If you transmit or provide the GIS Data (or any portion of it) to another user, you must provide a copy of this disclaimer and the accompanying metadata for this dataset to the user.

IANAL but is that enforceable? I think we should get some advice on if this is allowed to be used in OpenStreetMap with that constraint.

DcJoeS:
Yeah, me neither, but this is supposed to suffice as it is noted here:  https://wiki.openstreetmap.org/wiki/Metropolitan_Council

grischard:
Thank you @DcJoeS! Would it please be possible to add https://www.co.dakota.mn.us/Policies/Pages/default.aspx as a privacy policy url?

grischard:
Looks good, thank you @DcJoeS !

andrewharvey:
It has been raised in #946 that specifically the 2019 fall imagery here has licensing constraints which make it unable to be used for tracing in OSM. https://resources.gisdata.mn.gov/pub/gdrs/data/pub/us_mn_co_dakota/base_aerialphotography_2019_fall/metadata/metadata.html

So unless there is some other discussion of this where it can be pointed out it can be used, we'll need to remove this one from the index and really all data already traced with it should be removed from OSM out of caution.

rbuffat:
@grischard @andrewharvey I confused the fall and the spring imagery. This PR included the spring imagery which is not from Nearmap: https://resources.gisdata.mn.gov/pub/gdrs/data/pub/us_mn_co_dakota/base_aerialphotography_2019_sprg/metadata/metadata.html

PR #758, opened 2019-12-30
Fix icon URL

magol:


Klumbumbus:
The icon is pretty big. In JOSM I used the smaller icon variant commons provides: https://josm.openstreetmap.de/wiki/Maps/Sweden?sfp_email=&sfph_mail=&action=diff&version=91&old_version=89&sfp_email=&sfph_mail=

And the polygon should be closed.

magol:
@Klumbumbus  thanks for the comments. I have updated the commit according to you suggestions

PR #761, opened 2020-01-12
Categorize all global layers

simonpoole:
Does what the title says. I didn't sync this with JOSM but suspect it is essentially the same, as the global layers are simple (compared to the national ones).


Klumbumbus:
There is one difference. I used "osmbasedmap" for the Locator overlay in JOSM. "qa" doesn't really fit for this layer.

grischard:
Synced in cc74fe4, thank you both.

PR #762, opened 2020-01-12
strict tests modifications

rbuffat:
I switched to a different git diff command. I'm not 100% sure that this will cover all cases correctly.

The strict tests are only executed if git diff returns something. The script does not fail when nothing is returned.  Thus, for now, the logs should be checked if the strict checks are executed on the right files.


grischard:
I think that's going in the right direction. I've added a check to (hopefully) only run on PRs, based on https://denysdovhan.com/building-pull-requests-using-travis/.

grischard:
Actually, this is ready to go, no? Merging to see if it breaks anything :). Thank you very much @rbuffat.

rbuffat:
It looks as there are still some issues:

The http request for the privacy policy of https://travis-ci.org/osmlab/editor-layer-index/builds/645937685?utm_source=github_status&utm_medium=notification fails because it looks as the webserver rejects http requests without a user agent.

The following code works:
```
headers = {'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36'}
r = requests.get("https://www.turku.fi/turku-tieto/tietosuoja/kayttotiedot-ja-evasteet",  headers=headers)
```

grischard:
Ahh, that's what's happening. Fixed in 91fa444.

rbuffat:
@grischard  Thanks for fixing it!

PR #765, opened 2020-01-19
open in JOSM using remote control

andrewharvey:


grischard:
The code looks good; is there an easy way of testing it?

andrewharvey:
`python -m SimpleHTTPServer` here when checked out on this branch, then you should be able to go to `http://localhost:8000`

grischard:
Ah, that's clever. If you've tested it, let's merge it!

andrewharvey:
great, it's now live at https://osmlab.github.io/editor-layer-index/

PR #766, opened 2020-01-19
add maxar open data for australian bushfires pre-event

### LINKED ISSUE #763: Maxar Australian Bushfire Imagery
Maxar have generously donated some satellite imagery of Australian bushfire affected
areas and they explicitly allow tracing for OSM,
https://blog.maxar.com/open-data-program/2020/open-data-response-to-the-australian-wildfires.

I've posted to the local mailing list on how to use these layers in JOSM and iD https://lists.openstreetmap.org/pipermail/talk-au/2020-January/013499.html.

Although it would be great to have these work out the box, I'm not sure ELI is quite setup to deal with this because:

- There are a lot of different layers (covering different dates), is this a problem?
- Because iD only ingests updates each release, even if I keep ELI master up to date, end iD users would rarely see current imagery layers (though JOSM users who point their imagery sources to ELI would).

Is it worth indexing this in ELI?

@grischard I know you had concerns about file size, but I think the more sources the better, if downloads become too large that's a good problem to have and can be solved in other ways on the software level, not the ELI index level.

andrewharvey:
closes #763

At https://www.digitalglobe.com/ecosystem/open-data/australia-wildfires under license it says:

> OpenStreetMap Exception
> In the licensing terms, DigitalGlobe has a stated exception for OpenStreetMap. OSM is allowed to use imagery released under the Open Data Program. Even though OSM enables commercial usage, OSM is a critical platform for the disaster response community. Allowing usage of our imagery in OSM will enable faster data access by local communities as well as global relief efforts.


grischard:
Thanks @andrewharvey

sources/oceania/au/nsw/Maxar-1920-Bushfires-PreEvent.geojson
```diff
@@ -0,0 +1,134 @@
...
+        "license": "CC-BY-NC-4.0",
+        "license_url": "https://www.digitalglobe.com/ecosystem/open-data/australia-wildfires",
+        "attribution": {
+            "url": "https://www.digitalglobe.com/ecosystem/open-data/australia-wildfires",
+            "text": "©2020 DigitalGlobe",
+            "required": true
+        },
+        "url": "https://cogeo.xyz/tiles/6b6f479fbacd9a42e9e38bc5c0c6889f009beae13b07742ec4a1648f/{zoom}/{x}/{y}.png",
+        "max_zoom": 19,
+        "min_zoom": 0,
```
vincentsarago, line 18: @grischard the `min_zoom` for this layer is zoom 11 see: https://cogeo.xyz/tiles/6b6f479fbacd9a42e9e38bc5c0c6889f009beae13b07742ec4a1648f/tilejson.json

grischard: Fixed in 92fa1ed, thank you Vincent!

don-vip:
What is this? A map of bushfires that occurred in 1920?

vincentsarago:
😬 I guess the name `Maxar 1920 Bushfires Pre-event` should be updated to `Maxar 2019/2020 Australia Bushfires Pre-even` or something

andrewharvey:
It cover's the 2019-2020 Australian bushfire season including the state of NSW only, before and after satellite imagery. Maybe I short-handed the name too much. Happy if it's extended.

don-vip:
`Maxar 2019/2020 Australia Bushfires` is enough for me. I find "pre-event" hard to understand and translate.

andrewharvey:
Though when more post event imagery is released it's nice to have both so they'll need different names. Does this need to be translated?

don-vip:
In JOSM we translate all imagery names. Don't the other editors do the same?

andrewharvey:
Ok so pre event means imagery from before the bushfires and post event means after the bushfires.

don-vip:
ok thanks. It was especially difficult when I considered 1920 to be a date. With 2019/2020 it is easier to understand :)

andrewharvey:
name update proposed in #772

PR #767, opened 2020-01-19
 Slovenia road lines: updated copyright year in attribution, fixed links

stefanb:
Updated road line data with fossgis/wms.openstreetmap.de@6d092f2
Fixed attribution links

### DIFF
--- a/sources/europe/si/GURSBuildingoutlines.geojson
+++ b/sources/europe/si/GURSBuildingoutlines.geojson
@@ -718,7 +718,7 @@
         "attribution": {
             "required": true,
             "text": "CC-BY \u00a92020 Geodetska uprava Republike Slovenije (gu.gov.si).",
-            "url": "http://egp.gu.gov.si/"
+            "url": "https://egp.gu.gov.si/"
         },
         "country_code": "SI",
         "icon": "https://osmlab.github.io/editor-layer-index/sources/europe/si/GURSRoadlines.png",
--- a/sources/europe/si/GURSBuildingoutlines.geojson
+++ b/sources/europe/si/GURSBuildingoutlines.geojson
@@ -717,8 +717,8 @@
     "properties": {
         "attribution": {
             "required": true,
-            "text": "CC-BY \u00a92020 Geodetska uprava Republike Slovenije (gu.gov.si).",
-            "url": "https://egp.gu.gov.si/"
+            "text": "CC-BY \u00a92020 Geodetska uprava Republike Slovenije (gov.si).",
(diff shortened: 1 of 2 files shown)


simonpoole:
We now have a "category" field as JOSM has that would be good if it gets added on any changed of existing sources, see https://github.com/osmlab/editor-layer-index/pull/733

grischard:
We can also simply drop the year?

stefanb:
@grischard, it is good to have the year visible, to remind us (OSM mappers and that tile server maintainers) when the data was obtained and valid, so we can easier make decisions when mapping.

Also http://www.e-prostor.gov.si/fileadmin/struktura/ANG/General_terms.pdf mentions:

> The data user shall indicate [...] date of the database

So that is giving them the attribution closest to what is reasonably possible.

In OSM the attribution is given with https://wiki.openstreetmap.org/wiki/Tag:source%3DGURS on changesets and/or map elements.

stefanb:
@simonpoole, not sure how to categorise these tiles. They are both vectors rendered from their official shapefiles, not really a full "map". Should they be categorised as `map` or `other`, or something completely different?

stefanb:
@simonpoole, also they are both mostly transparent, so they are intended to be used as an overlay.

stefanb:
Both can be viewed at
* https://wms.openstreetmap.de/slippymap/GURS-building-outlines
* https://wms.openstreetmap.de/slippymap/GURS-road-lines

grischard:
Thank you @stefanb, sorry this was in limbo for a bit. You're very welcome to add category in another PR, but I'm merging it as it is for now.

stefanb:
Thanks!
I will categorise the layers as soon as I understand the categories better. @simonpoole, any hints or pointers to category documentation / examples for similar layers?

grischard:
It seems like there's a good case to be made for either 'map' or 'other'. Since you wouldn't use this as a full map, and it's more of a map extract, I'd go for 'other', which also seems to be how JOSM categorises such things.

stefanb:
Basically, they are both very specialized thematic maps, with transparent background, to be used as overlay on any regular map or ortophoto when editing OSM data (with ugly inscription on each tile to discourage any other general use).

grischard:
Would 'other' seem reasonable to you then?

stefanb:
Well, i see "other" same as not having the category set at all and am fine with either. Is there any practical difference? Any known apps / services using the category?

simonpoole:
Imho vespucci 14.1 is the only app outside of JOSM that uses the categories, others are however sure to come.

PR #773, opened 2020-01-21
Add category attribute to most German sources

simonpoole:
This categorizes most of the German sources.

Notes:

- we currently haven't agreed how to classify IR and NIR imagery
- it would in general be really helpful if the description attribute was present, as is, I'm not quite sure how anybody, not to mention newbies could make sense out of lot of the layers.

### DIFF
--- a/sources/europe/de/Bavaria(80cm).geojson
+++ b/sources/europe/de/Bavaria(80cm).geojson
@@ -4,6 +4,7 @@
     "id": "bavaria-80cm",
     "name": "Bavaria (80 cm)",
     "type": "wms",
+    "category": "photo",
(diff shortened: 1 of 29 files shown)


Klumbumbus:
There are three differences:

````
* Different categories ('map' != 'other'): [DE] Deutsche Bahn VzG lines January 2017
* Different categories ('map' != 'other'): [DE] Deutsche Bahn VzG lines Nov 2015
* Different categories ('historicphoto' != 'photo'): [DE] Munich latest aerial imagery 60cm
````
https://josm.openstreetmap.de/wiki/ImageryCompare

simonpoole:
I would argue that the JOSM classification of the Munich imagery is wrong, as it is clearly outdated (5 years old).

Klumbumbus:
Meanwhile it is from 2017 not 2015 anymore, so ~2.5 years. I updated the entry in JOSM https://josm.openstreetmap.de/wiki/Maps/Germany?action=diff&version=211

(Even if it would be still from 2015 I would still classify it as photo, as 2015 is not sooo old, there is no newer source available of that area and the global sources seem all to be older too.)

PR #777, opened 2020-01-22
add environment canterbury imagery

andrewharvey:


andrewharvey:
let's hold off merging, since the resolution seems to be capped...

grischard:
You could also replace the coastline in the polygon with one big triangle that englobes the coast but no other landmass.

Should we just close this if the resolution is junk?

andrewharvey:
> You could also replace the coastline in the polygon with one big triangle that englobes the coast but no other landmass.

I don't like that and think it looks ugly, a couple extra nodes to make it look nicer isn't going to make a difference in size in the grand scheme of things.

> Should we just close this if the resolution is junk?

It only seems to be /tile URL which is capped, when you load it into QGIS as an ESRI service it works with the full resolution, so it could still be salvageable if we can work out if it's possible to make the same kind of call that ELI supports.

grischard:
Hah, I know you don't like it :).

Let's keep it open then.

rbuffat:
@andrewharvey The folder https://gisbasemap.ecan.govt.nz/arcgis/rest/services/Imagery/Imagery_basemap_WGS84_4 seems to not exist anymore. But there seems to be other imagery available: https://gisbasemap.ecan.govt.nz/arcgis/rest/services/Imagery

andrewharvey:
I think the issue here @rbuffat is all those other endpoints don't offer a projection that iD could support (unless that's changed). I guess we can still add them anyway.

rbuffat:
oh, I didn't check this. Would it be worth a shot to just ask them if they would consider enabling the WMS or REST API, respectively, adding an EPSG:3857 WMTS tileset?

rbuffat:
@andrewharvey The ESRI REST API as WMS hack seems also to work with the [following imagery](https://gisbasemap.ecan.govt.nz/arcgis/rest/services/Imagery/Canterbury_Imagery_2010_2015/MapServer):
```
"type": "wms",
"url": "https://gisbasemap.ecan.govt.nz/arcgis/rest/services/Imagery/Canterbury_Imagery_2010_2015/MapServer/export?f=image&format=png&imageSR=3857&bboxSR=3857&bbox={bbox}&size={width},{height}&foo={proj}",
"available_projections": [
    "EPSG:3857"
],      
```

andrewharvey:
@rbuffat Good idea. I just checked the waiver and realised I really limited it's scope to fixed endpoints https://wiki.openstreetmap.org/wiki/File:EnvironmentCanterbury_AerialimagerywaiversfortracingforOpenStreetMap.pdf though the mosaic is covered, so I'll use that. This should be good now.

PR #782, opened 2020-02-02
Remove Juno NYC traces, thanks @LucasLarson and @zverik

LucasLarson:
Juno was a rideshare company that published the previous two days of its drivers’ traces in New York City to ~~`https://gps-tile.junolab.net`~~. In late 2019, the company folded and removed the service from their website.

This pull removes @GoJuno’s layer from the index (and undoes @Zverik and @grischard’s [commit eab607b](https://github.com/osmlab/editor-layer-index/commit/eab607b058d79b1e300131bff30d9217e845c196)).


Zverik:
I agree with this: could not find a place to upload the latest docker container.

PR #786, opened 2020-02-05
Replace Ville de Nyon aerial imagery (2010->2016)

yblatti:
The city of Nyon (CH) offers a new version of it's aerial imagery. Hosted by ASIT VD.
Flight date: 2016-03-10 (https://www.asitvd.ch/md/1c473b39-61e0-5694-b5c4-c5c5b3ce62f1).
2010 files will be retired in a few months.


andrewharvey:
It can be handy to have the older imagery available as a reference. If it's still a few months away from being retired, it wouldn't hurt to leave it in place and take it down when it's retired. Though if it's easier for you to take it down now that's okay too.

andrewharvey:
What license is this imagery under and where can I find more details about the OSM permission?

ref https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L82-L89

yblatti:
> 
> 
> It can be handy to have the older imagery available as a reference. If it's still a few months away from being retired, it wouldn't hurt to leave it in place and take it down when it's retired. Though if it's easier for you to take it down now that's okay too.

We'd like to retire the old tiles as soon as possible, we'll just keep them the time needed for the new ones to be deployed to all editors.

yblatti:
> 
> 
> What license is this imagery under and where can I find more details about the OSM permission?
> 
> ref https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L82-L89

This imagery if offered by Ville de Nyon to the OSM community for mapping, but is not distributed under any open data license. Should we use `"license": "COMMERCIAL"` and `"permission_osm": "explicit"` in this case ?

yblatti:
We can ask @rbovard from @sitnyon to confirm...

rbovard:
Thanks @yblatti for your work.

As you say, the City of Nyon would like to make the 2016 orthophoto available to the OSM community for mapping (as we did with the 2010 edition). This is not open data, but an additional resource for the project.

@andrewharvey What kind of license should we use in this case?

Otherwise I don't think we should keep the 2010 orthophoto any longer than necessary, as the last one has a larger coverage.

andrewharvey:
- No problem removing the 2010 source if you want to retire the source as soon as possible.
- It's fine if the imagery is not open data but in that case per https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#what-imagery-licenses-are-compatible-with-this-index we need the aerial imagery tracing waiver https://drive.google.com/file/d/0B3PN5zfbzThqLXg1TUlxalAtVE0/view. Here is an example of one completed by another organisation https://wiki.openstreetmap.org/wiki/File:OSM_waiver_AirborneResearchSouthAustralia.pdf
- Then `license_url` should be to where this tracing waiver permission can be found.
- `license` can be set to `COMMERCIAL` to indicate it's not open data, it's the closest tag we have at the moment for "All rights reserved".

rbovard:
Thanks a lot for those informations @andrewharvey.

I've uploaded the waiver here: https://wiki.openstreetmap.org/wiki/File:OSM_waiver_Nyon.pdf

@yblatti Could you please update the PR according to the discussion?

andrewharvey (approved):
Looks good, thanks!

yblatti:
Awesome, thanks @andrewharvey , @rbovard & @sitnyon!

grischard:
Thanks everyone!

Keep in mind that the iD editor doesn't update its layers right away ; you'll still get http requests for the 2010 tiles until a new version of iD is released.

andrewharvey:
In the meantime iD users can use the URL as a custom layer (but then loose a lot of the benefits here (bounds, min/maxzoom, proper source reference).

PR #787, opened 2020-02-06
Stricter checks for TMS layers

grischard:
* Check whether layer is responding at min/max zoom, on all mirrors
* Check whether max/min zoom could be expanded (but sometimes you just get a blank tile or zoomed in pixels)
* Tighten some wms checks while I was at it


rbuffat:
@grischard Do you intend to work more on this PR or should we close it? I think at least some of the modifications are included in the strict check.

grischard:
It wasn't a very good idea because the coverage polygons often go outside of the imagery to save space. If you see nothing interesting in it, we can close it.

rbuffat:
@grischard I will go over it and try to integrate missing things in the check / strict check if I find any.

Regarding the coverage polygons, I fear there is no one-fits-all solution. In some cases, a coverage as close as possible to the available imagery is favored and some favor an approximate solution that minimizes the file size. 

I see here the following options:
- As an alternative, we could include a region property that contains an id for the region (e.g. the name of the country or admin level and name etc.) that approximates best the coverage of the imagery.
- Implement simplification algorithms that simplify geometries if a minimal file size is preferred.

grischard:
We don't want to simplify geometries always. For example, people living close to the French border in Luxembourg sometimes end up using the older French layer, because it's marked as best too.

An algorithm that simplifies polygons only over the sea would be a pretty cool thing to have.

I was thinking we could simply have a 'test' position in layers, and use a centroid-ish point otherwise.

simonpoole:
> We don't want to simplify geometries always. For example, people living close to the French border in Luxembourg sometimes end up using the older French layer, because it's marked as best too.

Users making a non-informed decision on which imagery to use would seem to be completely independent of the question of if the boundaries are simplified or not. Actually one of the problems is that often the imagery boundaries in ELI don't actually correspond to the coverage boundaries, but are often the slightly simplified admin boundaries. 

I don't think there is a real solution to the issue currently given the current data model and the desire to keep users in the dark.

A very current example btw: http://maps.zh.ch?topic=OrthoZH&scale=20000&x=2670553.51&y=1249294.67&srid=2056&offlayers=ortho_s_fcir_2014%2Cortho_s_2014%2Cortho_w_fcir_2015%2Cortho_w_2015%2Cortho_s_fcir_2018%2Cortho_s_2018%2Cortho_s_fcir_2020%2Cortho_s%2Cortho_s_fcir%2Cortho_w%2Cortho_w_fcir%2Cortho%2Clk500%2Clk200%2Clk100%2Clk50%2Clk25%2Cup8%2Cup24%2Ckachelung_2014%2Ckachelung_2015%2Ckachelung_2018%2Ckachelung the 5cm/pixel imagery that extends far into the Kanton Aargau. Obviously it would be advantageous to use the actual coverage boundary for the is imagery.

rbuffat:
@grischard I went over your changes to the check script and I think the only thing that is currently missing is the check for the magic byte. I will try to integrate this in the strict_check soon. The rest seems already be included in the strict_check. 

@grischard @simonpoole The imagery boundaries problem is probably something without a solution that works for all cases.  Something I was thinking about would be to use for the geometry element in the geojson the exact boundaries of the imagery and optionally having a property containing an id of the best matching corresponding administrative boundary. An OSM editor could then maintain the administrative boundary geometry in the resolution he finds suitable for his use case. But this probably would require to maintain a dataset with the geometries for these administrative boundaries within ELI. Also, this would require editors to support MultiPolygons. Also, an issue with this approach is that imagery available outside of the area of interest of the buyer of the imagery might be of lower quality, e.g. regarding orthorectification.

PR #798, opened 2020-02-26
orthophoto Ukraine - Mykolaiv 2018

bezdna:


sources/europe/ua/Ukraine-Mykolaiv2018(mbk).geojson
```diff
@@ -0,0 +1,2569 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "UkraineMykolaiv2018",
+        "name": "Ukraine - Mykolaiv 2018",
+        "type": "tms",
+        "url": "http://mbk.mkrada.gov.ua/map/rtile/carto_2013928517078615452/ua/{zoom}/{x}/{y}.png",
+        "category": "photo",
+        "permission_osm": "explicit",
```
andrewharvey: could you also include a license_url where more details about permission to use this data in OSM can be found?

andrewharvey (changes requested):

sources/europe/ua/Ukraine-Mykolaiv2018(mbk).geojson
bezdna: there is no specific indication of the license, well, there is a law of Ukraine that allows the use of open data in the public domain and use it for any purpose, including commercial

grischard:
Hi @bezdna! Looking very good.

The polygon looks very detailed, and everyone who ever edits OpenStreetMap anywhere in the world will have to download it. Since I don't have access to an Ukraine proxy, I can't check what the actual imagery is like. The polygon is used to show what layers are available in an area, and don't have to be pixel-perfect. Do you think the polygon can be simplified?

sources/europe/ua/Ukraine-Mykolaiv2018(mbk).geojson
andrewharvey: @grischard another reason why we need that "license" field reinstated, so in this case it can be set to `PUBLIC_DOMAIN` to keep track of why this data can be used in OSM.

bezdna:
> Hi @bezdna! Looking very good.
> 
> The polygon looks very detailed, and everyone who ever edits OpenStreetMap anywhere in the world will have to download it. Since I don't have access to an Ukraine proxy, I can't check what the actual imagery is like. The polygon is used to show what layers are available in an area, and don't have to be pixel-perfect. Do you think the polygon can be simplified?

polygon simplify.

Tested with VPN, like a proxy is not necessary

grischard:
Looks good to me then. Pity that you need an Ukrainian IP address to access it - someone using a VPN or roaming on mobile will get broken images, but I suppose that's a minor problem. It will also mess up the http checks of our JOSM friends - CC @Klumbumbus. We can always mapproxy it or something if necessary.

grischard:
Thank you again @bezdna and good luck with your PRs on the openstreetmap website :)

Klumbumbus:
>Pity that you need an Ukrainian IP address to access it - someone using a VPN or roaming on mobile will get broken images, but I suppose that's a minor problem. It will also mess up the http checks of our JOSM friends - CC @Klumbumbus. 

It doesn't need an Ukrainian IP. Works fine for me from Germany. https://josm.openstreetmap.de/mapsview?entry=Ukraine%20-%20Mykolaiv%202018

PR #800, opened 2020-02-28
add Ukraine - Vinnytsia 2018 and set default for Mykolaiv

bezdna:


grischard:
I don't think you want 'default', which will show this in the world-wide menu - what are you trying to achieve? Maybe 'best' to mark this as the best for the region?

bezdna:
> I don't think you want 'default', which will show this in the world-wide menu - what are you trying to achieve? Maybe 'best' to mark this as the best for the region?

yes you are right 
I will fix

grischard:
Thank you @bezdna !

Klumbumbus:
The shape of Vinnytsia is unclosed.

bezdna:
> The shape of Vinnytsia is unclosed.

changed here https://github.com/osmlab/editor-layer-index/pull/802

PR #803, opened 2020-03-04
Add Gothenburg orthophoto

NKAmapper:
Product description: [link](https://catalog.goteborg.se/store/6/resource/2691)
Licence: CC0

Will replace the [Gothenburg city map](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/se/gotenburg_citymap.geojson) as best in ELI


grischard:
Cool, thank you @NKAmapper!

PR #804, opened 2020-03-05
Create Frankfurt-am-Main-2017.geojson

grischard:
Not tested


bhousel:
> Not tested

I tested and it looks good to me..

You can test an ELI file here:  https://ideditor.github.io/imagery-index/index.html
This site is a good test because it reuses the actual iD code to create the tile layer.

grischard:
That's really cool. Does it take URL parameters?

bhousel:
> That's really cool. Does it take URL parameters?

Not yet, but it's a good idea..

PR #807, opened 2020-03-08
move Kanton_Solothurn_SOGIS_DSM_2019 into sources/europe/ch

andrewharvey:
#806 added this one too the root directory, this pr puts it into the ch directory

cc @voongit @grischard


grischard:
Perfect!

PR #810, opened 2020-03-08
Create Kanton_Solothurn_SOGIS_DTM_2018.geojson

voongit:
I saw, that the 2019 relief only provides the southern half of the canton. The 2018 relief is the north, including most of Basel Land as well. This would also require modifications to the polygon .. but I do not know how to do that.


rbuffat:
@voongit  Would you be willing to test the following jupyter notebook: https://mybinder.org/v2/gh/rbuffat/eli-helper/master?urlpath=apps%2Fcreate_simplified_geometry_appmode.ipynb

You could build your query using the [districts](https://de.wikipedia.org/wiki/Kanton_Solothurn#Bezirke): e.g. "Bezirk Olten; Bezirk Gösgen".

andrewharvey:
Is this one still needed now that we have https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/Kanton_Solothurn_SOGIS_DTM_2019.geojson?

voongit:
Hi there

I think so, because both DTM and DOM for Solothurn cover the Cantone partially (opposed to the older ones, that covered all). If you load them somewhere, you should see them cover half the area each ... I guess they were made over a two year project.

Cheers,

Philippe






Am Sonntag, 6. September 2020, 02:55:07 MESZ hat Andrew Harvey <notifications@github.com> Folgendes geschrieben: 







Is this one still needed now that we have https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/Kanton_Solothurn_SOGIS_DTM_2019.geojson?

Klumbumbus:
A elevation model shouldn't be marked best, and the icon url is dead.

PR #815, opened 2020-03-13
Implement JOSM attributes

grischard:
See https://josm.openstreetmap.de/ticket/17285#comment:15

`mirror` is still missing.


simonpoole:
While I believe having equivalent attributes in ELI and JOSM is a good idea, could we confirm that these are all actually in use in the JOSM side and not things that were introduced for a specific case and then never reused?

Marc-marc-marc:
> could we confirm that these are all actually in use in the JOSM side

it is the trend to request "confirmation of use" before a reconciliation :(
what was the defect? the size of the filetransfert for Vespucci? the size on the smartphone? or is it just to keep a schema with no unused elements?

ping @stoecker @Klumbumbus for the ack

stoecker:
"transparent" and "format" are relatively new and currently no wms_endpoint entries use them. But they are needed for proper support of wms_endpoint.

"custom-http-headers" usually will probably be used extremely seldom for "official" services, that's more a feature to get some strange services to work. It is needed from time to time, but at least I'm always happy if it works without.

"last-check" is used, but not as active as it should be - I'm not sure, maybe that idea is a failure and obsoleted by the wiki preview and the automatic checks we do nowadays.

Everything else is actively used.

simonpoole:
> 
> 
> > could we confirm that these are all actually in use in the JOSM side
> 
> it is the trend to request "confirmation of use" before a reconciliation :(

Just to avoid us adding something, potentially adding support for it in code and JOSM saying "this hasn't really worked" ... definitely asking for a quick "OK" is sensible.

stoecker:
Regarding last-check: https://josm.openstreetmap.de/ticket/18926

stoecker:
last-check is gone now. We have better (i.e. working :-) mechanisms now to server the intended purpose to check the validity of the dataset.

grischard:
Oh great, now you've broken my PR :-)

grischard:
Merging this as it is. I'll admit the limits of my intellect when it comes to having the mirror tag in the json schema, and happily take any PRs.

simonpoole:
As I understand it that is simply an alternative site for the same data, so shoud be modeled as a json object mirror containing some of the fields with different values. What needs to be checked if one of these objects is enough or if we need an array of them.

don-vip:
> As I understand it that is simply an alternative site for the same data

Yep.

grischard:
> What needs to be checked if one of these objects is enough or if we need an array of them.

Some like https://josm.openstreetmap.de/wiki/Maps/Czech%20Republic#PrahaIPRlow-vegetationorthophoto have two mirrors.

Klumbumbus:
You forgot ``logo-url`` in this PR?

grischard:
I did.

schema.json
```diff
@@ -172,7 +172,86 @@
...
+                    "description": "Specify a checksum for tiles, which aren't real tiles. `type` is the digest type and can be MD5, SHA-1, SHA-256, SHA-384 and SHA-512, value is the hex encoded checksum in lower case. To create a checksum save the tile as file and upload it to e.g. https://defuse.ca/checksums.htm.",
+                    "type": "string"
+                },
+                "metadata-header": {
+                    "description": "header-name attribute specifies a header returned by tile server, that will be shown as `metadata-key` attribute in Show Tile Info dialog",
+                    "type": "string"
+                },
+                "valid-georeference": {
+                    "description": "Set to `true` if imagery source is properly aligned and does not need imagery offset adjustments. This is used for OSM based sources too.",
+                    "type": "string"
```
don-vip, line 198: why not boolean?

grischard: Uh huh. Why not!

PR #816, opened 2020-03-15
Add/correct JOSM attributes

grischard:
Follow-up of #815

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -219,6 +219,50 @@
                     },
                     "additionalProperties": false
                 },
+                "mirror": {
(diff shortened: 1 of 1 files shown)


simonpoole:
If https://json-schema.org/understanding-json-schema/structuring.html is correct you should put the elements that can be present twice in a definition block and reference (DRY).

The mirror (I don't like that name because it implies something that typically isn't true, the servers are simply using the same source, so "alternative" or something like that would be better, but OK) objects should be in an array as there can be multiple.

grischard:
TODO from comments on #815:

- Add `logo-url`
- ~~Set `valid-georeference` to Boolean~~

andrewharvey:
> Set valid-georeference to Boolean

this was done in #824

PR #818, opened 2020-03-24
Update Czech pLPIS based on recent JOSM change

Jakuje:
JOSM change:
https://josm.openstreetmap.de/wiki/Maps/Czech%20Republic?sfp_email=&sfph_mail=&action=diff&version=41&old_version=40&sfp_email=&sfph_mail=


grischard:
Perfect, thank you!

PR #820, opened 2020-04-08
Croatia update layers

hbogner:
Increase max_zoom for Knin and Zagreb
Zoom is available up to lvl 20.
Simplify area for dof.
Add cathegory


hbogner:
Any progress on merging this?

PR #821, opened 2020-04-12
Kentucky Aerial Photography and Elevation Data Program

1ec5:
Added two WMS orthoimagery sources from Kentucky Aerial Photography and Elevation Data Program (KYAPED), commonly known as KyFromAbove:

* “KyFromAbove” is a mosaic of 6-inch and 1-foot leaf-off orthoimagery taken between 2012 and 2014 in [these 38 counties](http://overpass-turbo.eu/s/SI3) in western and central Kentucky, plus parts of Bullitt and Meade counties in the vicinity of [Fort Knox](https://en.wikipedia.org/wiki/United_States_Bullion_Depository). 💰
* “KyFromAbove 2019 6in” contains 6-inch leaf-off orthoimagery taken last year in [these 63 counties](http://overpass-turbo.eu/s/SI4) covering all of eastern Kentucky.

There is some overlap between the two layers, but the second layer is much more up-to-date. It’s the first 6-inch imagery ever released publicly for most of eastern Kentucky, a region that still needs a lot of attention in OpenStreetMap. It would be a “best” layer except for the Lexington area, where Mapbox has higher-resolution coverage. It was released sometime in the past month, and I can’t wait to start tracing from it as soon as iD picks it up.

Fixes #612.

PR #823, opened 2020-04-17
Do not verify ssl connections in case of failure

rbuffat:
In https://github.com/osmlab/editor-layer-index/pull/821 verification of a SSL connection fails. For the purpose of ELI, an SSL connection is not strictly required.

With this PR, first, an SSL connection using up-to-date Root Certificates (using the certifi library) is tried.
If this connection fails, as fallback a connection without SSL verification is tried. 

@grischard  I did not extensively test this, please have a careful look at it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ pyyaml>=4.2b1
 transifex-client==0.13.7
 colorlog==3.1.4
 requests==2.22.0
+certifi
--- a/scripts/check.py
+++ b/scripts/check.py
@@ -17,7 +17,7 @@ Suggested way of running:
 find sources -name \*.geojson | xargs python scripts/check.py -vv
 
 """
-
+import certifi
 import json
 import io
 from argparse import ArgumentParser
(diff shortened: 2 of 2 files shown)


stefanb:
A sane workflow for cases when the TLS certificate of the service is not trusted would be to provide a public key in the source json description, so that other services can use the service while validating the certificate, not just ignoring it altogether.

rbuffat:
@stefanb This check is only applied to sources changed in a PR. The goal of it is to provide feedback to the creator of the PR if the URL exists or not (e.g. if there is a typo). There is never the need for transmission of any data. 

In my opinion, requiring a public key for a certificate, that can change at any point and will never be tracked, does not provide a significant benefit. 

The best way imho would be to fix why requests fails to verify "https://kentucky.gov" in https://github.com/osmlab/editor-layer-index/pull/821, while common web browsers don't. 
I don't see at the moment a solution for that, maybe you can look at it.

stefanb:
https://www.ssllabs.com/ssltest/analyze.html?d=kentucky.gov
> This server's certificate chain is incomplete. Grade capped to B.

1ec5:
Thank you for the analysis, @stefanb. I’m not sure what to do at this point about #821. For what it’s worth, the tile server at kyraster.ky.gov [also earns a B](https://www.ssllabs.com/ssltest/analyze.html?d=kyraster.ky.gov) for supporting TLS 1.0–1.1, but it has no chain issues. Only the privacy policy at kentucky.gov has chain issues.

As @rbuffat points out, the reachability check is only intended to guard against typos at the time that a PR is authored. The check runs only in PR builds, not push builds, so this repository provides no guarantee about the ongoing reachability of the privacy policies it lists. (It also doesn’t prevent existing layers from pointing to unencrypted HTTP tile server URLs.)

> provide a public key in the source json description

So I’d follow the links in [this post](https://stackoverflow.com/a/46604917/4585461) and paste the keys into the JSON somewhere, correct? Since the reachability check doesn’t run on push builds, is there a risk of this data silently going stale?

If the privacy policy’s certificate is a showstopper, then I suppose a simpler workaround would be to set the `privacy_policy_url` to a page on a correctly configured server that contains a link to kentucky.gov. It isn’t as if any clients would do anything automatic with this URL.

grischard:
Our friends on the JOSM side have the same issue. I'd much rather see something that uses the system certificates, like the browsers do and like https://stackoverflow.com/questions/42982143/python-requests-how-to-use-system-ca-certificates-debian-ubuntu describes.

rbuffat:
Closing this as it seems not to be the solution.

PR #826, opened 2020-04-26
Update URLs in Norway

NKAmapper:
Update URLs/layers of Navigational aids and Administrative boundaries.
Also add privacy policy URLs.


Klumbumbus:
One difference to the JOSM entry. Which is the better choice?

* Attribution URL differs ('https://www.kartverket.no/kunnskap/Norges-grenser/' != 'https://kartkatalog.geonorge.no/metadata/administrative-enheter-wms/666e4559-60bf-4a1d-9e72-c43502a9a58b'): [NO] Kartverket Administrative Boundaries overlay [kartverket-abas]

PR #827, opened 2020-04-28
add new imagery available in Belgium (SPW OrthoPhoto 2018)

nobohan:
Add the orthophoto plan for 2018 WMS available in Belgium. The new geojson file is based on the same service as the SPW OrthoPhoto 2015 and SPW OrthoPhoto 2016.

Thanks in advance,

nobohan


nobohan:
just added the privacy_policy_url to complete the source

PR #833, opened 2020-05-05
Adding Cadastral Surveying layer from Bern, Switzerland

habi:
From https://opendata.swiss/en/dataset/amtliche-vermessung-farbig-wmts, with help from http://lists.openstreetmap.ch/mailman/listinfo/talk-ch


habi:
Mersibien!

Klumbumbus:
The Terms of use of "BY" at https://opendata.swiss/en/terms-of-use/ require: "You must provide the source (author, title and link to the dataset)." This is not compatible with OSM. It would require an explizit permission from the provider or an confimation that mention on https://wiki.openstreetmap.org/wiki/Contributors is sufficient.

simonpoole:
The other problem with this is that I believe only JOSM supports WMTS endpoints, so an actual TMS url would be far better in EPSG:3857 (this should be possible to deduce from the URLs generated by QGIS or JOSM). Not to mention that using a WMTS endpoint for exactly one layer doesn't make sense for JOSM either.

PS: it seems as if the WMTS server doesn't actually support either EPSF:3857 or EPSG:4326 so my complaint is mostly moot.

habi:
As can be seen in the first commit of this PR (https://github.com/osmlab/editor-layer-index/commit/9e8fa8cfd631a412292efc615189584ba7d4516f) we have the explicit permission to use this layer for updating OSM from the head of the office of  'Geoinformation Bern’ via email.

Klumbumbus:
OK, great.

habi:
> The other problem with this is that I believe only JOSM supports WMTS endpoints, so an actual TMS url would be far better in EPSG:3857 (this should be possible to deduce from the URLs generated by QGIS or JOSM). Not to mention that using a WMTS endpoint for exactly one layer doesn't make sense for JOSM either.
> 
> PS: it seems as if the WMTS server doesn't actually support either EPSF:3857 or EPSG:4326 so my complaint is mostly moot.

This actually stems from the earliest communication regarding this layer. It was suggested (by you :) ) that I submit this here.
I've been using the layer in JOSM and wasn't really aware that this repository here ist not per default used by JOSM...

simonpoole:
> This actually stems from the earliest communication regarding this layer. It was suggested (by you :) ) that I submit this here.
> I've been using the layer in JOSM and wasn't really aware that this repository here ist not per default used by JOSM...

That's not the problem, the way the layer is provided by Berne is technically incompatible with most editors. The solution to that would be proxying (that is reprojecting and caching) it through the SOSM mapproxy instance.

habi:
> The solution to that would be proxying (that is reprojecting and caching) it through the SOSM mapproxy instance.

Would that be something that is possible, or is this too convoluted a process?

simonpoole:
Sure, that is what it is there for.

habi:
> Sure, that is what it is there for.

Who do I have to nudge to make this happen?

Klumbumbus:
> I've been using the layer in JOSM

How did you use it? The highest zoom I can get is this, which is kinda useless: https://map.bern.ch/arcgis/rest/services/AV_farbig/MapServer/WMTS/tile/1.0.0/AV_farbig/default/default028mm/4/151/179.png

habi:
> > I've been using the layer in JOSM
> 
> How did you use it? The highest zoom I can get is this, which is kinda useless: https://map.bern.ch/arcgis/rest/services/AV_farbig/MapServer/WMTS/tile/1.0.0/AV_farbig/default/default028mm/4/151/179.png

I've added `https://map.bern.ch/arcgis/rest/services/AV_farbig/MapServer/WMTS?` as WMTS layer in JOSM and it was much better some days/weeks ago.
To double-check I removed the layer and added it again, now it's in the low resolution (exactly how you mention).
I hope this is just a temporary hiccup because like this it is *very* useless.

habi:
Seems to have been a temporary problem, now I see the 'cadastral survey' down to this: https://map.bern.ch/arcgis/rest/services/AV_farbig/MapServer/WMTS/tile/1.0.0/AV_farbig/default/default028mm/9/7559/8959.png (the [Wöschhüsi](https://www.openstreetmap.org/way/38533413) in the Matte).
JOSM takes some seconds to load the highest resolution, though...

PR #839, opened 2020-06-10
7.5 cm orthophotos from the city of St. Gallen

simonpoole:


rbuffat:
@simonpoole a merge of upstream/gh-pages hopefully solves the issue that `$(git diff --name-only $TRAVIS_COMMIT_RANGE)` returns a lot of unrelated files that are then subject to a check for the privacy policy.

simonpoole:
Thanks!

PR #842, opened 2020-06-16
adding FOMI 2007-2010 ortophotos

7F11:


7F11:
I'm newbie on Github. Any hints are welcome if I missed something.

7F11:
> ERROR:root:Error in sources/europe/hu/Hungary-FOMI-2007-2010.geojson : sources/europe/hu/Hungary-FOMI-2007-2010.geojson has no privacy_policy_url. Adding privacy policies to sources is important to comply with legal requirements in certain countries.

It may be _important_ but it is not _mandatory_ according to [schema](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json).

andrewharvey:
> It may be important but it is not mandatory according to schema.

I agree, if the source doesn't supply a privacy policy then it's impossible for us to add one.

The permission given "Director of Hungarian Institute of Geodesy, Cartography and Remote Sensing (FĂMI) licensed to use ortophoto 2007-2010 to be loaded into OpenStreetMap editor software and used for editing OSM data free of charge." I'll trust that was obtained from the data source and acts as a sufficient explicit permission.

andrewharvey (approved):

PR #847, opened 2020-07-06
Add wms layers for Kanton Schaffhausen

rbuffat:
This PR adds WMS layers for the Canton Schaffhausen:
- Orthophoto 2010
- Orthophoto 2013
- Orthophoto 2016
- Digital surface model relief 2013

For licence, I could only find the following information:
>Der Kanton Schaffhausen verfolgt seit 2012 im Bereich der Geoinformation die Open Government Data (OGD)-Strategie. Die  Geodaten stehen frei und kostenlos zur Verfügung. 
https://sh.ch/CMS/Webseite/Kanton-Schaffhausen/Beh-rde/Verwaltung/Volkswirtschaftsdepartement/Amt-f-r-Geoinformation-252309-DE.html

### DIFF
--- a/sources/europe/ch/Kanton_Schaffhausen_dsm_2013.geojson
+++ b/sources/europe/ch/Kanton_Schaffhausen_dsm_2013.geojson
@@ -1,0 +1,152 @@
+{
+    "type": "Feature",
+    "properties": {
+        "name": "Kanton Schaffhausen, Relief 2013",
+        "i18n": false,
+        "type": "wms",
+        "url": "https://wms.geo.sh.ch/raster?LAYERS=Relief_2013&STYLES=default&FORMAT=image/jpeg&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap",
+        "min_zoom": 8,
+        "license_url": "https://sh.ch/CMS/Webseite/Kanton-Schaffhausen/Beh-rde/Verwaltung/Volkswirtschaftsdepartement/Amt-f-r-Geoinformation-252309-DE.html",
+        "privacy_policy_url": "https://sh.ch/CMS/Webseite/Kanton-Schaffhausen/Beh-rde/Regierung/Staatskanzlei/Impressum/Datenschutz-1711122-DE.html",
+        "id": "Kanton-Schaffhausen-DSM-2013",
+        "country_code": "CH",
+        "default": false,
+        "best": false,
+        "start_date": "2013",
+        "end_date": "2013",
+        "category": "elevation",
+        "available_projections": [
+            "EPSG:4326",
+            "EPSG:2056",
+            "EPSG:21781"
+        ],
+        "attribution": {
+            "text": "Kanton Schaffhausen, Relief 2013",
+            "required": true
+        }
+    },
+
+    "geometry": {
+    "type": "Polygon",
+    "coordinates": [
+        [
+            [
+                8.6,
+                47.56
+            ],
+            [
+                8.6,
+                47.52
+            ],
+            [
+                8.52,
+                47.52
+            ],
+            [
+                8.52,
+                47.6
+            ],
+            [
+                8.44,
+                47.6
+            ],
+            [
+                8.44,
+                47.64
+            ],
+            [
+                8.36,
+                47.64
+            ],
+            [
+                8.36,
+                47.72
+            ],
+            [
+                8.4,
+                47.72
+            ],
+            [
+                8.4,
+                47.76
+            ],
+            [
+                8.44,
+                47.76
+            ],
+            [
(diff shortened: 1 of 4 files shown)


andrewharvey:
Where can I find the license or permission which allows this to be used in OSM?

rbuffat:
> For licence, I could only find the following information:
> 
> > Der Kanton Schaffhausen verfolgt seit 2012 im Bereich der Geoinformation die Open Government Data (OGD)-Strategie. Die  Geodaten stehen frei und kostenlos zur Verfügung.
> > https://sh.ch/CMS/Webseite/Kanton-Schaffhausen/Beh-rde/Verwaltung/Volkswirtschaftsdepartement/Amt-f-r-Geoinformation-252309-DE.html

This site states that the geodata is "open" and can be downloaded without cost. Furthermore, they state that the follow an open government data strategy for their geoinformation.  However, they do not mention any license.  Even a web search using `site:sh.ch lizenz` does not yield any useful results. I downloaded the DOM dataset from their geoportal. They do not include any license information with the dataset. Thus, I assume they do not apply any license to their data. If they would apply a license to their data, they would make it really, really hard to find.

There is also Art. 7.1 in the "Kantonales Geoinformationsgesetz (KGeoIG)":
```
1 Geobasisdaten des kantonalen Rechts sind öffentlich zugänglich
und können von jeder Person genutzt werden, sofern keine überwiegenden öffentlichen oder privaten Interessen entgegenstehen. 
```

Just a side note: the canton of Schaffhausen is a state in Switzerland: https://en.wikipedia.org/wiki/Canton_of_Schaffhausen

rbuffat:
I wrote an email to gis@ktsh.ch to ask for clarification.

andrewharvey:
I'm not a legal expert so will defer to others here, in many places data is automatically covered by copyright as the default and so needs a license compatible with OSM to use.

rbuffat:
I already got a reply: 


> Guten Tag Herr Buffat
> 
> Besten Dank für die Anfrage. Die Daten unterstehen keiner Lizenz, das ist richtig. Bei einer früheren Anfrage seitens OSM wurde der Hinweis auf https://wiki.openstreetmap.org/wiki/Contributors#Switzerland gemacht, dass Daten des Amt für Geoinformation genutzt werden. Evtl. könnte man den Hinweis um das Luftbild und Höhenmodell ergänzen. Ansonsten unterstehen die Daten aber keiner Einschränkung und können sehr gerne für die Verbesserung der OSM-Daten in Schaffhausen genutzt werden.
> 
> Ich hoffe das reicht Ihnen so, ansonsten können Sie sich gerne wieder bei mir melden.
> 
> Freundliche Grüsse


Deepl translation:

> Good afternoon Mr. Buffat
> 
> Many thanks for the inquiry. The data is not subject to a license, that is correct. In a previous request from OSM, https://wiki.openstreetmap.org/wiki/Contributors#Switzerland was mentioned that data from the Office of Geoinformation is used. Possibly the aerial photograph and elevation model could be added to the reference. Otherwise the data are not subject to any restrictions and can be used to improve the OSM data in Schaffhausen.
> 
> I hope this is enough for you, otherwise you can contact me again.
> 
> Friendly greetings
> 
> Translated with www.DeepL.com/Translator (free version)

andrewharvey (approved):

andrewharvey:
@rbuffat thanks for following up

Klumbumbus:
The shapes are way off. I improved them here: https://josm.openstreetmap.de/wiki/Maps/Switzerland?action=diff&version=193

rbuffat:
@Klumbumbus Can you elaborate why do you think the shapes are "way off"? If I'm not mistaken ELI only supports Polygon and not Multipolygons: https://github.com/osmlab/editor-layer-index/blob/gh-pages/scripts/check.py#L155

sh

Feel free to create a PR with updated shapes.

Klumbumbus:
These are not proper shapes for these sources imho:

grafik
grafik

You can take mine in geojson format from here if you want: https://josm.openstreetmap.de/maps?format=geojson&pretty=true

rbuffat:
I'm afraid this link seems not to produce a valid geometry:

See https://tools.ietf.org/html/rfc7946#section-3.1.6
```
   o  For Polygons with more than one of these rings, the first MUST be
      the exterior ring, and any others MUST be interior rings.  The
      exterior ring bounds the surface, and the interior rings (if
      present) bound holes within the surface.
```
Thus, the geometry should be a MultiPolygon and not Polygon, which the ELI checker does not allow. I don't know why, maybe because some editors are not capable of parsing Multipolygons? 

The shapes are based on the geometry of the administrative boundary, in a way that it is guaranteed that the area of the administrative boundary is included and file size is small. We are talking here of maximal ~5km overlap. In my opinion, they are adequate and I would not consider them "way off". But as I said before, feel free to create a PR, if you think you have better geometries.

PR #857, opened 2020-07-21
Add check for valid geometries, allow MultiPolygons

rbuffat:
Currently, ELI does not fully check the validity of the provided geometries (E.g. for overlapping exterior or interior rings). Having valid geometries would allow reducing the risk for complications when the sources are processed by tools assuming that the geometries are valid.

This PR adds a check for validity using ~~Fiona and~~ shapely: https://shapely.readthedocs.io/en/latest/manual.html#object.is_valid
Currently, only a warning is printed if the check does not succeed.  If this is something that should be included in ELI, this should be converted to raise an exception once all sources pass.


rbuffat:
It looks as multiple sources misuse interior rings of Polygons to emulate MultiPolygons. Thus, fixing these sources makes only sense if MultiPolygons are supported.

rbuffat:
scripts/convert_poly_to_multi.py is the script I implemented to make the sources valid. It uses Shapely's make_valid function, which is not yet included in a released version yet. 

101 sources were invalid and could be made valid by converting them to MultiPolygons. I only looked at a few of them to verify, and the ones I looked at seemed fine. With that, all geometries are valid according to shapely. 

This PR also modifies check.py to allow for MultiPolygon. I assume this not all that is needed to support them?

rbuffat:
convert_xml.py fails. Is there somewhere a specification that can be found for how multipolygons should be encoded in this format?

rbuffat:
After reading https://josm.openstreetmap.de/wiki/Maps I suspect that JOSM does not support polygons with holes. I thus adapted the convert to XML script to ignore all interior rings of polygons (holes) while exporting to XML. I'm not using JOSM so it would be good if somebody that does checks if still everything works. 

Also, as this PR possible breaks stuff, it would be good if all consumers of the editor-layer-index are aware of the changes and test them before this is merged.

rbuffat:
The following scripts now support MultiPolygons:
* scripts/check.py
* scripts/convert_geojson_to_legacyjson.py
* scripts/convert_individual.py 
* scripts/convert_xml.py

I did not look at geojsonify.py, as this file seemed not to be in use anymore. Thus what remains is to test if everything works. The file format should not have changed for consumers of imagery.xml.  However, imagery.geojson contains now MultiPolygon and Polygon geometries. I don't know everybody that uses imagery.geojson, but the consumers should test that they can handle MultiPolygon before this PR is merged. Alternatively, we could generate an imagery_mp.geojson with the adapted script, while still generating the old geojson with the invalid Polygons.

andrewharvey:
Thanks for the PR. All the Python code is outside what I'm comfortable reviewing, so hopefully someone else who's more familiar with the Python bits of this project will be able to review.

rbuffat:
Summary: Currently ELI requires all geometries to be Polygons. However, quite a few of the sources encode MultiPolygons incorrectly as a Polygon misusing interior rings (which would represent holes in the OpenGIS Simple Features Specification). This is problematic, as this leads to invalid geometries and prevents the further processing of these geometries with standard tools.

@Klumbumbus @simonpoole The XML format seems not to support holes. But these interior rings can simply be discarded when exporting. Is there some easy way to check if the created imagery.xml is valid and is understood by JOSM/Vespucci? E.g. a schema?

@quincylvania @bhousel would iD break if the imagery.geojson would contain Polygons and MultiPolygons?

Klumbumbus:
> Is there some easy way to check if the created imagery.xml is valid and is understood by JOSM

You could load the xml in JOSM and watch the Java console.

rbuffat:
> > Is there some easy way to check if the created imagery.xml is valid and is understood by JOSM
> 
> You could load the xml in JOSM and watch the Java console.

Sorry, I'm not that familiar with JOSM. Is there also a way to automate this so that it could be integrated in a test that automatically tests the produced imagery.xml each time it is built?

Klumbumbus:
Probably this one: https://josm.openstreetmap.de/browser/josm/trunk/resources/data/maps.xsd

simonpoole:
> @Klumbumbus @simonpoole The XML format seems not to support holes. But these interior rings can simply be discarded when exporting. Is there some easy way to check if the created imagery.xml is valid and is understood by JOSM/Vespucci? E.g. a schema?

Vespucci has been using the geojson format for many years in the meant time, however currently we only support polygon geometries, but adding MP support is trivial, but obviously not backwards compatible.

rbuffat:
@Klumbumbus @simonpoole Thanks for your feedback. 

It looks as the XML currently produced is not valid (there is no namespace, the description tag has no lang attribute and the country codes ZZ and XN are not included in the enumeration). 

It should be feasible to generate (geo)jsons for Polygon only and mixed Polygon/MultiPolygons.

Klumbumbus:
> the country codes ZZ and XN are not included in the enumeration

These are not official ISO 3166-1 alpha-2 codes

rbuffat:
> > the country codes ZZ and XN are not included in the enumeration
> 
> These are not official ISO 3166-1 alpha-2 codes

@Klumbumbus 
What should be done for sources that can not be matched to these codes? Just omit the country-code element? E.g. https://github.com/osmlab/editor-layer-index/blob/14ebe20c1cec505c0e57158222690e91c9483c89/sources/europe/Nordicsnowmobileoverlay.geojson
What should be written to the `lang` attribute of `description` if the language is not known?

Klumbumbus:
>What should be done for sources that can not be matched to these codes? Just omit the country-code element? 

I would say so.

>What should be written to the lang attribute of description if the language is not known?

One could assume english, but there are some descriptions in ELI which are not english.

* Description differs  ('Routes du réseau classé (autoroutes, nationales, départementales)' !=  'Routes of the classified network (highways, nationals,  departmentals)'): [FR] Route 500 [route500]
* Description differs ('Digitale Repräsentation des Geländes' != 'Digital representation of the terrain'): [CH] Canton Thurgau relief DTM [kt_tg_dtm_hillshade]
* Description differs ('Basisplan-AV des Kantons Thurgau farbig & schwarzweiss mit Liegenschaften, Nomenklatur, Landes-, Kantons- und Gemeindegrenzen, Einzelobjekten, Bodenbedeckung, Strassennamen, Höhenkurven und Geländemodell' != 'Base plan cadastral surveying of the canton Thurgau coloured & black & white with properties, nomenclature, national, cantonal and municipal boundaries, individual objects, land cover, road names, contour lines and terrain model'): [CH] Canton Thurgau Base plan cadastral surveying [kt_tg_av]
* Description differs ('Langsamverkehr: Verlauf der Wanderwege mit symbolischer Darstellung der Wegweiser, Unterführungen, Treppen und Brücken. Informationen über die Wegoberfläche und Rollstuhlgängigkeit.' != 'Slow traffic: the course of the hiking trails with symbolic representations of signposts, underpasses, stairs and bridges. Information about the path surface and wheelchair accessibility.'): [CH] Canton Thurgau hiking paths [kt_tg_ww]
* Description differs ('Langsamverkehr: Radweg-Netz (Routen) signalisiert durch Wegweiser' != 'Slow traffic: cycle path network (routes) indicated by signposts'): [CH] Canton Thurgau cycle routes [kt_tg_radrouten]
* Description differs ('Ausdehnung: ganzer Kanton Aufloesung: 10cm per Pixel Bildflug: 1.4.2011' != 'Expansion: whole canton Resolution: 10cm per pixel Photo flight: 2011-04-01'): [CH] Kanton Zug 2011 [Zug-2011]
* Description differs ('Digitale Orthophotos für das gesamte Berliner Stadtgebiet mit einer Bodenauflösung von 0,20 m im Blattschnitt 2 km x 2 km.' != 'Digital orthophotos for the entire Berlin city area with a ground resolution of 0.20 m in a sheet section of 2 km x 2 km.'): [DE] Berlin aerial photograph 2011 [Berlin-2011]
* Description differs ('Orthofoto fuer den Nordteil des Kantons Zug. Geflogen am 20.04.2016. 10cm Aufloesung per Pixel. Abdeckung: Gemeinden Cham, Steinhausen und Baar, sowie Teile von Zug Befliegung vor Laubaustrieb, ohne Schneeabdeckung.' != 'Orthophoto for the northern part of the canton Zug. Flown on 20.04.2016. 10cm resolution per pixel. Coverage: Communities of Cham, Steinhausen and Baar, as well as parts of Zug.'): [CH] Kanton Zug Nord 2016 [Zug-2016]
* Description differs ('Orthofoto fuer den Suedteil der Gemeinde Zug. Geflogen per Drohne an mehreren Tagen Anfangs April 2018. 10cm Aufloesung per Pixel. Abdeckung: Suedteil der Gemeinde Zug Befliegung vor Laubaustrieb, ohne Schneeabdeckung.' != 'Orthophoto for the southern part of the municipality of Zug. Flown by drone on several days in early April 2018. 10cm resolution per pixel. Coverage: Southern part of the community of Zug. Flying before leaves grow, without snow cover.'): [CH] Kanton Zug Gemeinde Zug Süd 2018 [Zug-2018]
* Description differs ('Réseau hydrographique: cours et plans d'eau et POI' != 'Hydrographic network: courses and water features and POIs'): [FR] BD Carthage [route500hydro]
* Description differs ('Darstellung der DGM Schummerung in Graustufen. Die imaginäre Lichtquelle befindet sich im Nordwesten. Eine Erhebung erscheint am Nordwesthang hell und am Südosthang dunkel. Ebenen sind mit mittlerer Helligkeit gefärbt.' != 'Display of the DTM shading in grayscale. The imaginary light source is located in the northwest. An elevation appears bright on the northwest slope and dark on the southeast slope. Layers are colored with medium brightness.'): [DE] North Rhine-Westphalia DTM hillshade [nrw_dtm_wms]

(maybe more)

rbuffat:
@simonpoole We should not change the format of an imagery* file if it breaks something. I changed the script that produces the imagery.geojson to produce Polygons with the same behavior as before. Only the *.geojson files in the sources folder are Polygon / MultiPolygons. 

@Klumbumbus  In general there seems to be no enforcement/requirement of a specific language for the description element in ELI. Would something like lang="unknown" work?

What do you think would be best for sources with a Polygon with more coordinates than specified in maps.xsd schema? Should they be dropped or should just the bounding box be used as geometry? Another option would be to implement an automatic simplification process.

Klumbumbus:
>Would something like lang="unknown" work?

I don't think thats needed. It works in JOSM without the language attribute.

>What do you think would be best for sources with a Polygon with more coordinates than specified in maps.xsd schema?

I would simplify the polygon. More than 999 points per shape are not needed in my opinion.

andrewharvey:
> I would simplify the polygon. More than 999 points per shape are not needed in my opinion.

This could also be done as a post-processing simplification. The source could store the full resolution data but then the outputs which editors consume is simplified.

rbuffat:
> > Would something like lang="unknown" work?
> 
> I don't think thats needed. It works in JOSM without the language attribute.

The `lang` attribute is marked as required in the maps.xsd. Thus, validation against the xsd fails without it. 

> This could also be done as a post-processing simplification. The source could store the full resolution data but then the outputs which editors consume is simplified.

Simplification looks like a good idea to me.

simonpoole:
I've finally got around to looking at this from a Vespucci pov. The current geometry processing code will fail gracefully if it encounters an MP and 15.2 will support processing geojson MPs.

With other words from a Vespucci pov this can be changed when convenient, though a format version bump in the output file would be helpful to allow continued processing of fake multi polygons as in the current format.

simonpoole:
@grischard seems as if the JOSM configuration doesn't contain the schema version which is a bit of a bummer.

simonpoole:
@rbuffat I just realized that this PR was never merged. Vespucci, in case it wasn't clear, has supported proper MPs in the configuration for over a year.

rbuffat:
I created a follow-up PR to this one: https://github.com/osmlab/editor-layer-index/pull/1450. I plan to merge this branch later this week.

**What will change?**

- imagery.geojson: After merging this PR, sources can have geometry type Polygon or MultiPolygon.
- imagery.json / imagery.xml: There should be no change in the format of the files. However, if in the future a source contains a Polygon with holes, the holes will not be exported (=only the exterior rings are used, as the geometry encoding for these formats does not support holes).

rbuffat:
Implemented with https://github.com/osmlab/editor-layer-index/pull/1450

PR #817, opened 2020-03-23
Maaamet layer addition (Estonia)

kallejre:
Added ~~3~~ ~~5~~ ~~7~~ 15 new Estonian ~~Maa-amet~~ map layers. 

This PR adds 3 new layers from Estonian land board, using existing corine land layer as template. Added layers are DSM (Digital Surface Model, essentially lidar scan), Hybrid and alternative Cadastre map (compared to existing one this has true colours, addresses data layer and hopefully better aerial image quality).

Later replaced broken hillshading layer with working wgs84-compatible layer and added modern Etak basemap (kaart24). More layers: fixed privacy policy and zoom levels for all layers, added nautical chart and Tallinn city planning department's map (EPSG:3301, but later has number of building levels). 

Also included 5 historic ortophotos from 1993 to 2019. Actually each of these layers include 3-4 sublayers from different years to get full coverage. Added 2 historic maps from 1899 & 1935 and one set of 6 sublayers from 1974-1989 with covering whole country at different zoom levels.

PR #858, opened 2020-07-21
Maa-amet layer addition (Estonia), attempt 2

kallejre:
This pull request is virtually duplicate of request #817 which became unmergeable and @andrewharvey suggested to open new one instead of rebasing old PR.

Insights to this PR: fixes, improves and optimizes 6 previously existing Estonian map layers (hence 18 000 deleted lines) and adds 17 new layers, featuring 8 historic maps/orthophotos. For more information about contents, visit PR #817. Currently there are issues with WMS server providing historical maps for years 1899 and 1935 and therefore two layers may appear pale white.


andrewharvey:
> The licensee undertakes to remove the reference to the origin of the data in its products or services in writing by the licensor
on request.

Is this going to be an issue?

Apart from that the license seems to be quite open, just the above clause and attribution.

kallejre:
No, i don't think it's going to be problem, because license with similar clause (from Maa-amet) has been used for OSM mapping for almost 9 years (+2 years with less refined license). However, there are few possible issues with merging this request. 
A) I edited check.py to add support for {wkid} in addition to {proj}. Was it allowed?
B) ~I started this pull request about half year ago. About that time Maa-amet data underwent reorganization due to merging different map applications. Before merging, all sources should be double-checked if they still work.~ I checked all layers and they produce expected result. However cadastral2 and TallinnLV maps might not hold up to quality expected by end users?
C) Also Travis complained about missing `style=` in wms URL. Should it be added before merging?

Asteliks:
Hi this PR seems to be waiting for quite a while now dead in the water. What do we want to do with it?

iandees:
Looks like this has a merge conflict. I'm going to close it out for now. If it's still useful, please resolve merge conflicts, push up to your branch, and leave a comment.

PR #864, opened 2020-07-28
Remove sources from kanton glarus

rbuffat:
As discussed in https://github.com/osmlab/editor-layer-index/pull/863


rbuffat:
@grischard I would welcome if this PR could be merged ASAP.

PR #867, opened 2020-08-05
Update Bordeaux 2016 hosting

gitouche-sur-osm:
This hosting needs to end. File are now served from another (slower) server.


gitouche-sur-osm:
no need for that anymore as wms.openstreetmap.Fr is hosting it

PR #870, opened 2020-08-06
Create Bayonne-2019.geojson

gitouche-sur-osm:
hosted by openstreetmap France since august 5, 2020.

### DIFF
--- a/sources/europe/fr/Bayonne-2019.geojson
+++ b/sources/europe/fr/Bayonne-2019.geojson
@@ -4,7 +4,7 @@
         "id": "Bayonne_2019", 
         "name": "Bayonne - Petit et Grand Bayonne 2019", 
         "type": "wms", 
-        "url": "https://wms.openstreetmap.fr/wms?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}", 
+        "url": "https://wms.openstreetmap.fr/wms?FORMAT=image/jpeg&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}", 
         "end_date": "2019",
         "start_date": "2019",
         "country_code": "FR",
--- a/sources/europe/fr/Bayonne-2019.geojson
+++ b/sources/europe/fr/Bayonne-2019.geojson
@@ -9,7 +9,8 @@
         "start_date": "2019",
         "country_code": "FR",
         "available_projections": [
-            "EPSG:3857"
+            "EPSG:3857",
+            "EPSG:4326"
         ],
         "attribution": {
             "text": "Ville de Bayonne - 2019", 
@@ -17,11 +18,12 @@
         },
         "max_zoom": 20,
         "license_url": "https://www.etalab.gouv.fr/licence-ouverte-open-licence",
-        "privacy_policy_url": "https://www.data.gouv.fr/fr/datasets/orthophotographie-petit-et-grand-bayonne-2019/"
+        "privacy_policy_url": "https://www.openstreetmap.fr/mentions-legales/"
     }, 
     "geometry": {
         "type": "Polygon", 
         "coordinates": [
+            [
                 [
                     -1.48262,
                     43.49176
@@ -123,5 +125,6 @@
                     43.49176
                 ]
             ]
+        ]
     }
 }
--- a/sources/europe/fr/Bayonne-2019.geojson
+++ b/sources/europe/fr/Bayonne-2019.geojson
@@ -1,0 +1,127 @@
+{
+    "type": "Feature", 
+    "properties": {
+        "id": "Bayonne_2019", 
+        "name": "Bayonne - Petit et Grand Bayonne 2019", 
+        "type": "wms", 
+        "url": "https://wms.openstreetmap.fr/wms?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}", 
+        "end_date": "2019",
+        "start_date": "2019",
+        "country_code": "FR",
+        "available_projections": [
+            "EPSG:3857"
+        ],
+        "attribution": {
+            "text": "Ville de Bayonne - 2019", 
+            "required": true
+        },
+        "max_zoom": 20,
+        "license_url": "https://www.etalab.gouv.fr/licence-ouverte-open-licence",
+        "privacy_policy_url": "https://www.data.gouv.fr/fr/datasets/orthophotographie-petit-et-grand-bayonne-2019/"
+    }, 
+    "geometry": {
+        "type": "Polygon", 
+        "coordinates": [
+                [
+                    -1.48262,
+                    43.49176
+                ],
+                [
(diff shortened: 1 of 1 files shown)


rbuffat:
Completely unrelated, but as you seem to be from France, there seems to be a bunch of broken imagery: https://rbuffat.github.io/eli_watchdog/ -> europe / fr.

rbuffat:
@gitouche-sur-osm Would you mind to change them FORMAT to image/jpeg (smaller file size to transfer images) and include "EPSG:4326" in available_projections?
For the privacy policy, it would probably make sense to use the policy of https://wms.openstreetmap.fr.

gitouche-sur-osm:
Hello,

I did the switch to jpeg, that works.

As for projection, EPSG:4326 doesn't seem to work : https://wms.openstreetmap.fr/wms?FORMAT=image/jpeg&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS=EPSG:4326&WIDTH=512&HEIGHT=512&BBOX=-164187.5908892,5386823.7610392,-164111.1538642,5386900.1980642

About the policy, I don't know. The current URL (https://www.data.gouv.fr/fr/datasets/orthophotographie-petit-et-grand-bayonne-2019/) is where the data were officialy uploaded as opendata, right from the source. wms.openstreetmap.fr is only hosting a copy of it for OSM editors.

rbuffat:
The EPSG:4326 projection uses degrees as unit for the coordinates. Thus, it is not possible to use the same coordinates as with EPSG:3857. 
The projection seems to work: https://wms.openstreetmap.fr/wms?FORMAT=image/jpeg&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS=EPSG:4326&WIDTH=256&HEIGHT=256&BBOX=-1.477661132812506,43.49079023700749,-1.4762878417968832,43.491786530833764 

There is a tool to facilitate the creation of the WMS parameters, in case you want to check it out: https://hub.gke.mybinder.org/user/rbuffat-eli-helper-7x5hfnuv/apps//create_wms_template_appmode.ipynb

The privacy policy is about what information a server saves about a user when it is accessed. E.g. the IP address, cookies, browser information etc. Thus, only the https://wms.openstreetmap.fr server sees this information, as the data is served by this server.

sources/europe/fr/Bayonne-2019.geojson
```diff
@@ -0,0 +1,127 @@
...
+            "text": "Ville de Bayonne - 2019", 
+            "required": true
+        },
+        "max_zoom": 20,
+        "license_url": "https://www.etalab.gouv.fr/licence-ouverte-open-licence",
+        "privacy_policy_url": "https://www.data.gouv.fr/fr/datasets/orthophotographie-petit-et-grand-bayonne-2019/"
+    }, 
+    "geometry": {
+        "type": "Polygon", 
+        "coordinates": [
```
rbuffat, line 25: The coordinates need to be wrapped in an additional [ ]

gitouche-sur-osm:
> The EPSG:4326 projection uses degrees as unit for the coordinates. Thus, it is not possible to use the same coordinates as with EPSG:3857.
> The projection seems to work: https://wms.openstreetmap.fr/wms?FORMAT=image/jpeg&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=bayonne_2019&STYLES=&SRS=EPSG:4326&WIDTH=256&HEIGHT=256&BBOX=-1.477661132812506,43.49079023700749,-1.4762878417968832,43.491786530833764

Ok, I see. So just adding the EPSG:4326 coordinates should work. Thank you.

> 
> The privacy policy is about what information a server saves about a user when it is accessed. E.g. the IP address, cookies, browser information etc. Thus, only the https://wms.openstreetmap.fr server sees this information, as the data is served by this server.

Best I could find is https://www.openstreetmap.fr/mentions-legales/
I don't see any privacy or GDPR statement, but that is the "legal terms" page for openstreetmap.fr.

andrewharvey:
I found an english version of the license at https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf which does require attribution, I know for CC BY we ask for a waiver to clarify that attribution on the wiki is okay (and every reuse only needs to attribute OSM). Would we also need that waiver here too?

cquest:
**Attribution**: it just needs to be in the changeset source, or on objects. We've been doing like that for years. No waiver needed.

**Privacy policy**: we dont keep track of anything on wms.openstreetmap.fr, event the client IP is not logged

**WMS or TMS**: all layers available on wms.openstreetmap.fr are also available in TMS format in EPSG:3857 (256px tiles web mercator), example : http://wms.openstreetmap.fr/tms/1.0.0/bayonne_2019/20/543839/683479.jpeg
Original imagery is in EPSG:2154, don't use 4326 as quality will degrade.

rbuffat:
@cquest Thanks for your quick reply. 

Regarding attribution, as far as I'm aware the "proper" way for attribution is the [Contributors page in the OSM wiki](https://wiki.openstreetmap.org/wiki/Contributors). For sources requiring attribution, it should be checked if the data owner agrees that this kind of attribution is ok with them. But I have limited knowledge. @grischard Maybe you can comment on this as you have probably the most experience.

andrewharvey:
> Attribution: it just needs to be in the changeset source, or on objects. We've been doing like that for years. No waiver needed.

I don't think that's generally something that has been acceptable, we can't guarantee attribution will remain on the changeset tag or object tag, the only thing we can do is place attribution on the wiki page. I wish I could find a reference for this, but it's just based on anecdotal chatter. So happy to re-confirm this with LWG and/or DWG.

Marc-marc-marc:
> we can't guarantee attribution will remain on the changeset tag

changeset tags are immutable

rbuffat:
Just listing some resources:

There are discussions in the iD and JOSM issue trackers about attribution in changesets:
https://github.com/openstreetmap/iD/issues/6300
https://josm.openstreetmap.de/ticket/19911

There is an article from @simonpoole about the compatibility of attribution of CC-BY licenses with OSM: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

andrewharvey:
>> we can't guarantee attribution will remain on the changeset tag
> changeset tags are immutable

true, but we can't guarantee that contributors will set the changeset tag, and history gets lots as ways are split and worked on by subsequent contributors. we also can't guarantee that downstream consumers of OSM data will provide attribution to our individual contributors, this is one of the things that the CC BY waiver clarifies. 

> There is an article from @simonpoole about the compatibility of attribution of CC-BY licenses with OSM: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

Indeed, to quote,

> The reasons for this decision are twofold: CC BY 4.0 allows some more flexibility in how a licensee can attribute the original source and many people have suggested that our indirect attribution mechanism already fulfills the requirements of BY 4.0. However the OSMF believes that licensors of BY 4.0 material may be surprised by this and that we should, for courtesy and good form reasons, continue to ask for explicit permission to attribute via our websites.

Now this is not CC BY, but still an attribution only like license isn't necessarily compatible to our high standards, the same way CC BY isn't.

grischard:
The overwhelming consensus in the French community is that no waiver is needed, and that the current way is fine. This includes data producers and the drafters of the licence.

rbuffat:
@grischard Thanks for the clarifications!

simonpoole:
Just a note on this topic, which has been discussed somewhere on the repo before: it would be completely possible to add an attribution tag, or more than one to the changesets if there was a flag that indicated this in ELI and there was agreement on where the attribution would go (not the source tag).

The changeset tags are immutable as has been pointed out and as such are not a bad place to put attribution, it is true that it might be painful to trace back to the initial changeset from elements have been modified, I would consider that more of an issue for a OSM contributor wanting to know from where data originated, not a legal issue (there is clearly no IP affected at the level of individual elements).

PR #874, opened 2020-08-07
updated freemap outdoor bounding polygon and minor details

zdila:
ref. https://github.com/osmlab/editor-layer-index/pull/739#issuecomment-669729472


andrewharvey (approved):

Klumbumbus:
The update of the attribution text and url should be done to the 4 other freemap.sk sources too. (car is without SRTM.)

andrewharvey:
If using the public domain SRTM data, it doesn't require attribution, so while it's still preferable to provide the attribution regardless it's not required.

Klumbumbus:
I think you misunderstood me. The old attribution is: "_Copyright ©2007-2012 Freemap Slovakia (www.freemap.sk). Some rights reserved._" which is outdated and doesn't even mention osm which is bad since the data is mostly osm data. The new attribution is much better: "_Map © Freemap Slovakia, data © OpenStreetMap contributors and SRTM_".
And for the variant car "_and SRTM_" should be omitted since it doesn't contain SRTM data.

andrewharvey:
Ah sorry @Klumbumbus yes I did misunderstand, I thought you meant car was missing the SRTM attribution and saying it needed it.

@zdila did you want to update the attribution for the other sources too?

zdila:
@andrewharvey [done](https://github.com/osmlab/editor-layer-index/pull/885).

Issue #537, opened 2018-08-15
Add Orthophotograph of Turku


theel0ja:
## URL of the imagery
https://www.avoindata.fi/data/en_GB/dataset/turun-kaupungin-paikkatietorajapinnat

## Two letter country code
FI

## Desired layer name
Orthophotograph of Turku

## License
CC-BY 4.0

## Is the license compatible with OSM?
Requires permission. TODO

## Attribution (text, URL, required?)
`Turun kaupunki` / `City of Turku`

## Icon
(optional)

## Bounding polygon

TODO

ilmaisin:
Since Feb 4 2019, OSM community has had a special waiver that allows the use of City of Turku data, see http://www.turku.fi/avoindata/jhs189 and  http://dev.turku.fi/openstreetmap-lupa.pdf.

ilmaisin:
Here's the bounding polygon.
[turku-ortophoto-polygon.geojson.txt](https://github.com/osmlab/editor-layer-index/files/4121925/turku-ortophoto-polygon.geojson.txt)

andrewharvey:
Does anyone want to have a go at a PR, I'm happy to help review it.

theel0ja:
It seems to have been added into JOSM: https://josm.openstreetmap.de/wiki/Maps/Finland

PR #875, opened 2020-08-07
add Turku 2018 true orto

crimsondusk:
Same source as https://github.com/osmlab/editor-layer-index/issues/537

In addition to the regular orthophoto, city of Turku also has a better quality ortophoto on the inner city area.

### DIFF
--- a/sources/europe/fi/turku-orto-2018-true.geojson
+++ b/sources/europe/fi/turku-orto-2018-true.geojson
@@ -3,7 +3,7 @@
     "properties": {
         "attribution": {
             "required": true,
-            "text": "\u00a9 Turun kaupunki",
+            "text": "© Turun kaupunki",
             "url": "https://www.turku.fi/turku-tieto/kartat-ja-paikkatieto"
         },
         "available_projections": [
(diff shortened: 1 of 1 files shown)


sources/europe/fi/turku-orto-2018-true.geojson
```diff
@@ -0,0 +1,54 @@
+{
+    "type": "Feature",
+    "properties": {
+        "attribution": {
+            "required": true,
+            "text": "\u00a9 Turun kaupunki",
```
andrewharvey: you should be able to just insert the unicode © character

sources/europe/fi/turku-orto-2018-true.geojson
```diff
@@ -0,0 +1,54 @@
...
+        "best": true,
+        "country_code": "FI",
+        "description": "Ortophotos from the city of Turku",
+        "icon": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/d7/Turku.vaakuna.svg/200px-Turku.vaakuna.svg.png",
+        "id": "turku-orto-2018-true",
+        "license_url": "https://dev.turku.fi/openstreetmap-lupa.pdf",
+        "privacy_policy_url": "https://www.turku.fi/turku-tieto/tietosuoja/kayttotiedot-ja-evasteet",
+        "max_zoom": 20,
+        "min_zoom": 4,
+        "name": "City of Turku ortophoto - 2018 True ortho",
```
andrewharvey, line 24: if this is from 2018, did you want to add the start_date and end_date properties as 2018? eg https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/oceania/au/act/ACTmapi-Imagery202001.geojson?short_path=9d34d6f#L18-L19

andrewharvey (approved):

sources/europe/fi/turku-orto-2018-true.geojson
crimsondusk: Found and added the date

sources/europe/fi/turku-orto-2018-true.geojson
crimsondusk: Fixed. I just copied this from turku-orto.geojson so that file has this escape sequence as well

Issue #872, opened 2020-08-06
WMS -> WMTS for Geoportal 2: Orthophotomap (aerial image)


Asteliks:
Requesting to move form WMS to WMTS for the main map for Poland "Geoportal 2: Orthophotomap (aerial image)"

## URL of the imagery
WMS: https://mapy.geoportal.gov.pl/wss/service/img/guest/ORTO/MapServer/WMSServer
WMTS: https://mapy.geoportal.gov.pl/wss/service/WMTS/guest/wmts/ORTO
source: https://www.geoportal.gov.pl/uslugi/usluga-przegladania-wms

## Two letter country code
PL

Issue #845, opened 2020-06-24
Geoportal2 (primary aerial in Poland) is broken in parts of the country


matkoniecz:
In iD it is showing as a blank white space. The same in JOSM and Vespucci.

JOSM lists also Geoportal WMTS that works fine.

https://github.com/osmlab/editor-layer-index/blob/d9ac23080bb038385279477f000c28a95e61be36/sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson

See also https://josm.openstreetmap.de/ticket/19656

EDIT: it actually is limited to Województwo Małopolskie....

https://www.openstreetmap.org/edit?editor=id#map=10/50.2714/20.9313

![screen06](https://user-images.githubusercontent.com/899988/89934461-e652e080-dc10-11ea-86c7-d0ae3554511e.png)

matkoniecz:
As temporary fix it may be useful to set Bing as the best aerial in Poland to prevent malfunctioning aerial in iD.

maro-21:
It's not broken, it works. I map almost every day and haven't seen any blank tiles.
They are black just before loading, but never seen white tiles.
In which region do you map?

matkoniecz:
Kraków

maro-21:
I've checked several major cities in Poland (Warszawa, Łódź, Wrocław, Poznań, Gdańsk, Bydgoszcz) and it works there. Looks like there is only some issue with Kraków.

Maybe you should set Bing as default imagery for Kraków for now.

matkoniecz:
Have you checked Kraków? Because I would expect local simultaneous independent issues on all hardware/networks that I tested (not so many) rather than Kraków lag in Geoportal WMS

matkoniecz:
Wait what. It is actually location-based with Kraków heavily impacted.

https://www.openstreetmap.org/edit?editor=id#map=17/50.63030/17.99701 https://www.openstreetmap.org/edit?editor=id#map=19/50.06186/19.94080

![screen04](https://user-images.githubusercontent.com/899988/89933680-b6570d80-dc0f-11ea-8574-093f4f87caec.png)
![screen05](https://user-images.githubusercontent.com/899988/89933691-b8b96780-dc0f-11ea-86ec-3f68316dcbb2.png)

matkoniecz:
https://www.openstreetmap.org/edit?editor=id#map=10/50.2714/20.9313

[...]

PR #877, opened 2020-08-09
remove best marker from broken layer see #845 #872

matkoniecz:
I am not sure what exactly went wrong but for long time this layer is showing blank white instead of aerial tiles

I am not making PR removing it as maybe it is fixable, but it is no longer the best imagery


matkoniecz:
Failure is 

> ERROR:root:Error in sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson : sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson has no privacy_policy_url. Adding privacy policies to sources is important to comply with legal requirements in certain countries.

and therefore can be ignored as it is minor and not introduced in this PR anyway

grischard:
What about just nuking it if it's broken?

matkoniecz:
It sometimes works, but extremely rarely. Maybe it will change?

In JOSM I simply removed it as there is working wmts - that is sadly unsupported by iD and Vespucci

I intended this is band aid to unbreak iD that is now showing white as an aerial

andrewharvey (approved):

matkoniecz:
I posted at https://forum.openstreetmap.org/viewtopic.php?pid=797364#p797364 and https://t.me/OpenStreetMapPolska/1916 to make Polish OSM Community informed about the situation

maro-21:
It's not broken, it works. I map almost every day and haven't seen any blank tiles. Maybe it's only in the region where you edit.
Could you undo this change?

andrewharvey:
@matkoniecz @maro-21 if you two could come to some agreement if this should be best or not, I think that would help resolve the outstanding issue, then it well need to be another PR to revert the change if that's decided upon.

PR #878, opened 2020-08-12
Improve checks for newly added sources

rbuffat:
This has some overlap with https://github.com/osmlab/editor-layer-index/pull/787, sorry I did notice it too late.  Basically, this PR integrates checks implemented for https://rbuffat.github.io/eli_watchdog/ in the strict tests.

This PR separates the strict checks from the "classic" checks in a new file scripts/strict_check.py. 
The following checks are implemented:

TMS:
- Test all zoom levels between min_zoom and max_zoom.

WMS
- Validate URL
- Check mandatory GetMap parameters against GetCapabilities metadata
- Check if available_projections are advertised by the WMS Server
- Check if jpeg FORMAT is used if available (for photo and historic photo layers)
- Check if WMS version corresponds to the newest WMS version server understands. (Assuming that every OSM Editor supports the changed axis ordering for EPSG:4326 with WMS 1.3.0, otherwise this warning should better be omitted)
- Log WMS Metadata Fees / AccessConstraints
- When the server supports EPSG:4326, EPSG:3857 or CRS:84, check if they are included in available_projections
- Differentiate between true WMS sources and ERSI REST API sources encoded as wms source

WMS Endpoint:
- Just check if the GetCapabilities request can be parsed
- Log WMS Metadata Fees / AccessConstraints

WMTS Endpoint
- Just check if the GetCapabilities request can be parsed

Other imagery types are currently not checked. 

I tested the script locally, e.g. with `python scripts/strict_check.py sources/europe/*/*`. However, I did not test it on Travis. Thus, there might be still some bugs left.


andrewharvey (approved):

rbuffat:
I updated the feature list in the post above. I suspect there will always be something that can be improved or added.

andrewharvey:
@rbuffat @grischard happy for this to be merged?

rbuffat:
From my side, it can be merged.

It is nearly the same code as used for https://rbuffat.github.io/eli_watchdog, thus it is tested against quite a lot of different sources. But probably it's best to look at this as a work in progress as the "in production" will show if tests are too strict or should be adapted. Thus failing tests should be looked at with some caution. If it turns out to be counterproductive, it can be reverted at any time.

rbuffat:
- Add a check that only 5% of the geometry should be outside of a WMS layer bounding box. This check only works for valid geometries: see https://github.com/osmlab/editor-layer-index/pull/857
5% is an arbitrary value and should be adapted as needed.
- Check URLs for source['properties']['attribution']['url'] and source['properties']['icon']

rbuffat:
@andrewharvey @grischard I merging the script. I added a warning that it is in beta stage. I expect that there will be some issues. Please @ me if you notice any issues. Also, I want to emphasize that none of the checks should be considered as "carved in stone" and should be changed as needed.

PR #880, opened 2020-08-13
Resolve warnings for europe/ch

rbuffat:
This PR resolves warnings for europe/ch.

- Readd source for Kanton Zug 2019 that was accidentally deleted by travis deploy: https://github.com/osmlab/editor-layer-index/pull/866
- Set min_zoom from 10 to 12 for Lausanne - Orthophoto 2016
- Replace invalid "EPSG:84" with CRS:84 for Kanton Solothurn, DSM Relief 2018
- Add privacy policy for "Stadt Bern 10cm (2016)", "Kanton Solothurn, DSM Relief 2019" and "Kanton Solothurn, DTM Relief 2019"
- Add STYLES to WMS URLs where it was missing. "Styles" is mandatory according to the WMS specification (Table 8, [1]). However, the sources worked before without.
- Switch WMS sources to the latest WMS version the server understands
- Switch to image/jpeg if WMS server supports the format for photos, png for maps.
- Add category when it was missing
- Mark orthophoto for Schaffhausen as best

[1] http://portal.opengeospatial.org/files/?artifact_id=14416


rbuffat:
The tests fail because of the missing imagico.de privacy policy. I currently did not found one yet.

andrewharvey:
> The tests fail because of the missing imagico.de privacy policy. I currently did not found one yet.

I'm not worried about that, if upstream doesn't have a privacy policy, then we should just include `privacy_policy: null`. In my opinion the tests which check for a missing `privacy_policy` key, but if `privacy_policy: null` then that should pass the test.

andrewharvey (approved):
Looks fine to me, just say when it's ready to merge.

andrewharvey:
Where you changed the format, did you look at the map and decide which format is best? I think end of the day it should be a human looking at the map and deciding, not just picking which based on the category.

While most orthophotos should be better as jpeg if supported, map could be either jpeg or png. For example if it's a scan of a map that was saved with lossy compression, jpg would work better, but if it's a computer rendered map that's always been lossy, png would be better. So needs human eyes to make that call.

rbuffat:
@andrewharvey that's a good point. They are maps rendered from geodata. Actually, it would be nice to have a tool to test the size of the image formats. 

I'm also a bit confused between the categories photo and historicphoto. Intuitively, historicphoto refers for me to old (~older than at least 20-30 years) aerial imagery. However, it seems also to be used by more recent imagery. Also, I'm not really sure why it is used anyway.  Due to the end_date attribute, it should be known for each region which the most recent imagery source is.

rbuffat:
@andrewharvey except for the photo vs historicphoto discussion the PR should be ready to be merged.

andrewharvey:
> I'm also a bit confused between the categories photo and historicphoto. Intuitively, historicphoto refers for me to old (~older than at least 20-30 years) aerial imagery. However, it seems also to be used by more recent imagery. Also, I'm not really sure why it is used anyway. Due to the end_date attribute, it should be known for each region which the most recent imagery source is.

I agree this is not clear in the schema, with two conflicting views:

1. Any imagery where we have newer imagery available is marked as historic
2. Imagery from decades ago

Would be great to standardise this in the spec.

ELI is an index of map layers helpful for OSM editing, it's not a general purpose index of background maps which you might want to enable in your app. Therefore if the layer is decades old it has no practical value for mapping so shouldn't be in this index.

rbuffat:
> ELI is an index of map layers helpful for OSM editing, it's not a general purpose index of background maps which you might want to enable in your app. Therefore if the layer is decades old it has no practical value for mapping so shouldn't be in this index.

In europe /gb, nearly all of the sources are maps that are 60-100 years old.  Should they be removed or is there a place for old sources?

Klumbumbus:
> if the layer is decades old it has no practical value for mapping
If there is no other source in that region it could be useful for e.g river names or mountain heights or...

PR #887, opened 2020-08-16
Resolve warnings and some broken sources for europe/fr

rbuffat:
- Add categories (from JOSM index)
- Replace Toulouse sources from JOSM
- Fix Lyon URLs
- Update WMS urls


rbuffat:
It's better to split this in multiple PRs.

PR #893, opened 2020-08-20
restore best marker on partially broken layer see #845 #872 #877

matkoniecz:
Revert "remove best marker from broken layer see #845 #872"
This reverts commit c3fdd9f19217d08347514b365e3f4b1ae1b4af16.

Yes, this layer is boken in a big part of country but elsewhere it
works fine. So something less brutal should be done.


andrewharvey:
Maybe the a good middle ground is to split this into two sources with different regions based on where it's usually working vs never usually working, then you can have a different best value per region. Is that viable or is to too hard to determine this extent?

If it's not viable we can merge this.

matkoniecz:
> Is that viable or is to too hard to determine this extent?

It is viable and I have it on my TODO list, but I have no idea when I will be able to do this.

It appears that imagery is unavailable in https://www.openstreetmap.org/relation/224459 (województwo małopolskie)

For now I contacted operators of the service and they claim that WMS works correctly in Małopolska so my next step is to investigate what exactly fails and how.

matkoniecz:
I submitted this as it is better to have blank tiles in Małpoloska then poorly aligned substandard Bing everywhere, but either fixing service or special handling for Małopolska would be even better.

But I prefer to not be responsible for breaking imagery and iD and I have no idea when they will update and pull ELI data.

PR #895, opened 2020-08-22
Add Orthophoto 2019 (5cm) for CH/Geneva

alkino:
I only replace 2016 by 2019, I don't know how this file is generated.
Check this file twice before merging.

### DIFF
--- a/sources/europe/ch/Canton_de_Geneve_orthophoto2016_5cm.geojson
+++ b/sources/europe/ch/Canton_de_Geneve_orthophoto2016_5cm.geojson
@@ -3,10 +3,10 @@
     "properties": {
         "license_url": "https://ge.ch/sitg/media/sitg/files/documents/conditions_generales_dutilisation_des_donnees_et_produits_du_sitg_en_libre_acces.pdf",
         "privacy_policy_url": "https://ge.ch/sitg/conditions-utilisation",
-        "start_date": "2019",
-        "end_date": "2019",
-        "id": "Geneve-orthophoto-2019",
-        "url": "https://ge.ch/sitgags2/services/RASTER/ORTHOPHOTOS_2019/MapServer/WMSServer?LAYERS=0&STYLES=default&FORMAT=image/jpeg&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap",
+        "start_date": "2016",
+        "end_date": "2016",
+        "id": "Geneve-orthophoto-2016",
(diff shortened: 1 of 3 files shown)


rbuffat:
I would not replace a source that is still working, but rather create a new one. Old imagery might still be useful, e.g. to see why a feature was created in the past.

To create parameters for a WMS layer you can use the following tool: https://mybinder.org/v2/gh/rbuffat/eli-helper/master?urlpath=apps%2F/create_wms_template_appmode.ipynb In this case, you would need to use `https://ge.ch/sitgags2/services/RASTER/ORTHOPHOTOS_2019/MapServer/WMSServer?` as URL. 

To check if the geometry corresponds to the wms layer, you can use e.g. [QGIS](qgis.org). You can drag and drop the geojson you created to QGIS and add the WMS layer with: Layer -> Add Layer -> Add WMS/WMTS Layer. You first have to add the WMS server using New (You can use the url from above). Then you can connect the server and select the WMS layer and Add it to the map. 

Orthophotos are typically processed to match the ground (You can find some documentation of this process in French [here in Chapter 3](https://www.swisstopo.admin.ch/content/swisstopo-internet/fr/home/products/images/ortho/swissimage10/_jcr_content/contentPar/tabs_copy/items/60_1569482292365/tabPar/downloadlist/downloadItems/67_1588752711524.download/Produktinfo_SWISSIMAGE10cm_FR_bf.pdf) ). Typically, a dataset for a specific region (e.g. a Canton), is only required/tested to be accurate for this region. However, the dataset can span a greater area. For this area, the quality might not be the same. Thus, to be safe, it might be better to just include the boundaries of the region.

alkino:
I have make a fix.
I've try the two steps:
* parameters for WMS
* QGIS

For QGIS I have not an idea if it's good or no, there is yellow square and layer...

rbuffat:
If you right-click on the geojson layer -> properties you can adjust the transparency of the layer. The geometry is a rough approximation of Geneva, but it should be fine for this source.

If you feel that this new source is the best for the region, you can change the "best" property to true and set the property to false for the source that previously had best = true.

rbuffat:
The source is marked with "Accessible for consultation and extraction for free use (Open Data)" here: https://ge.ch/sitg/sitg_catalog/data_details/f498b260-5381-4912-b2e7-906a91902e3b/xhtml_raw

alkino:
So is it good or no? I've search for a better geojson area for geneva canton (after using transparency in QGIS) but cannot find it in open data of SITG, maybe something is available openly somewhere else?

rbuffat:
Looks good from my side. Merci!

alkino:
What is the next step to have this merged?

andrewharvey:
> What is the next step to have this merged?

@grischard do you have permissions to allow @rbuffat to merge PRs? Doesn't looks like I can make that change.

PR #896, opened 2020-08-23
Fix south tyrol

rbuffat:
- Fix broken url for "South Tyrol Orthofoto 2011 (highres)"
- Add DSM / DTM Hillshade layers. They are available in a 0.5 and 2.5 m resolution. The 0.5 resolution seems only to cover the area around urban settlements. Thus I also included the 2.5m layer, which covers a wider area.
- Add privacy policies

Licenses:
- DSM/DTM and Ortho 2017 sources CC0-1.0
- Ortho 2011, 2014 and 2014-2015: CC-BY 4.0 (Already included in index)

I did not find a way to directly link to the license information. It is necessary to search in the geocatalog: https://geoservices.buergernetz.bz.it/geokatalog for the data and then search for the datasets and click on metadata.


andrewharvey:
I know you're just fixing up the existing source and not adding it. But since CC BY 4.0 needs a waiver for use in OSM https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/ I'm hesitant to make changes, and if anything we should drop it, unless there was/is a waiver completed that we can include here.

rbuffat:
Keeping @Klumbumbus in the loop, as the CC-BY 4.0 sources are also included in JOSM

@andrewharvey The CC0-1.0 sources would be ok, right? I could create a separate PR for them.

andrewharvey:
Yes CC0 is fine, see https://wiki.osmfoundation.org/wiki/Licence/Licence_Compatibility

CC BY 4.0 does require the LWG waiver or equivalent.

rbuffat:
@andrewharvey Just FYI, while the 2011 layer is broken, the 2014 and 2014-2015 layers still work. 

The 2014, 2014/15 and 2017 are WMTS sources. The Geokatalog seems only to list WMS servers, not WMTS. The site where the WMTS server is linked from (http://geoportal.buergernetz.bz.it/geodatendienste.asp) mentions:

> Die Standardlizenz mit der die Daten ausgegeben werden:
> Falls nicht anders angegeben, werden die Daten des Geoportal Südtirol mit der Lizenz "Creative Commons Zero" (kurz CC0) ausgegeben.

Deepl translation:
> The standard licence with which the data is issued:
> If not stated otherwise, the data of the Geoportal South Tyrol are output with the licence "Creative Commons Zero" (short CC0).

The WMTS server does not state any metadata to indicate that these layers have a different license. This is probably the reason why the sources were added in the first place.

tyrasd:
Sorry to jump in this late, but I think you have made a mistake by deleting some of the South Tyrol imagery in this PR.

> remove CC-BY sources without waiver

Did you even search or ask around if such a waiver existed, e.g. by contacting the original contributor of these layers (i.e. [me](https://github.com/osmlab/editor-layer-index/commits/11eb3f800714b7100d0faaa7079e36aa4bdacdd0/sources/europe/it/SouthTyrolOrthofoto2014-2015.geojson))?

In fact, there exists an explicit waiver in the metadata [returned by the geoserver](https://geoservices.buergernetz.bz.it/geoserver/ows?SERVICE=WMS&REQUEST=GetCapabilities) which was even linked in the `license_url` of the removed layer (and is even mentioned on the OSM wiki at https://wiki.openstreetmap.org/wiki/AltoAdige_-_S%C3%BCdtirol):

> TERMS OF CARTOGRAPHIC NETWORK SERVICES OF THE SOUTHTYROLEAN GEOPORTAL: The Service and its contents can be used for the following purposes: 1. Private use, 2. Research, 3. Culture 4. Use in schools / education, 5. **Production of Open Data derived products**, 6. Use for club purposes, 7. Use within emergency service organizations, 8. Use within authorities, but only to meet statutory duties. 9. Use for the creation of secondary products that can be used commercially: Unless otherwise indicated, the data of the Geoportal Südtirol be issued with the disclaimer "Creative Commons Zero" (short CC0). Information about applied license are given for each dataset in the Open Data Portal Südtirol (German: http://daten.buergernetz.bz.it, Italian: http://dati.retecivica.bz.it). The massive download of large amounts of data is not recommended because this may result in relatively long waiting times or in particularly critical cases also in the termination of the download process.ion about applied license are given for each dataset in the Open Data Portal Südtirol (German: http://daten.buergernetz.bz.it, Italian: http://dati.retecivica.bz.it). The massive download of large amounts of data is not recommended because this may result in relatively long waiting times or in particularly critical cases also in the termination of the download process.

I will revert  the PR now.

rbuffat:
@tyrasd You are totally right, I should have contacted you.

The issue with cc-by is if the way of attribution of OSM is OK for the data owners. See also https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/
I did not find any waiver for this, but it would be great if you have one!

The terms you linked seem not to include anything regarding attribution, or I overlooked it. When I looked in the geokatalog, the data was licensed with cc-by and not cc-0, where attribution would be no issue: https://geoservices.buergernetz.bz.it/geokatalog

tyrasd:
> The terms you linked seem not to include anything regarding attribution […] the data was licensed with cc-by

Here's how I understand the terms: The data can be used without further restrictions for the “_production of open data derived products_” (point 5 above) **or** under the terms of the CC-BY license (or CC-0 if explicitly specified) for the production of commercial secondary products (point 9 above). I came to this conclusion because if the CC-BY terms were meant to be applied to all listed points, there would be no need to list these special usages in the first place. Further, the permission to use these WMS/WMTS layers for OSM mapping existed for quite some time even before the province adopted the CC-0/CC-BY licenses. Also, I have looked through my email archives and found a discussion I had in 2013 with a responsible person of the GIS department, which explicitly confirmed my question if these terms allow the usage of the WMS/WMTS services for OSM mapping. I can't publicly post the verbatim email content here for privacy reasons (since it was a personal correspondence), but paraphrasing they wrote that: “_while not all data is currently released as CC0, you can use the all geoserver layers as the base for the manual mapping in OSM_”. Another confirmation of this can be found here: https://lists.openstreetmap.org/pipermail/talk-it/2012-February/026835.html (there, it is given specifically for the 2011 imagery):

> […] könnt ihr die orthofotos 2011 für "OSM-Zwecke" auf jeden fall benutzen.

Meaning that for "OSM purposes" it is definitely allowed to use the orthofotos 2011 imagery.

Do you need more confirmation than this?

andrewharvey:
> Do you need more confirmation than this?

I know it's a pain but the OSMF waiver https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates would be ideal if you can secure it.

rbuffat:
@simonpoole Maybe you can help clarify the situation. My understanding is that at least the attribution part of https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/ applies also to CC-BY imagery/maps that are used for tracing.

simonpoole:
In the case of aerial imagery/DTM/DSMs* the issue is a bit convoluted. 

What "we" really want is a confirmation from the provider that tracing from the imagery doesn't create a derived work and that they have no rights in any so created data regardless of the licence of the imagery. There is little to no specific case law on this topic (funny enough there is a case in Austria that concerned tracing from a photographic image, however that had a clear creative component, so is probably not really applicable to this situation), but I think it is fair to say that it would be an uphill fight for anybody actually trying to claim such rights in any jurisdiction I know something about. 

However asking for clarification is simply a matter of good form and should be done IMHO regardless of the legal merits of such claims.

Naturally the imagery still needs to be available on open terms or with explicit permission that we can display and use it in editors to start with, this is given in this case.

&#42; I would this consider this different from a layer that directly shows data as an image, just as a clarification.

PR #897, opened 2020-08-23
Fix sicily 2008, add sicily 2012

rbuffat:
- Fix broken url for "Sicily - Italy" source
- Add 2012 source from the same portal

While the WMS server does not indicate any access restrictions, the following page indicates the data is licensed under CC BY-SA 3.0 IT: http://www.sitr.regione.sicilia.it/geoportale/it/Home/Terms

For the ATA 2007-2008 layer, I found the following information in the osm wiki: https://wiki.openstreetmap.org/wiki/Sicilia/Comunit%C3%A0
@cortesimone you seemed to have had contact with the provider.  Is it necessary to reask for permission? If you are not https://wiki.openstreetmap.org/wiki/User:Simone, I apologize.

### DIFF
--- a/sources/europe/it/Sicily-Italy_2008.geojson
+++ b/sources/europe/it/Sicily-Italy_2008.geojson
@@ -20,7 +20,8 @@
             "required": true
         },
         "license_url": "http://www.sitr.regione.sicilia.it/geoportale/it/Home/Terms",
-        "privacy_policy_url": "http://pti.regione.sicilia.it/portal/page/portal/PIR_PORTALE/PIR_Privacy"
+        "privacy_policy_url": "http://pti.regione.sicilia.it/portal/page/portal/PIR_PORTALE/PIR_Privacy",
+        "category": "photo"
     },
     "geometry": {
         "type": "Polygon",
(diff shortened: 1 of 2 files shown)


andrewharvey:
If we don't have any explicit permission and the source is CC BY SA 3.0, then we'll need to obtain the CC BY waiver or equivalent to include this.

If the Italian OSM community could chime in here?

cortesimone:
I just wrote to my contacts there, will update this issue with the relevant news.

rbuffat:
@cortesimone did you already heard something back? 
Also is this imagery the same as https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/it/PCN2012-Italy.geojson and we could remove it instead?

cortesimone:
hi, I spoke to them over an email thread, they say they are fine with the use we make of the images, but they have CC-BY-SA. not much more whey can do about that.

rbuffat:
@cortesimone thanks for asking! I checked and they are not identical to the PCN sources. As they are CC-BY-SA we should either document the permission to use them or remove the sources from ELI. 

There are templates for a waiver https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates (As the data is now CC-BY-SA, I would use the "Template text for aerial imagery waivers" template)
Alternatively, if they allowed the use of the data for OSM in your email conversation, we could post the conversation (anonymized) to osm wiki to document the permission. Important would be that they understand how their data is used, how attribution is done in OSM (using https://wiki.openstreetmap.org/wiki/Contributors), and what the license of the OSM data.

PR #912, opened 2020-08-31
Ukraine - Andriivska OTG 2020

bezdna:


andrewharvey:
Thanks for the PR.

> https://zakon.rada.gov.ua/laws/show/2939-17

Is there an evaluation of this if this is compatible with OSM or not?

bezdna:
> Thanks for the PR.
> 
> > https://zakon.rada.gov.ua/laws/show/2939-17
> 
> Is there an evaluation of this if this is compatible with OSM or not?

https://wiki.openstreetmap.org/wiki/Uk:Вікіпроект_Україна/Публічні_дані

andrewharvey:
Thanks @bezdna seems like this has had the due diligence done and okay to be used.

bezdna:
> Thanks @bezdna seems like this has had the due diligence done and okay to be used.

same for all other map tiles from Ukraine, please merge all other PR

Issue #602, opened 2018-12-12
LINZ imagery api key


andrewharvey:
NZ's LINZ imagery was using a shared API key for access to the web services. That API key has been revoked (by Koordinates) and we can't pre-bundle a key going forward according to the service provider (Koordinates), so these sources are now broken.

OSM contributors can still use the LINZ imagery in iD and JOSM but they need to manually create their own API key and add a custom layer which isn't ideal for the casual mapper, but can work for power mappers.

The alternative is we need to host this imagery elsewhere either on OSMF or local chapter infrastructure, or donated via Mapbox.

simonpoole:
See https://github.com/osmlab/editor-layer-index/pull/548 (if implemented the editors could if they saw fit display a popup and ask for the key).

How much (size wise) imagery are we talking about?

andrewharvey:
> See #548 (if implemented the editors could if they saw fit display a popup and ask for the key).

That would work.

> How much (size wise) imagery are we talking about?

Imagery is 10TB as TIFF, 1.3TB as JPG, 966GB as JPEG2000, which costs $540 to order (it's over the 3.5GB download limit).

Topo is 10GB as JPG or 20GB as TIFF

simonpoole:
@Firefishy any thoughts? At least for us (SOSM) while we have quite a lot of infrastructure available and host a number of sources, we would really need to think about the disk space requirements.

Klumbumbus:
>That API key has been revoked and we can't pre-bundle a key going forward according to the service provider (Koordinates), so these sources are now broken.

For me the two sources seem to work fine in JOSM.

Firefishy:
Would they allow us to have an OSM proxy server which added the key?

andrewharvey:
> For me the two sources seem to work fine in JOSM.

Actually that's the case for me too. They said they revoked the API key so I don't know.

> Would they allow us to have an OSM proxy server which added the key?

I'll ask.

grischard:
> Would they allow us to have an OSM proxy server which added the key?

This is what the French community does for the IGN Orthophotography, actually. The French RP whitelists some user agents and referers, and has an api key for everyone else.

grischard:
@andrewharvey 	I was just talking about API keys, and wondering if you'd been able to ask whether an OSM proxy would be ok. Otherwise, I don't really want to require casual mappers to register for API keys.

andrewharvey:
[...]

PR #913, opened 2020-09-05
Update NZ Aerial Imagery url to use authoritative LINZ Basemaps service

dwsilk:
LINZ has released a new authoritative basemaps service. This will be updated more frequently than the LINZ Data Service set that the current caching proxy server uses as its source.

If you encounter any issues with this service, please raise them on the issue tracker at https://github.com/linz/basemaps.

Thanks to @Cameron-D for setting up the caching proxy server that was used to reinstate access to aerial imagery sourced from LINZ in the interim (ref: https://github.com/osmlab/editor-layer-index/issues/602).

### DIFF
--- a/sources/oceania/nz/LINZ-NZ-Aerial-Imagery.geojson
+++ b/sources/oceania/nz/LINZ-NZ-Aerial-Imagery.geojson
@@ -6,10 +6,11 @@
             "text": "Sourced from LINZ CC-BY 4.0",
             "required": true
         },
-        "license_url": "https://wiki.openstreetmap.org/wiki/Contributors#LINZ",
+        "license_url": "https://wiki.openstreetmap.org/wiki/Contributors#New_Zealand",
         "name": "LINZ NZ Aerial Imagery",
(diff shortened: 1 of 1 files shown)


dwsilk:
Note that this service also provides `webp` which has smaller file sizes than `png` with the same quality, but I didn't see any other services using `webp` here and don't want to break stuff.. so have added a `png` configuration.

andrewharvey:
> Note that this service also provides webp which has smaller file sizes than png with the same quality, but I didn't see any other services using webp here and don't want to break stuff.. so have added a png configuration.

Since this is aerial imagery can we use `jpg`?

I'd prefer `webp` too, but since we don't have a way in the editor layer index currently to list multiple image format URLs, and since the latest Safari still doesn't support webp https://caniuse.com/#feat=webp (though looks like it's slated for the next release) we should hold off on `webp` for now.

dwsilk:
Yep that makes sense.

I have force pushed a single commit that should be ready to merge.

andrewharvey (approved):

PR #921, opened 2020-09-10
Add more sources from data.tnris.org

rbuffat:
This PR adds more sources from data.tnris.org

All are Public Domain (Creative Commons CC0) 
https://creativecommons.org/publicdomain/zero/1.0/

See e.g. here: https://data.tnris.org/collection/3ae404a0-f976-4064-9d3e-81e5a719ca3c


sources/north-america/us/tx/San_Antonio_River_Orthophoto_2016_WMS.geojson
```diff
@@ -0,0 +1,264 @@
+{
+    "type": "Feature",
+    "properties": {
+        "license_url": "https://data.tnris.org/collection/bfdb7a76-a36c-4e9b-ba18-e8c8b259ed91",
+        "id": "san_antonio_river_2016_wms",
+        "type": "wms",
+        "name": "San Antonio River Authority Imagery 2016",
+        "url": "https://webservices.tnris.org/arcgis/services/StratMap/StratMap16_NC_CIR_12in_SARA/ImageServer/WMSServer?LAYERS=0&STYLES=default&FORMAT=image/jpeg&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap",
+        "start_date": "2016",
```
andrewharvey, line 9: the description says it's late march, so "2016-03" could be used as the start/end date, though what you've got here is good enough.

andrewharvey (approved):
I'll ignore the CI fail because the URLs are accessible for me.

sources/north-america/us/tx/San_Antonio_River_Orthophoto_2016_WMS.geojson
rbuffat: that's a good point. I kind of never bothered to look further as the year. But in areas with a lot of imagery coverage, a more precise date could be useful!

rbuffat:
@andrewharvey thanks for looking at the PRs. Enjoy your weekend!

Klumbumbus:
The shape if Georgetown is about 500km off ;)

rbuffat:
@Klumbumbus Thanks for reporting this. It looks as I forgot to update the geometry. I will create a PR to fix this.

PR #922, opened 2020-09-11
Use JOSM categories when not present in ELI

rbuffat:
This PR imports categories from JOSM if source id and country code match and the ELI source contains no category using the following script: https://github.com/rbuffat/editor-layer-index/blob/ada2de2604ca22a593e714d4ba52d37b46ee5dfb/scripts/sync_categories.py


andrewharvey:
should be okay licensing wise because JOSM index is CC BY SA, which is the same as ELI's license.

looks like a lot of other formatting changes, should be okay though.

rbuffat:
@andrewharvey The formatting changes are as the scripts read the json files and then writes them with 
`json.dump(source, out, indent=4, sort_keys=False)`
A lot of the sources use an indent of 4, but all other will have significant changes in whitespace. The order of the attributes should however be conserved. Preserving the whitespace would be more complicated to implement and I think it's not worth the effort.

JSOM mentions `License: Like the whole wiki (see page footer) the imagery database is licensed under Creative Commons (CC-BY-SA). Content since April 2014 is dual licensed with LGPL (except contents copied from ELI, which is only CC-BY-SA).`: https://josm.openstreetmap.de/wiki/Maps, but never mentions which sources are copied from ELI. Thus I would not worry too much here...

rbuffat:
@andrewharvey @grischard I think it would be good to decide soon if this should be merged or not. Otherwise, the PR needs constantly to be updated, which would be a waste of time.

grischard:
I don't think iD uses the categories, but am always in favour of anything that reduces the delta between JOSM and ELI. Merge away!

PR #924, opened 2020-09-11
Add wms.openstreetmap.fr sources

rbuffat:
These sources are imported from JOSM
All use https://wms.openstreetmap.fr as host for the imagery and there is no privacy policy for this domain.


rbuffat:
Merge based on the assumption that the folks from wms.openstreetmap.fr only include imagery that can be used.

PR #925, opened 2020-09-12
Create Susono-orthophoto.geojson

kudarisenmon:
added Susono city orthophoto


andrewharvey (approved):
Thanks for including the licensing details. I found the waiver at https://wiki.openstreetmap.org/wiki/File:Susono_%E8%A8%B1%E8%AB%BE%E6%9B%B8.pdf so all good here.

PR #929, opened 2020-09-25
Fix min_zoom for oceania / au / qld sources

rbuffat:
These sources seem only to deliver tiles for zoom level 4 and above.


andrewharvey:
Actually I was thinking of removing this source, I added it a bit too prematurely because who LWG had yet to make a determination if the waiver we got was sufficient because it was not a standard one.

rbuffat:
@andrewharvey In case you want to delete them, would it be better to merge this PR before? If the LWG decides in favor, the source can be restored and this change is not lost.

andrewharvey:
Okay I'll merge, delete and then if we get back an okay from LWG can restore.

PR #930, opened 2020-09-26
Update geometries for north-america / us / tx

rbuffat:
Use geometries provided by data.tnris.org
See https://github.com/osmlab/editor-layer-index/pull/921#issuecomment-698557224


rbuffat:
The test fails as the requests library is unable to check https://data.tnris.org/collection. I'm still not sure why. However, the links work in the browser.

gdt:
I looked at the travis report and grabbed the URL.   I can open it fine in firefox, but when using "wget" I get a 404.  I wonder if the site behaves differently based on user agent, or something like that.

andrewharvey (approved):
based on what @gdt pointed out I think this is fine to fail

rbuffat:
Thanks @andrewharvey!

PR #934, opened 2020-09-30
Added 2019 10cm ortho and LiDAR hillshade for Luxembourg

DavidMoraisFerreira:
Hi,
Both the 10cm ortho and the [LiDAR hillshade](https://lidar-hillshade-2019.openstreetmap.lu/) have been available in JOSM for a while now. Instead of sending custom urls for iD editors, we might as well add it directly to ELI.

David


andrewharvey (approved):
https://data.public.lu/en/datasets/bd-l-ortho-webservices-wms-et-wmts/ confirms CC0

Issue #876, opened 2020-08-07
Requesting the addition of an imagery layer depicting current Poznan, Poland


Asteliks:
Requesting the addition of an imagery layer "ortogpkg:ortofotomapa_biezaca". This layer is updated every two years, currently displaying Poznan of 2018 (Poland) but will be updated later this year to display Poznan 2020. When updated the server adres will stay the same so adding this layer will allow us to always have the latest imagery layer.

## URL of the imagery
Server link: http://wms2.geopoz.poznan.pl/geoserver/gwc/service/wms

WMTS: http://wms2.geopoz.poznan.pl/geoserver/gwc/service/wms?request=GetCapabilities&version=1.1.1&tiled=true

or: http://wms2:8080/geoserver/gwc/service/wmts?REQUEST=GetCapabilities

"on demand" version [WMS]: http://wms2:8080/geoserver/ortogpkg/wms

List of services with the requested one marked ("ortogpkg:ortofotomapa_biezaca"):
link: http://sip.geopoz.pl/sip/uslugi/get_uslugi_lay/id_rap/99/v_serwer_r/29
![image](https://user-images.githubusercontent.com/62251475/89661891-9227c380-d8d3-11ea-942b-8ceb0a7f3ee7.png)

link to list server services: http://sip.poznan.pl/sip/uslugi/get_uslugi_lay/id_rap/99/v_serwer_r/29

## Two letter country code
PL

## Desired layer name
PL: Poznań: Orthophotomap bieżąca (aerial image)
EN: Poznań: Orthophotomap current (aerial image)

## License
https://www.geopoz.pl/bip/index.php?t=200&id=2449

## Is the license compatible with OSM?
Yes. As far as I'm concerned the license hasn't changed and we are already using at least two of their maps ("Poznań: Orthophotomap 2016 (aerial image)" and "Poznań: Orthophotomap 2014 (aerial image)"). According to the license we can use the map freely but need to mention the source "Zarząd Geodezji i Katastru Miejskiego GEOPOZ".

## Privacy policy (URL)
https://www.geopoz.pl/bip/index.php?t=200&id=2449

PR #935, opened 2020-10-01
added the current layer for Poznan (PL)

Asteliks:
This pull request solves the [Poznan new map](https://github.com/osmlab/editor-layer-index/issues/876) issue

### DIFF
--- a/sources/europe/pl/PoznaOrthophotomapBiezaca(aerialimage).geojson
+++ b/sources/europe/pl/PoznaOrthophotomapBiezaca(aerialimage).geojson
@@ -3,7 +3,7 @@
     "properties": {
         "name": "Poznań: Orthophotomap Bieżąca (aerial image)",
         "type": "wms",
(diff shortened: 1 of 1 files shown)


Asteliks:
On my side ready for merge, sorry for the commits during the request

Asteliks:
thanks

maro-21:
Write please either in English or in Polish... "Poznań: Orthophotomap Bieżąca (aerial image)" isn't a good name. The name should be in English and you should add `"i18n": true"` so it can be translatable. Thank you.

"Bieżąca" does't say much - why not just add a year when the photos were taken?

Asteliks:
> Write please either in English or in Polish... "Poznań: Orthophotomap Bieżąca (aerial image)" isn't a good name. The name should be in English and you should add `"i18n": true"` so it can be translatable. Thank you.

Naming was coordinated with the local authorities responsible for the creation of the source, what is more there is no explicit text in the documentation saying that the name must be in English [source](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json). That is why we have several sources not named in English [source](https://osmlab.github.io/editor-layer-index/).

> "Bieżąca" does't say much - why not just add a year when the photos were taken?

There is no year as this source will always present the latest available orthophoto map

PR #937, opened 2020-10-03
Sync europe / se

rbuffat:
This PR adds 2 sources from the JOSM index:

The license of the lidingo-orto source is CC0:  https://lidingo.entryscape.net/catalog/1/datasets/90/ => Ortofoto 2012 via Karttjänst => Creative Commons CC0 1.0 (Public Domain Dedication, No Copyright)

The sollentuna-orto source does not state a specific license, only that the data is open: https://www.sollentuna.se/kommun--politik/kvalitet-och-utveckling/oppna-data-i-sollentuna/flygbilder/ respectively https://www.sollentuna.se/kommun--politik/kvalitet-och-utveckling/oppna-data-i-sollentuna/

Google translate:
```
Open data is information that is available for anyone to use, reuse and share. Open data can, for example, be used to create new digital services.
```


sources/europe/se/LidingOrthophoto.geojson
```diff
@@ -0,0 +1,78 @@
+{
+    "type":"Feature",
+    "properties":{
+        "id":"lidingo-orto",
+        "name":"Liding\u00f6 Orthophoto",
```
andrewharvey, line 5: Is there a reason UTF characters can't be used directly?

PR #940, opened 2020-10-04
Fix north-america / us / forest service overlay url

rbuffat:
Sync currently not working "U.S. Forest Roads Overlay" source with JOSM


andrewharvey (approved):

simonpoole:
I asked Clifford ~2 weeks about this entry and didn't get a clear answer back, 

IMHO https://github.com/osmlab/editor-layer-index/blob/f20589a74fd20e78c4222304f9148ce7c79c1a18/sources/north-america/us/USForestService.geojson is the same data, works and @systemed is I believe OK with it being listed here.

rbuffat:
@simonpoole is this a suggestion to remove this source in favor of the osm.cycle.travel source? I suppose the main difference to the other source is that this source is marked as overlay and has a transparent background. There is a discussion [here](https://josm.openstreetmap.de/ticket/16161 ).

Ideally, it would be good if each source would include the license and documentation that the data can be used (which is lacking for this source).

Issue #905, opened 2020-08-24
User-Agent encoded in URL


rbuffat:
A few sources encode "User-Agent" in source['properties']['url']:
https://github.com/osmlab/editor-layer-index/search?q=User-Agent&unscoped_q=User-Agent

Is this something that is expected. If not, should the tests be adapted to check for this?

simonpoole:
Seems to be a, slightly crazy, JOSMism. It simply defines a header to be set/added on the request for the image.

BUT this should simply be an additional field in the configuration (and that goes for JOSM too).

In ELI this should be added to custom-http-headers though I suspect this might not be necessary for other clients.

rbuffat:
The schema specifies custom-http-headers as follows. I'm not an expert with JSON schema, but would this not only allow to specify only one custom HTTP header?
```
"custom-http-headers": {
    "description": "HTTP headers to be sent to server. It has two attributes header-name and header-value. May be specified multiple times.",
    "type": "object",
    "properties": {
        "header-name": {
            "type": "string"
        },
        "header-value": {
            "type": "string"
        }
    },
    "additionalProperties": false
},
```

This should allow to specify multiple headers:
```
    "custom-http-headers": {
        "type": "array",
        "items": {
            "description": "HTTP headers to be sent to server. It has two attributes header-name and header-value. May be specified multiple times.",
            "type": "object",
            "properties": {
                "header-name": {
                    "type": "string"
                },
                "header-value": {
                    "type": "string"
                }
            },
            "additionalProperties": false
        }
    }
```

grischard:
Or since it's a JOSMim and likely won't be necessary for ELI consumers, we can drop it in the URL.

Klumbumbus:
It doesn't need to be in the URL for JOSM as well afaik. JOSM has the `custom-http-headers` property for this.

rbuffat:
@grischard They can easily be moved to `custom-http-headers`, but could you look at the schema for it? 

https://github.com/osmlab/editor-layer-index/blob/cbf072b61508bddbe2a6dcbf617ed699defe0a40/schema.json#L210

Let's imagine a source needs two custom HTTP headers, e.g. a `X-WAAPI-TOKEN` and a `User-Agent` header. According to the current schema, I think these headers need to be specified as follows (please correct me if I'm wrong, I'm not that used to JSON schema):
```
{
    "type": "Feature",
    "properties": {
       ...,
        "custom-http-headers": {
[...]

PR #943, opened 2020-10-04
Move http headers from url to custom-http-headers

rbuffat:
See https://github.com/osmlab/editor-layer-index/issues/905

Also added category, privacy policy and updated wms URLs and available projections.


andrewharvey:
Sounds good, just need to resolve the conflicts.

PR #952, opened 2020-10-12
imagico.de osmim 2020-01 update

imagico:
Changes:

* moving urls to https
* adding additional properties (`category`, `permission_osm`, `privacy_policy_url`)
* some new sources

I also have made available another attribute `use` for all layers you can find in the tilejson descriptions:

https://imagico.de/map/osmim_tiles.php?layer=S2B_R022_N47_20191026T101029

which has a value of either `main`, `supplemental` or `old` - which indicates a subjective assessment of the primary use of the layer for mapping:

* `main` indicating it can well serve as the main source for remote mapping in the area, 
* `supplemental` indicating it is probably useful primarily as a supplemental source of information in addition to other image sources or local knowledge, for example autumn color images and 
* `old` indicating the image is of only limited use any more - usually because it has been superseeded by other, newer image sources - and is only there as a reference and possibly for resolving ambiguities in other images.


Klumbumbus:
The date of "Landsat off-nadir early July 2020" and "Landsat off-nadir late July 2020" is both 2020-07-12. Typo?

PR #685, opened 2019-07-08
Update aerial imagery in Capetown, South Africa

tsmock:
Add 2017 and 2018 imagery from Capetown, South Africa.

Largely copied from the 2015 aerial geojson file, with modifications for current years. I did leave a "best" tag in the 2018 imagery properties.

-----

This should fix the travis-ci build error ( https://travis-ci.org/osmlab/editor-layer-index/builds/555263087 ) from the last pull request.

PR #953, opened 2020-10-13
Fix africa / za / capetown imagery

rbuffat:
The "2017 Aerial Imagery from the City of Cape Town" and "2018 Aerial Imagery from the City of Cape Town" are currently broken. The 2018 imagery is specified as tms but the url does not work. The 2017 source is specified as wmts but without a wmts url.

This PR:
- changes the existing sources to wmts with the correct WMTSCapabilities.xml url
- adds the same sources as esri rest / wms so that they can be used in a wider range of editors


rbuffat:
The ssl certificate for http://odp.capetown.gov.za/datasets/aerial-imagery is not valid, but https://www.arcgis.com/sharing/rest/content/items/739759d8127f4d1f9ba8ef9019878147/info/metadata/metadata.xml?format=default&output=html contains the same information.

rbuffat:
Changes in this PR are not related to the license, which was discussed in https://github.com/osmlab/editor-layer-index/pull/685

PR #955, opened 2020-10-15
LPI is no more updating NSW DCS Spatial Services sources

andrewharvey:
- Update attribution text per https://www.spatial.nsw.gov.au/copyright
- Update attribution URL to current link
- Update license URL to new LWG waiver completed
- Update privacy policy URL
- Update names to remove reference to LPI which does not exist and not used by the source anymore, 

will ask local community for feedback before merging

### DIFF
--- a/sources/oceania/au/nsw/NSW-WebServices-AdministrativeBoundaries-County.geojson
+++ b/sources/oceania/au/nsw/NSW-WebServices-AdministrativeBoundaries-County.geojson
@@ -1,19 +1,19 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "LPI_NSW_Administrative_Boundaries_NPWS_Reserve",
+        "id": "LPI_NSW_Administrative_Boundaries_County",
         "attribution": {
-            "url": "https://www.spatial.nsw.gov.au/mapping_and_imagery/lpi_web_services",
-            "text": "© Department of Customer Service 2019",
+            "url": "https://www.spatial.nsw.gov.au/products_and_services/web_services/access_web_services",
+            "text": "© State of New South Wales (Spatial Services, a business unit of the Department of Customer Service NSW). For current information go to spatial.nsw.gov.au.",
             "required": true
         },
-        "license_url": "https://wiki.openstreetmap.org/wiki/Attribution/New_South_Wales_Government_Data",
-        "privacy_policy_url": "https://maps.six.nsw.gov.au/js/sixmaps/app/coreTerms.html",
-        "name": "LPI NSW Administrative Boundaries NPWS Reserve",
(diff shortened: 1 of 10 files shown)


rbuffat:
@andrewharvey Your PR had the honor to be the first with the new check script! 

`INFO:shapely.geos:Hole lies outside shell at or near point 140.998648 -28.9989332`
This is a warning from the shapely module and not the tests. I probably should suppress it, at least until there is a solution for https://github.com/osmlab/editor-layer-index/pull/857

`WARNING:root:Layer '6': CRS 'EPSG:4326,CRS:84' not included in available_projections but supported by server.`
This means that the WMS server supports the projections EPSG:4326 and CRS:84, but they are not included in the available projections. I chose these two projections as well as EPSG:3857 as I think they are most likely to be supported by an OSM editor (they require minimal transformation). What is your opinion about this warning? Should it be removed or should it be even an error?

`WARNING:root:Zoom level 19 not reachable. (Tested: 1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19) Tiles might not be present at tested location: 150.24284708393066,-31.545074200000002`
The script tests all levels for one location. If no level is reachable it will emit an error, if at least one is not found a warning is emitted. As discussed in https://github.com/rbuffat/eli_watchdog/issues/2, not all zoom levels might be present at every location. 

`ERROR:root:Source has no category`
Before it was not enforced to have a category attribute. But as at least one editor uses it and if ELI should be a universal index it would be good to include it.   

I would be happy to hear your feedback.

andrewharvey:
> This is a warning from the shapely module and not the tests. I probably should suppress it, at least until there is a solution for #857

Agreed.

> This means that the WMS server supports the projections EPSG:4326 and CRS:84, but they are not included in the available projections. I chose these two projections as well as EPSG:3857 as I think they are most likely to be supported by an OSM editor (they require minimal transformation). What is your opinion about this warning? Should it be removed or should it be even an error?

Most editors would be using EPSG:3857. I know JOSM can render in WGS84 but only if you go and change the setting, but if the source supports it no issue listing it here, but it's more just something that could be improved, not necessarily anything that is wrong with how it is.

I wouldn't treat it as a warning, maybe just a notice.

> The script tests all levels for one location. If no level is reachable it will emit an error, if at least one is not found a warning is emitted. As discussed in rbuffat/eli_watchdog#2, not all zoom levels might be present at every location.

Yeah I believe that in this case the highest zoom level is only available in some areas.

> Before it was not enforced to have a category attribute. But as at least one editor uses it and if ELI should be a universal index it would be good to include it.

No problem I can add it.

andrewharvey:
> will ask local community for feedback before merging

Feedback seemed okay with the rename, though I have left the ID the same, not sure who or how the ID is used though.

PR #950, opened 2020-10-07
Category update to old Poznań maps

Asteliks:
Added the category historicphoto to maps form 2014 and 2016 as there is a new map in town called [PoznaOrthophotomapBiezaca(aerialimage)](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/pl/PoznaOrthophotomapBiezaca(aerialimage).geojson)

PR #957, opened 2020-10-17
License url and privacy url updated

Asteliks:
Geopoz webpage has been updated on 14.10.2020 thus making some of the links outdated. See https://github.com/osmlab/editor-layer-index/pull/950 for details


andrewharvey:
While the fix to the URLs along looks great, the holdup for me is I'm not sure if this license is enough to allow use within OSM.

> the obligation to inform ZGiKM GEOPOZ within 14 days about the processing of reused information, in writing in any way, including to the following address: geopoz@geopoz.poznan.pl ,

This is unclear, do downstream OSM users need to do this, we can't expect them to.

> obligation to provide the date of generating information and the date of obtaining the information

All we can do is list on the wiki a date range of when data was extract from (start of when added to ELI to now), would that be sufficient?

In short, when it's got the LWG waiver, or public domain/cc0 it's clear, but a custom license like this it's not clear to me, so I'd like to see more information about if we can use this in OSM.

Asteliks:
I have contacted a person that has been immensely helpful in the past from Geopoz. They may change their license for us but that remains to be seen.

Asteliks:
@andrewharvey the author of the GEOPOZ maps has joined the conversation in https://github.com/osmlab/editor-layer-index/pull/950 I can confirm his identity based on their [web page](http://sip.geopoz.pl/sip/kontakt/). Could you take care of the formalities as I have no experience and easy way to document potential license agreements.

Asteliks:
These "new" links are no longer valid thus I'm closing this PR

PR #958, opened 2020-10-19
Update license urls for europe / ch / Geneva

rbuffat:
There is a new revision and the url has changed


rbuffat:
I did not spot any relevant difference in the license terms (although my French is not the best). The data is open source and regarding attribution, there is the OK that we can attribute the data over https://wiki.openstreetmap.org/wiki/Contributors#Switzerland:  https://wiki.openstreetmap.org/wiki/SITG_WMS

The strict test fails because the geometries are not optimal. But I would rather prefer to update the geometries when there is progress with https://github.com/osmlab/editor-layer-index/pull/857

PR #959, opened 2020-10-20
Update WMS parameters using GetCapabilities request

rbuffat:
WMS sources have an available_projection parameter that includes the projections supported by the WMSs server. While we can check if these projections correspond to the ones advertised by the server when the source is added, the server configuration can change at any time. To keep the sources up-to-date, it would be good to synchronize them periodically. 

This PR adds a script to synchronize WMS sources with the corresponding WMS servers and updates the sources.
WMS URLs are upgraded to the latest WMS version the server supports
The 'available_projections' attribute is updated using the projections advertised by the server

The script currently only updates sources, when:
- The URL contains only 1 layer
- The image hash for a test image is the same for the old URL as well as the updated URL
- ~The WMS server advertises less than 10 projections~ If server advertises more than 15 projections, the previously specified projections + EPS:4326, CRS:84 and EPSG:3857 are filtered if they are (still) advertised by the server


grischard:
Good idea! It should filter out synonymous projections, for example EPSGs 900913 3587 54004 41001 102113 102100 3785 which are alternative codes for 3857.

rbuffat:
- Filter EPSG:3857 alias if EPSG:3857 is advertised
- Some servers do not advertise EPSG:3857 or EPSG:4326. ~If one of these projections was previously included it is checked if they still work~. Include these projections when the same image as with the original configuration is returned.

rbuffat:
@andrewharvey @grischard I'm experimenting with a github action that will run as a cron job and creates automatically a PR: https://github.com/rbuffat/editor-layer-index/pull/1
https://github.com/rbuffat/editor-layer-index/blob/sync_wms_action/.github/workflows/sync_wms.yml
This cron job could e.g. run once a week, or once a month.

The action could also be configured to automatically deploy, but I suppose it would be better if there is manual control, at least until the script has proven to be very reliable.

rbuffat:
- Update sources only when either url or projections have changed

The Github Action seems to work. Currently, the action is configured to run every Saturday. If there are changes, the action creates a PR. If a previous PR by the action is still open, the action force-pushes to this PR.

If there are no objections I would merge this PR this week.

andrewharvey:
> If there are no objections I would merge this PR this week.

Sounds good to me.

PR #965, opened 2020-10-25
Improve MassGIS layers (us/na/ma)

gdt:
Fix min_zoom in MassGIS 2019 Orthos (thanks to josm for pointing this out; confirmed experimentally)

Improve metadata for L3 parcels layer.

### DIFF
--- a/sources/north-america/us/ma/MassGIS2019Orthos.geojson
+++ b/sources/north-america/us/ma/MassGIS2019Orthos.geojson
@@ -8,7 +8,7 @@
     "url": "https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/USGS_Orthos_2019/MapServer/tile/{zoom}/{y}/{x}",
     "country_code": "US",
     "best": true,
-    "min_zoom": 6,
+    "min_zoom": 7,
     "max_zoom": 20,    	      
     "attribution": {
         "url": "https://docs.digital.mass.gov/dataset/massgis-data-usgs-color-ortho-imagery-2019",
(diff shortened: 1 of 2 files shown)


andrewharvey (approved):
Thank you.

andrewharvey:
whoops just realised this failed due to a missing comma... I'll followup to fix it.

gdt:
I  force-pushed the comma after I saw the CI results, sorry.

andrewharvey:
> I force-pushed the comma after I saw the CI results, sorry.

doesn't look like it came through, it's okay I followed up with another PR to address it.

gdt:
Thanks for merging and for fixing and sorry for the mess.

Klumbumbus:
https://www.esri.com/en-us/privacy/overview as privacy policy for MassGIS? Is that really correct?

andrewharvey:
> https://www.esri.com/en-us/privacy/overview as privacy policy for MassGIS? Is that really correct?

The tiles are served by https://tiles.arcgis.com so yes that seems correct. The privacy policy should be for the tile server not the data owner.

Klumbumbus:
Right, I didn't know arcgis is a esri service.

gdt:
Yes it is correct, although I think from a public policy point of view not quite right.    The state has a privacy policy for state websites and points out that websites that they refer to or that serve data from them have their own policies.  I actually brought this up with MassGIS (corresponding with a human) and they confirmed the ESRI policy notion (and that ESRI only logs individual information for modify operations).   I believe MassGIS is going to update their website to make this clearer.

PR #972, opened 2020-10-31
Add warning to be cautious when adding new supported WMS versions

rbuffat:
See https://github.com/osmlab/editor-layer-index/pull/971#issuecomment-719909217

@simonpoole Would this resolve your concerns?


simonpoole:
Sure.

PR #975, opened 2020-10-31
NYS Orthos Online has moved

kennykb:
The imagery for New York State Orthos Online was moved to a different server. This change provides the updated URL and metadata.


sources/north-america/us/ny/NYS_Orthos_Online.geojson
```diff
@@ -1315,14 +1315,14 @@
...
-        "end_date": "2018",
+        "end_date": "2020",
         "icon": "https://osmlab.github.io/editor-layer-index/sources/north-america/us/ny/NYS_Orthos_Online.png",
         "id": "orthos.dhses.ny.gov_latest",
         "max_zoom": 19,
         "name": "NYS Orthos Online",
         "start_date": "2013",
         "type": "wms",
-        "url": "https://orthos.dhses.ny.gov/arcgis/services/Latest/MapServer/WmsServer?FORMAT=image/jpeg&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=0,1,2,3,4&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
+        "url": "https://orthos.its.ny.gov/arcgis/services/wms/Latest/MapServer/WmsServer?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=0,1,2,3,4,5,6&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}"
```
rbuffat: There is a comma missing here: BBOX={bbox}",

rbuffat:
@kennykb Thanks for creating this PR! I think there is a comma missing (See review above). 

It would also good to include a "license_url" and "privacy_policy_url". The license_url should point to a webpage that states the license terms of the data, the privacy_policy_url to a page where it is stated what data about users are saved.

kennykb:
@rbuffat - Thanks for noting the missing comma. I needed to leave for a few hours right after pushing the changes, and didn't have a chance to check on Travis until a short while ago.

I added the link to the privacy policy.

Nothing on the NYSGIS site has a license agreement available on line. Richard Welty and others have email trails confirming that the data are available and redistributable without fee for any lawful purpose; the state regards them as being in the public domain. I linked to the final report of the (initial) project, which has similar language. Otherwise, I'd have to link to things like letters from the NYS Committee on Open Government offering opinions that the state GIS data are Public Domain by law.

I think we're now down to an overly fussy geometry check. The problem is that the tiles to be served are a composite of five layers from images acquired in different years. No layer covers the entire state, thus the check of the "geometry" in the GeoJSON against the layer bounding box will fail for all five of them. Nevertheless, the composite does have complete coverage of the state. I don't know what to do about that in the GeoJSON, short of having a separate layer for each year and abandoning the 'latest' layer, which would be really inconvenient - a user of an editor would have to search for which year's imagery covers the region of interest.

Is this last failure fixable or overridable on your end?

rbuffat:
> I think we're now down to an overly fussy geometry check. The problem is that the tiles to be served are a composite of five layers from images acquired in different years. No layer covers the entire state, thus the check of the "geometry" in the GeoJSON against the layer bounding box will fail for all five of them. Nevertheless, the composite does have complete coverage of the state. I don't know what to do about that in the GeoJSON, short of having a separate layer for each year and abandoning the 'latest' layer, which would be really inconvenient - a user of an editor would have to search for which year's imagery covers the region of interest.

Thanks for reporting this. This a bug in the check script and should be fixed with https://github.com/osmlab/editor-layer-index/pull/973/commits/5cb88b5ba06e6f0a9668e7750d8ae14458e48157

> Nothing on the NYSGIS site has a license agreement available on line. Richard Welty and others have email trails confirming that the data are available and redistributable without fee for any lawful purpose; the state regards them as being in the public domain. I linked to the final report of the (initial) project, which has similar language. Otherwise, I'd have to link to things like letters from the NYS Committee on Open Government offering opinions that the state GIS data are Public Domain by law.

Ideally, it is clear from only the information in the geojson what the license situation is. It's a bit unlucky that they do not use more clear language on their webpage. From what I see there seems to be no mentioning of any usage limitations. But if there was already contact with NYSGIS, would it possible to add the replies from the NYSGIS (e.g. in an anonymized version without names) to the osm wiki and then link to this page? Or would this be overkill? @andrewharvey what is your opinion about this?

rwelty1889:
i'm having discussions with NYS GIS and with some folks on the OSM US board about getting the details right, we'll probably also ask for some advice from the LWG. rather than rush into it, let's get these details right once and for all so we can move on and not worry about this again.

andrewharvey:
> Ideally, it is clear from only the information in the geojson what the license situation is. It's a bit unlucky that they do not use more clear language on their webpage. From what I see there seems to be no mentioning of any usage limitations. But if there was already contact with NYSGIS, would it possible to add the replies from the NYSGIS (e.g. in an anonymized version without names) to the osm wiki and then link to this page? Or would this be overkill? @andrewharvey what is your opinion about this?

While in my jurisdiction, everything is assumed copyrighted, and then only permissions are added through a license, but in the US case where something might not have legal copyright to begin with then I can see how the publisher wouldn't mention any copyright or license, so it's then up to us to make the call. I'm not too familiar with how this works in the US case, so as long as there is some documented reasoning or thought gone into the analysis then I think it's okay to merge. But,

> i'm having discussions with NYS GIS and with some folks on the OSM US board about getting the details right,

That's great, and in that case we'd be most happy to wait until this happens and if all good this can be merged.

sd:
@rwelty1889 any progress with the licensing discussion?

kennykb:
I am told that the terms of conditions are those for OPEN NY, found [here](https://data.ny.gov/api/views/77gx-ii52/files/ef0c1840-ad54-4240-92fd-6397c49fde46?filename=OPEN-NY_20Terms_20of_20Use.pdf). It's a long document, but there's a three-paragraph executive summary at its head:

> This website creates unprecedented transparency across all levels of government and gives the people
> user-friendly access to incalculable amounts of information about New York. We invite you to connect with
> government by freely using the information tools we provide here. We ask only that you abide by our
> Terms of Use, to protect against abuses of this website and its datasets. Then use our government
> information however you wish, in private sector projects or in government collaborations, to generate useful
> apps or to evaluate our performance.

> At their core, the OPEN-NY Terms of Service are among the least restrictive of any terms of service in
> furtherance of promoting the sharing, utilization, and reuse of Open Data. The OPEN NY Terms of Service
> do not contain restrictions requiring members of the public to use attribution, to re-post the license terms
> with any re-uses of the data, to impose share-alike or technical restrictions, nor require the public to obtain
> pre-approval before re-use of the data. We created these OPEN-NY-specific terms because we found
> that other commonly cited standardized terms are far more restrictive.

> OPEN-NY guards against one basic misuse of New York State's datasets -- doing something illegal with
> them (such as creating an app that purposely gives falsely dangerous driving directions). Keep your reuses lawful.
> So long as you are not doing anything malicious with NYS data, you may use it as you wish,
> subject to no other requirements. 

The data set in question, while hosted at a different site in the 'ny.gov' domain, is indeed an OPEN-NY data set,
confirmed by its being indexed at https://data.ny.gov/Public-Safety/Statewide-Digital-Orthoimagery/amsv-rmfz
with the notation 'DISCLAIMERS: Limitiations: No limitations on use of publicly available imagery.'

I'm also going to add the infrared layers to the GeoJSON.  I'll update the pull request when I'm done.

kennykb:
Oh!  It would save me quite a bit of work if you guys could update and test, given the troubles I've had with the validator (I'm doing 'cargo cult programming' on the GeoJSON).   

Everything in the original PR should stand except for adding the appropriate license links from the last comment.

The current URL's are:

Visible light:

`wms[19]:https://orthos.its.ny.gov/arcgis/services/wms/Latest/MapServer/WmsServer?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=0,1,2,3&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}`

Infrared:

`wms[19]:https://orthos.its.ny.gov/arcgis/services/wms/Latest_cir/MapServer/WmsServer?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=0,1,2,3&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}`

Both of these are tested in JOSM and work as of this evening.

grischard:
https://data.ny.gov/dataset/OPEN-NY-Terms-Of-Use/77gx-ii52 are unhelpful, especially when "stupid laws new york" is just a google search away. I think the intent is clearly there, and says that they would send a notice to cease using the data if they believe we are using it unlawfully. If there's no further feedback, I think we can reasonably use this one.

Friends don't let friends write custom data licences. They should have used CC-0.

kennykb:
Oh, I just noticed up-thread that the privacy policy URL was also wanted.  https://its.ny.gov/its-internet-security-and-privacy-policy

rbuffat:
That slipped by somehow, sorry for that. Thanks @kennykb @grischard for all the research!

PR #976, opened 2020-10-31
Fix europe / pl / Gliwice

rbuffat:
The following sources are not working at the moment:
- Gliwice: Buildings
- Gliwice: Orthophotomap 2013 (aerial image)

This PR updates the two sources and adds orthophotos for the year 2017, 2018 and 2020.

License:
The WMS servers are linked from https://msip.gliwice.eu/portal-planistyczny/uslugi-danych-przestrzennych.html 
This is a subpage for https://msip.gliwice.eu/portal-planistyczny/otwarte-dane.html

Which states: 

> OPEN DATA is complete and up-to-date information on local spatial development plans in force in the city of Gliwice. They are shared:
> 
> in an effective way (e.g. using dedicated portals and services) - and also used for any purpose, including economic purposes;
> in a permanent manner and without unnecessary barriers;
> without limitation, e.g. due to the rights, privileges of a specific group of people, etc .;
> in as little processed as possible (not in aggregated and modified form);
> in open data formats;
> in such a form that they can be read automatically.
> Openness of public data is the basis for modern city management, especially for:
> 
> transparent decision making,
> social participation,
> creating new values ​​and innovations by entrepreneurs and non-governmental organizations (including new services).

@jendrusk @Asteliks Could you have a look at this PR? Especially if there are more license terms that I have missed. The translation sounds good (any purpose, including economic purposes, without limitation), although it mentions nothing about attribution.

### DIFF
--- a/sources/europe/pl/GliwiceBuildings.geojson
+++ b/sources/europe/pl/GliwiceBuildings.geojson
@@ -4,34 +4,22 @@
         "id": "Gliwice-buildings",
         "name": "Gliwice: Buildings",
         "type": "wms",
-        "url": "http://185.60.246.14:9090/isdp/gs/ows?FORMAT=image/png&transparent=true&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=default:EGIB_budynek,default:pkt_adr&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
+        "url": "http://msip-mapa.um.gliwice.pl:9090/isdp/gs/ows/wms?LAYERS=default:EGIB_budynek,default:pkt_adr&STYLES=&FORMAT=image/png&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&TRANSPARENT=True",
         "country_code": "PL",
+        "overlay": true,
         "available_projections": [
-            "EPSG:3329",
-            "EPSG:3333",
-            "EPSG:3330",
-            "EPSG:3331",
-            "EPSG:3334",
-            "EPSG:3332",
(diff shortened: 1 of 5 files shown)


Asteliks:
It all seems good, but the given link does not look like a full license. I will call them on Tuesday because on Monday we have a holiday here

Asteliks:
Apparently today is only free for students thus I was able to contact the unit responsible for these sources and I got great news:
-Yes we can use their maps but note that even before they publish a new map on their server they sent it over to [GUGIK](http://www.gugik.gov.pl/) (the unit responsible for the main map of Poland)
-Apparently there has been a change in the Polish law (this is fresh just from 2 month ago) which makes all the Ortofotomaps are open and their local webpage licenses outdated. Based on [poz. 782](https://isap.sejm.gov.pl/isap.nsf/download.xsp/WDU20200000782/O/D20200782.pdf)

rbuffat:
@Asteliks Thanks for your research! I'm not really familiar with the situation in Poland.

> Yes we can use their maps but note that even before they publish a new map on their server they sent it over to GUGIK (the unit responsible for the main map of Poland)

There seems to be an existing layer for GUGIK
https://raw.githubusercontent.com/osmlab/editor-layer-index/2d0575624402911505a50c9b978c70abb34fae72/sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson

Thus, would it be better only to keep the Gliwice: Buildings layer?

Asteliks:
If I understand the new regulations correctly all the maps made by local authorities (excluding the private sector) are to be merged with the GUGIK nationwide layers thus making the need for splitting to local ones obsolete. Unfortunately, as of today the new regulations are not implemented everywhere. As such I would keep all the local maps for now and focus on implementing all the available layers from GUGIK to which we have open access by the new law. Link to GUGIK sources: https://www.geoportal.gov.pl/uslugi/usluga-przegladania-wms

Issue #974, opened 2020-10-31
geoserver.dados.al.gov.br not reachable


rbuffat:
The WMS server `http://geoserver.dados.al.gov.br:8080/geoserver/Alagoas/ows?` seems to be down for a long time (Although the server could just be geoblocked). This server is used for quite a few sources: https://github.com/osmlab/editor-layer-index/search?p=1&q=geoserver.dados.al.gov.br The URLs use WMS in version 1.1.0, which is really, really old. This could indicate that the server is shut down.

According[ to this site](http://dados.al.gov.br/dataset/imagens-de-alta-resolucao), the license of the data seems to be CC-BY-SA.   I did not yet find a waiver for this server. As this website still exists and links to this WMS server could indicate that the server is only temporarily down.

@naoliv @afmenez @willemarcel 
You seem to be from Brazil and have contributed to ELI. Do you know anything about the status of this server? Also, it would be helpful if you could help identify if the server is just geoblocked. I tried with https://perfops.net/ping-from-brazil, but not sure if this service is legit. Also, I tried to contact them via informacao@seplag.al.gov.br, but so far I got no response.

naoliv:
The server/service is down (I've tested from multiple places here, including an university).

Taking a look at the history of these entries, they seem to have assumed that the data could be used for any purpose, based on our information access law (Lei de Acesso à Informação - LAI).

I personally do not see it that way and I understand that the government can provide _access_ to some information, which does not mean unrestricted use of it; I also talked to some colleagues who have studied law and they share this view.

In this case, for example, the government provides access to the data, but the data is under the CC-BY-SA license (which is, in a way, restrictive for OSM).

I would remove/disable them until the license is further clarified.

I am filling out some forms asking about this (service and license); if I have any answer I will get back to you here.

rbuffat:
@naoliv Thanks for checking! I agree it would be better to remove the sources until there is further clarification.

Klumbumbus:
cc @Skippern

naoliv:
Sorry I forgot about this issue.
I filled a request in http://e-sic.al.gov.br, with protocol number `1951/2020` and the answer was:

[...]

PR #977, opened 2020-10-31
Remove sources using geoserver.dados.al.gov.br

rbuffat:
The WMS server geoserver.dados.al.gov.br is not working for some time. This PR removes all sources using this service until the license/server status is clarified. See discussion in https://github.com/osmlab/editor-layer-index/issues/974


rbuffat:
There were no objections to this, thus I merge it.

andrewharvey:
> There were no objections to this, thus I merge it.

All good, thanks for your efforts.

PR #978, opened 2020-11-01
Remove defunct europe / pl / Rzeszów sources

rbuffat:
The following sources are currently not working:
- Rzeszów: Buildings
- Rzeszów: Orthophotomap (aerial image)

There seems to be a geoportal of the City of Rzeszów: http://osrodek.erzeszow.pl that links to a wms server with a different url, however, it states: 

> The information contained in the Geoportal of the City of Rzeszów and the printouts generated from it do not constitute documents in administrative and other proceedings. They are for illustrative purposes only and may only be used for approximate identification and spatial location of land and building records and other database items. Copying and publishing data without permission is prohibited.

This does not look compatible with OSM. Thus I suggest removing these sources with this PR. 

The sources were added to ELI by an automatic JOSM sync: https://github.com/osmlab/editor-layer-index/commit/380df00dfb00848050037021c9d7c354d473a97e#diff-4885a4eb298861971ef7711a7fe19c5bb384e377fcf1e5750fa9b00ff51a76f0

The wms server seems also to be listed here: https://otwartedane.erzeszow.pl/dataset/serwer-wms
From google translation, I assume that `otwartedane` means open data. This sends some mixed signals.

@jendrusk @Asteliks @grischard Again I would be happy if you could have a look at it. It could be that I have overlooked something. 

CC @Klumbumbus because they were originally added to JOSM and maybe you know something more.


grischard:
Were you able to find out how these got added to JOSM?

grischard:
Looks like https://josm.openstreetmap.de/wiki/Maps/Poland?sfp_email=&sfph_mail=&action=diff&version=6&old_version=5&sfp_email=&sfph_mail= and https://josm.openstreetmap.de/wiki/Maps/Poland?sfp_email=&sfph_mail=&action=diff&version=28&old_version=27&sfp_email=&sfph_mail=

~~ImageryCompare has different URLs for the Rzeszów layers, and shows that difference as outdated/wrong in ELI (gold in the ImageryCompare list)~~ see below

grischard:
I can't find a formal licence text anywhere either.

I'm not familiar with how Poland has transcribed the EU open data directive, but it might not be legal for them to have this kind of restriction anyway. The Croatian approach that @hbogner has taken is that the data is, by law, automatically open.

https://dane.gov.pl/media/ckeditor/2019/07/04/open-data-good-practice-guide.pdf is the Polish government's good practice guide, which recommends adding a licence.

https://www.gov.pl/documents/31305/0/ponowne_wykorzystywanie_informacji_sektora_publicznego_-_podrecznik.pdf/dbcc1070-efd6-85d1-cf22-81df4e87b48a is the more detailed guide they link to, but my Polish isn't good enough to understand anything. It does seem to mention licences a lot.

@jendrusk would you know more?

Klumbumbus:
> @Klumbumbus because they were originally added to JOSM and maybe you know something more.

No, I have no more information.

Klumbumbus:
>ImageryCompare has different URLs for the Rzeszów layers, and shows that difference as outdated/wrong in ELI (gold in the ImageryCompare list)

No, the 2 sources of this pull request (ids `Rzeszow-buildings` and `Rzeszow-aerial_image`) are with the same URL in JOSM. Only the min_zoom differs.

Maybe you meant `rzeszowski-buildings` and `rzeszowski-aerial` which is in the ignorelist with different URLs?

hbogner:
> I'm not familiar with how Poland has transcribed the EU open data directive, but it might not be legal for them to have this kind of restriction anyway. The Croatian approach that @hbogner has taken is that the data is, by law, automatically open.

Just a clarification, all the published data should be open data automatically by law, with exceptions where institutions/agencies have the right to restrict the use of data in certain cases.
In this case they claim state aerial imagery is proprietary data which they sell and they offered to sell me the aerial imagery for +200 000€, and didn't reply to my request to use of their wms service. Later they changed terms of use for wms service which allows reuse and redistribution. They still won't allow us to use any vector data...

matkoniecz:
wait, why PR removing defunct layers has "waitfor - license / usage permission"?

PR #986, opened 2020-11-07
Limit WMS sync bot to osmlab/editor-layer-index repository

rbuffat:
I did not consider that this bot would run on every fork of Eli (as github actions unlike e.g. travis are activated by default).

This PR should limit the bot to this repository. Also, the scheduled time is moved to Sunday 00:00,  as then it should run when it's weekend all over the world.

### DIFF
--- a/.github/workflows/sync_wms.yml
+++ b/.github/workflows/sync_wms.yml
@@ -2,14 +2,13 @@ name: ELI WMS sync
 
 on:
   schedule:
-    - cron: '0 0 * * 6'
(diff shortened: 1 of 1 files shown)


rbuffat:
@andrewharvey I'm merging this right now to limit the number of affected forks.

andrewharvey:
> @andrewharvey I'm merging this right now to limit the number of affected forks.

all good, I'm short on time and your PRs shouldn't be held up, so if you think it's good and there's no opposition, please go ahead.

PR #928, opened 2020-09-25
Update and rename itu-lotes to SP_Itu_-_Lotes

IgorEliezer:
Hello,

We're committing a new geojson for this city (ref #835). Please review and tell us if something is missing or not working.

Regards,

PR #988, opened 2020-11-12
Check bounds to be within EPSG:4326 range

rbuffat:
Check that all geometries are within the EPSG:4326 bounds.


rbuffat:
@grischard @Klumbumbus I remove the defunct itu-lotes source: wrong coordinate projection, deres-ext:TRECHOINFO layer is not advertised by wms server. There is a PR open to update this source: https://github.com/osmlab/editor-layer-index/pull/928

PR #991, opened 2020-11-23
Filter deprecated EPSG codes in wms_sync

rbuffat:
JOSM seems to ignore deprecated EPSG codes, let's do the same here.


grischard:
Excellent!

PR #992, opened 2020-11-25
imagico.de osmim 2020-02 update

imagico:
* adding various new images for the Antarctic (many of them have a variable max_zoom because of the variable scale of the mercator projection).
* changing category of older images from `photo` to `historicphoto`.


rbuffat:
Thanks!

PR #996, opened 2020-11-29
europe / ch / Kanton Zürich: add 2017 elevation

rbuffat:
The Kanton Zürich added new elevation layers to his opendata elevation WMS server: https://opendata.swiss/de/dataset/wms-digitales-hohenmodell-ogd1

The data is open data and there no requirement to give attribution. The data for these new layers is not from the Kanton Zürich itself, but from the Federal Office of Topography: "Befliegung Bund".

@simonpoole 
Based on http://lists.openstreetmap.ch/pipermail/talk-ch/2020-August/010699.html I assume the canton is allowed to offer this data as opendata.


simonpoole:
> @simonpoole
> Based on http://lists.openstreetmap.ch/pipermail/talk-ch/2020-August/010699.html I assume the canton is allowed to offer this data as opendata.

That would be my understanding too.

PR #999, opened 2020-12-01
Check if source intersect country of country_code

rbuffat:
Currently, there is no test for the country_code attribute. 

This PR adds a check that the source geometry intersects the geometry of the country with the same ISO3166-1 country code. This check requires a dataset with the country geometries. "scripts/data/simplify_geometries.py" is a script to extract a dataset out of downloads from https://osm-boundaries.com. This dataset currently does not include countries for "AQ", "ZZ", "PS", "XN", "EU", "BV". While this dataset is based on osm data, I'm not sure if osm-boundaries.com data can be used. If not, https://github.com/ideditor/country-coder/blob/master/src/data/borders.json might be an alternative. 

```
License
Unless otherwise stated, Ground Zero Communications AB and/or its licensors own the intellectual property rights for all material on OSM-Boundaries. All intellectual property rights are reserved. You may access this from OSM-Boundaries for your own personal use subjected to restrictions set in these terms and conditions.

You must not:

Republish material from OSM-Boundaries
Sell, rent or sub-license material from OSM-Boundaries
Reproduce, duplicate or copy material from OSM-Boundaries
Redistribute content from OSM-Boundaries
```

### DIFF
--- a/scripts/check.py
+++ b/scripts/check.py
@@ -18,14 +18,14 @@ find sources -name \*.geojson | xargs python scripts/check.py -vv
 
 """
 
-import json
 import io
-from argparse import ArgumentParser
-from jsonschema import ValidationError, RefResolver, Draft4Validator
-import colorlog
+import json
 import os
-
-from shapely.geometry import shape
+from argparse import ArgumentParser
+import colorlog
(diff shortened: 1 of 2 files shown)


rbuffat:
Let's close this now and eventually look at this later.

PR #1002, opened 2020-12-05
TNRIS link updates

TXStormchaser:
The old TNRIS webserver will shut down in January, 2021. I have updates the Texas links to the new imagery server.


rbuffat:
@TXStormchaser Thanks for the PR! Is there a reason that you switch the format from jpeg to png? Typically, jpeg is preferred because of the smaller file size. 

The checks for the license URL fail as the server returns HTTP 404 codes despite serving the site's content.

TXStormchaser:
@rbuffat I had originally changed these in OSM, then copied the links over here. I normally use PNGs, so it defaulted to that. I can change those if JPGs are preferred.

I don't know why those URLs failed. I just hand-checked them all, and they all connected correctly, albeit slowly.

rbuffat:
@TXStormchaser that would be great! 

I also don't know why the server does that. It also happens in the browser.
tnris

sources/north-america/us/tx/Smith_County_Orthophoto_2019_WMS.geojson
```diff
@@ -4,10 +4,10 @@
         "license_url": "https://data.tnris.org/collection/38f1b2e7-6fc7-4357-80a4-1ee33d7abaaf",
         "id": "smith_county_2019_wms",
         "type": "wms",
-        "name": "Smith County Imagery 2019",
+        "name": "TX: Smith County Imagery 2020",
```
rbuffat: https://data.tnris.org/collection/38f1b2e7-6fc7-4357-80a4-1ee33d7abaaf names the dataset  "Smith County Imagery 2019", but this could be an error on this site. In the description, it says `The Smith County GIS Consortium consists of the Smith County 911 Communications District, City of Tyler, Smith County Appraisal District, and Smith County. This 6-inch resolution leaf-off aerial imagery was acquired in three missions all flown on December 11, 2019. ` Although the WMS URL includes SmithCo_2020_NCCIR_6in.

TXStormchaser: I noticed that, too. The data was taken in 2019, but wasn't posted until 2020. I went with the posting date instead of the collection date, since it was within a month of 2020. I'll change it to better reflect collection date.

rbuffat:
@TXStormchaser Thanks already for all the changes! One last question: you added a "TX:" in front of every name. Is there a particular reason for that? I don't have an opinion on this, I'm just curious. 

The only implication I can think of is that the name is typically written in the imagery_used tag in the changeset (I think at least JOSM and iD do it that way). Thus keeping the name stable helps to later identify a source used for a changeset.

TXStormchaser:
I did that to get the state name at the beginning for better sorting, to keep the Texas sources together. I also did that to differentiate between similar locations. A good example is the Smith County dataset. There's 5 Smith Counties in the US, so adding the TX to the start was a way to at least identify which state that Smith County data belonged to.

rbuffat:
@TXStormchaser I see. Thanks for all the work!

PR #578, opened 2018-10-28
Add orthophotos for Kalmar municipality

NKAmapper:
Three versions needed to cover the area.
Permission included as a pdf.

PR #1005, opened 2020-12-12
Fix europe / se / kalmar sources, add 2020 source

rbuffat:
* Fix broken europe / se / kalmar sources thanks to @NKAmapper: https://github.com/osmlab/editor-layer-index/pull/578#issuecomment-741721496
* Add additional source for 2020. 

There is a pdf in Swedish regarding the license/permission to use, but I don't understand Swedish: https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/se/Kalmar_permission.pdf


rbuffat:
Based on the Kalmar entry in
https://wiki.openstreetmap.org/wiki/Contributors#Sweden 
and https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/se/Kalmar_permission.pdf

Issue #894, opened 2020-08-21
Travis @traviscibot deploy undoes recent commits 


rbuffat:
As reported by @Klumbumbus in https://github.com/osmlab/editor-layer-index/pull/866, the travisbot undoes commits when a job on travis is canceled. 
It happened again in https://github.com/osmlab/editor-layer-index/commit/a54c7570b4708c0572eea377c399fca72182e9b2

@grischard could you switch off auto cancellation in travis?
https://docs.travis-ci.com/user/customizing-the-build/#building-only-the-latest-commit

**Until this is configured, it is important that only one PR at a time is merged, each after all jobs on https://travis-ci.org/github/osmlab/editor-layer-index/builds have finished.** 

For long-term it would be good to separate the deploy branch:  e.g. having a trunk/main/sources branch, where stuff is committed to, and gh-pages, where it is deployed to.

andrewharvey:
yikes, this happen before too.

I'm all for having a separate deployment branch. In my view dist files shouldn't be committed, at least not in the same branch as development work happens.

There was some work before to stop using the gh-pages branch, I'd be happy to see us move to master/main branch as the default for sources and then a dist branch for publishing the outputs.

rbuffat:
> There was some work before to stop using the gh-pages branch, I'd be happy to see us move to master/main branch as the default for sources and then a dist branch for publishing the outputs.

This is probably quite a complex task.  This is a tool that might be helpful: https://github.com/sgillies/mainer But I never used it before, thus I don't know how good it works.

grischard:
Editors used to have the gh-pages branch url hardcoded, so that was a no-no. Now that the most recent versions of everything but iD have switched…

grischard:
It looks like I don't have the permissions to change the travis settings.

simonpoole:
The actual proper solution is to create a release automatically, using a branch is just kicking the can a bit down the road. Apps can easily retrieve the latest release (see for example https://github.com/simonpoole/beautified-JOSM-preset/releases where I generate a github release from the build script, and, for example, https://github.com/simonpoole/beautified-JOSM-preset/releases/latest/download/vespucci_icons.zip gets the latest release of that specific file).

The main reason why this hasn't happened here is that the output file location tends to be hardwired in numerous places, so a move to a different solution requires some kind of transition solution.

a) deciding how often or / and on what kind of event a new release is authored and the versioning convention

b) providing updated files at the historic locations during transition (a year or so).

[...]

PR #1006, opened 2020-12-12
Switch from travis to github workflows

### LINKED ISSUE #981: Travis sunsetting?
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

We will get 1000 free minutes before we have to do something.

rbuffat:
This PR switches the CI from travis to github actions and eventually fixes https://github.com/osmlab/editor-layer-index/issues/981

- [x] Add github workflow with strict checks for pull requests
- [x] Add github workflow that runs basic checks, exports imagery.* files and deploy to gh-pages when something is pushed to the repository
- [x] Validate imagery.xml against the maps.xsd schema from JOSM
- [x] Missing: migrate i18n script / Transifex
- [x] Missing: Completly eliminate the need for shell commands (e.g. find). This should enhance cross-platform compatibility.

One important change is that with this PR there will be two branches required. E.g. a "develop" or "main" branch, where new changes are pushed and "gh-pages" (which should pretty much correspond to the current gh-pages branches without the scripts). The workflow then generates updated imagery* files and pushes them to gh-pages ([as well as all sources, site and *md files](https://github.com/osmlab/editor-layer-index/blob/b2a4911ccae1fbf3cd3d35a17ff6ff0cea474dd8/.github/workflows/push.yml#L47-L56)). This solves https://github.com/osmlab/editor-layer-index/issues/894 and should be backward compatible, as hardcoded URL will still work.


rbuffat:
The PR is ready to be reviewed. If everything is fine it is probably easier to directly push this branch to osmlab/editor-layer-index instead of merging it into gh-pages and then creating a new branch. 

The WMS sync bot only works on the default branch set in the github UI: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch
I don't have permission to change this. Although as long as also the scripts are "deployed" into gh-pages, it should be possible to change the target branch of the PRs created by the bot manually afterward. 

To name the second branch `develop` is just a suggestion. I'm not good at naming things, thus it would be appreciated if somebody else has an idea.

grischard:
Wow. This is an impressive and huge PR, and I think it might take a while to review.

You're changing quite a number of things unrelated to switching from Travis to github workflows.

The gh-pages deploy needs to include all the files that are statically hosted on github, like icons. It could be easiest to just dump everything across. The thought behind having just one branch was that it would be more newbie-friendly. Using the string 'gh-pages' was fossilised for certain editors; I'm not sure if that's still the case.

You've changed the check scripts to run on directories instead of files. Can a check still run on a file?

rbuffat:
> You're changing quite a number of things unrelated to switching from Travis to github workflows.

You are right, that's true. One thing lead to another... I will create a new PR with just the github actions and as less as possible other things, probably later this week.

grischard:
Heh, I know how that goes :). If you've tested it and are happy with how it works and solves the two things I've noticed, I don't want to be the cause of extra effort.

rbuffat:
@grischard  sorry, it took a bit longer than I hoped for. I think it is always very valuable if somebody else can look at the changes before something is merged.  I will look into creating new PRs for the other changes after the github actions are finished. This is the most urgent issue at the moment.

PR #1012, opened 2020-12-23
Switch from travis to github workflows

rbuffat:
This PR is a follow up to https://github.com/osmlab/editor-layer-index/pull/1006 and contains only the github workflows. 

This PR is for review only. Once it is checked I will push a new branch to the repository. The name of this new branch (which should then be the target for all following PRs) could e.g. be: develop, maint, main or something else. I don't have a preference here. 

Separating the deploy and gh-pages branches have the advantage that there are no race conditions between pushes and deploys. As we deploy the whole new branch to gh-pages, there should be no changes for people using hard-coded URLs.

As github actions using cron only operate on the default branch, the github default branch needs to be set to the new branch so that the WMS sync bot works: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch (the bot should work regardless, but the target of each PR needs to be switched to the new branch otherwise both branches can get out of sync.)

An alternative to two branches would be if it would be possible to limit the deploy to the  imagery* and en.yaml files. Not sure if this is possible though.


rbuffat:
@grischard It seems as deploying to gh-pages works, but I need to test this a bit more. Hopefully, I will have time for that in the next few days.

rbuffat:
@grischard It looks as travis is already not working anymore in https://github.com/osmlab/editor-layer-index/pull/1014, thus migration is a bit urgent now.

I tested the actions on my eli fork for the gh-pages branch and it seemed to work. My fear is tough that through some hiccup a deploy could erase all files other than the newly created (https://github.com/rbuffat/editor-layer-index/blob/30c8cd7f19b7e3d4904e74dd32113460b8b00b58/.github/workflows/push.yml#L38-L51) 
However, this did not happen yet. I created a backup of the current gh-pages branch so that in case this would happen there is a backup https://github.com/rbuffat/editor-layer-index/tree/upstream_backup

grischard:
Hmm. Well, if worst comes to worst, we roll back the changes, no?

iD bundles the geojson separately, and as far as I know, the other editors have switched to JOSM.

rbuffat:
@grischard yes, I suppose so

rbuffat:
@grischard It would be good to merge the current gh-pages in all open pull requests. By any chance, do you know how to do this?

grischard:
@rbuffat I think it can only be done manually, by clicking 'update the base branch' near the bottom. This only works if the person has allowed the maintainers to make changes.

PR #1021, opened 2021-01-11
WMS Sync: check if projections return an image

rbuffat:
Some WMS servers advertise projections but do not actually support them. This PR sends a GetMap request for each projection and checks if an image is returned. 

The updated script modifies locally the following sources, but there are always some server/network issues, thus more WMS servers could be affected. 

```
        modified:   sources/europe/ee/EstoniaBasemap-Maaamet.geojson
        modified:   sources/europe/ee/EstoniaCIR-Maaamet.geojson
        modified:   sources/europe/ee/EstoniaCadastre(Maaamet).geojson
        modified:   sources/europe/it/PCN2012-Italy.geojson
        modified:   sources/europe/no/KartverketAviationObstructionsoverlay.geojson
        modified:   sources/europe/no/KartverketContourLinesoverlay.geojson
        modified:   sources/europe/se/SollentunaOrthophoto.geojson
        modified:   sources/north-america/ca/Canvec-French.geojson
        modified:   sources/north-america/ca/Canvec.geojson
        modified:   sources/north-america/us/USDA-NAIP.geojson
```

### DIFF
--- a/requirements-wms_sync.txt
+++ b/requirements-wms_sync.txt
@@ -1,4 +1,4 @@
-pillow==8.0.1
+pillow==8.1.0
 aiohttp==3.7.3
 aiofiles==0.6.0
 imagehash==4.2.0
@@ -6,5 +6,6 @@ mercantile==1.1.6
 shapely==1.7.1
 pyproj==3.0.0.post1
 jsonschema==3.2.0
-colorlog==4.6.2
+colorlog==4.7.2
 pyproj==3.0.0.post1
+python-magic==0.4.18
--- a/scripts/sync_wms.py
+++ b/scripts/sync_wms.py
@@ -21,6 +21,8 @@ from pyproj.crs import CRS
 import aiofiles
 from aiohttp import ClientSession
 from shapely.ops import cascaded_union
+import magic
+
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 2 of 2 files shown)


rbuffat:
@grischard @don-vip I relaxed the condition for when it is checked for EPSG:3857 and EPSG:4326 despite them not advertised by the GetCapabilities request. Previously only equal ImageHashes were accepted, now there is a small difference between the images allowed. This tolerates some difference possibly due to different reprojection/compression. 

I tested all sources where this leads to modifications and all of these sources looked good. But I will review the next PR of the WMS bot again once this is merged. 

The log message to look out looks like this:
```
INFO:root:sources/europe/ch/Kanton_Schaffhausen_ortho_2013.geojson: Add EPSG:3857 despite not being advertised  : 7874c2df8982435c - 7874c2df8980035c: 2
        Image URL: https://wms.geo.sh.ch/raster?LAYERS=Luftbild_2013&STYLES=default&FORMAT=image/jpeg&CRS=EPSG:4326&WIDTH=512&HEIGHT=512&BBOX=47.694974341862824,8.63525390625,47.709761542666385,8.6572265625&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap
        ImageHash: 7874c2df8980035c
        Image URL: https://wms.geo.sh.ch/raster?LAYERS=Luftbild_2013&STYLES=default&FORMAT=image/jpeg&CRS=EPSG:3857&WIDTH=512&HEIGHT=512&BBOX=961272.0677143765,6056258.625091084,963718.0526195022,6058704.609996207&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap
        ImageHash: 7874c2df8982435c
```

Ideally, we should annotate the geojson files with all things that are different from the corresponding GetCapabilities request (Removed projections, respectively added projections). This would help to track the changes afterward. I'm not yet sure what the best way would be. I think JSON does not support comments.

PR #1038, opened 2021-01-31
Fix sources with empty wms bot test image

rbuffat:
Fix sources from https://github.com/osmlab/editor-layer-index/pull/1031#issuecomment-770284681 with `has category * but image hash is 0000000000000000`.

- GliwiceOrthophotomap* sources had the wrong min_zoom set. 
- Rio_2015.geojson wrong max_zoom and not ideal geometry.


rbuffat:
The workflow has troubles with the `(` in the filenames. See https://github.com/osmlab/editor-layer-index/issues/1039
As the changes are of minor nature I will merge despite failing tests.

PR #1042, opened 2021-02-02
Fix min_zoom for sources/north-america

rbuffat:


imagico:
Could you explain what the source of these min_zoom values is?  In 

https://raw.githubusercontent.com/osmlab/editor-layer-index/4765cdbf060a9008ca3732de32972397dcc6d954/sources/north-america/us/osmim-imagicode-LC80770232017156LGN00.geojson

for example you add a min_zoom value of 5 while the layer returns perfectly valid tiles at lower zoom levels:

https://www.imagico.de/map/osmim_tiles.php?z=3&x=0&y=5&layer=LC80770232017156LGN00

For better understanding:  The OSMIM tile serving script has no lower zoom level limit.  For empty tiles it returns a 302 redirect to an empty PNG at the lower zoom levels (that includes those tiles within the layer bounds but where the image coverage is too small to produce a non-empty tile) and a 404 at the higher zoom levels.

rbuffat:
> Could you explain what the source of these min_zoom values is?

@imagico I would say to a degree purely subjective judgment. It is the first zoom level where the[ image hash ](http://www.hackerfactor.com/blog/index.php?/archives/432-Looks-Like-It.html) is not the same as an all-white image. 

osmim-imagicode-LC80770232017156LGN00

I can't say what a reasonable definition for min_zoom is that works for all cases. 
* The first zoom level an image is returned?
* The first zoom level a nonuniform image is returned?
* The first zoom level where at least X pixels are not the background color?
 
In practice, for sources with "small" geometries, the first zoom levels are probably not the most relevant ones.

What is your opinion? 

Here some more examples:

JapanGSIairphotoImagery
USGSImagery

imagico:
Without knowing what the practical purpose of the min_zoom value is i cannot form an opinion.  By concern is future updates of the layer information.  If there are subjectively chosen min_zoom values added i can no more automatically update the layer metadata and would need to move to just adding new layers instead of updating everything.

rbuffat:
@imagico If you prefer I can revert the changes for all osmim sources.

PR #1045, opened 2021-02-03
stadtplan does not exist anymore

datendelphin:


rbuffat:
The source seems still to work, do you know more about it?

datendelphin:
I just went through the list of layers on mapproxy.osm.ch in preparation to migrate it to a new server. While the proxy cache is still there, the underlying wms is not reachable anymore. It looks like geo basiskarte zuerich might be an alternative, but that wms supports more projections so no need for the sosm mapproxy. Though this file could be recycled maybe, not sure of the extent.

rbuffat:
@datendelphin that explains it. merci!

rbuffat:
@datendelphin 

https://raw.githubusercontent.com/osmlab/editor-layer-index/gh-pages/sources/europe/ch/Stadt_Bern_10cm-2016-TMS.geojson

and 

https://raw.githubusercontent.com/osmlab/editor-layer-index/gh-pages/sources/europe/ch/Stadt_Bern-10cm25cm-2012.geojson

seem to be not working anymore. They are both also available from a WMS server: https://www.bern.ch/open-government-data-ogd/ogd-nach-themen/geodaten-ogd

Should we change them from the sosm mapproxy to the WMS server?

datendelphin:
Ah missed that one in ELI. Yes the wms server now supports the required projection so there is no need for the mapproxy anymore. If it does not work, we can add it back to the mapproxy of course

PR #1049, opened 2021-02-07
Create TIGER 2020 roads layer.

iandees:
Adds the TIGER 2020 roads layer based on the recently-released TIGER 2020 data.


rbuffat:
@iandees Is there somewhere a website indicating the license of the data?

iandees:
No. It's OSM data and US Census data, which is public domain (doesn't have a license).

rbuffat:
I think it would be nice if every source includes enough information so that somebody not familiar with the situation can find out about the license the data was published (e.g. if there is an official homepage that states that the data is public domain or an osm wiki entry that describes the situation). 

Maybe https://wiki.openstreetmap.org/wiki/TIGER?

PR #923, opened 2020-09-11
Add Kanton Basel Stadt 2017

rbuffat:
This source is listed in JOSM but not here.

"This imagery is provided via a proxy operated by https://sosm.ch/"

PR #1052, opened 2021-02-07
Add Kanton Basel Stadt 2020 imagery

rbuffat:
There is new imagery for 2020 on the WMS servers of Basel Stadt: https://www.geocat.ch/geonetwork/srv/ger/md.viewer#/full_view/7f23054c-4979-45be-be0c-d89f9a9c2756
respectively https://opendata.swiss/de/dataset/true-orthofoto-2020-marz
The license is probably  CC BY 3.0 CH: https://www.geo.bs.ch/nutzung/nutzungsbedingungen.html (I'm not 100% sure which category the data belongs to, but I guess it is the same as the 2017 imagery). 

@simonpoole Could you have a look at this? Is this dataset also covered by the waiver you obtained in https://github.com/osmlab/editor-layer-index/pull/923#issuecomment-766756894 ?


simonpoole:
Quoting from the waiver

> ...with respect to any imagery Kanton Ba- sel-Stadt has released or releases in the future under a CC BY 3.0 license, Kanton Basel-Stadt waives any and all conditions of the license that could conflict with or prohibit free and unencum- bered use of that imagery to derive or extract data,...

So yes, it covers this imagery too. My bad that I haven't put the waiver anywhere reasonable yet, I'll get that done once I'm back from vacation.

rbuffat:
@simonpoole  Thanks!

PR #1054, opened 2021-02-12
Add Berlin-2020-TrueDOP.geojson most recent

polarbearing:


rbuffat:
@polarbearing Thanks for this pull request. I would propose https://codefor.de/datenschutz/ as privacy policy url as the url points https://tiles.codefor.de/berlin-2020-truedop.... , thus the tiles seem to be hosted by codefor.de and not a berlin.de domain.

I assume the license of the data is „Datenlizenz Deutschland – Namensnennung – Version 2.0", same as the previous Berlin datasets, and there is explicit permission for Berlin https://wiki.openstreetmap.org/wiki/File:2019-06-03_Datenlizenz_Deutschland_Berlin_OSM.pdf

You can also remove the "best": true flag from https://raw.githubusercontent.com/osmlab/editor-layer-index/gh-pages/sources/europe/de/Berlinaerialphotograph2019.geojson

polarbearing:
@rbuffat -  codefor.de (Open Knowledge Foundation) is mirroring the data from fisbroker (fbinter.stadt-berlin.de) (official governmental geo portal). In the JOSM list, we have both URLs for stadt-berlin.de and for codefor.de as mirror, is that technically possible here also?

Thus as stadt-berlin.de has ordered the images being taken, processed them and is the license holder, it is my belief that they should be responsible for the privacy side as well?

Your assumption about the DL-DE-2.0 is correct, it has an attribution flaw for OSM which needs to be fixed by a waiver, which for Berlin is in the linked PDF. 

As for the "best" flag, yes, if it does not hurt to have the flag twice for a few days, my plan was to go through the other Berlin files the next days, and update the license link (which changed last year for all layers), improve the naming of the layers and fix the 'true' and 'historic' flags. Is that ok?

rbuffat:
> @rbuffat - codefor.de (Open Knowledge Foundation) is mirroring the data from fisbroker (fbinter.stadt-berlin.de) (official governmental geo portal). In the JOSM list, we have both URLs for stadt-berlin.de and for codefor.de as mirror, is that technically possible here also?

There is currently no possibility to mark mirrors, although it would be possible to just add a second geojson. But I would suggest keeping only one source. 


> Thus as stadt-berlin.de has ordered the images being taken, processed them and is the license holder, it is my belief that they should be responsible for the privacy side as well?

The privacy policy typically contains information about how personal data is saved (e.g. cookies, IP addresses, etc). If the server containing the tiles is operated by codefor.de typically their privacy policy applies. 

> Your assumption about the DL-DE-2.0 is correct, it has an attribution flaw for OSM which needs to be fixed by a waiver, which for Berlin is in the linked PDF.
> 
> As for the "best" flag, yes, if it does not hurt to have the flag twice for a few days, my plan was to go through the other Berlin files the next days, and update the license link (which changed last year for all layers), improve the naming of the layers and fix the 'true' and 'historic' flags. Is that ok?

No worries, should we leave this PR open during this time or would you prefer creating a second PR? Feel free to ask any questions if they arise.

polarbearing:
Ok, I have changed the privacy link to codefor.de, and also added them in the attribution line so they are recognised as mirror. We'll stick to the codefor.de then in this index. 

I prefer this PR to be merged and will make a separate PR for updating the other files, being different purposes.

polarbearing:
followup in #1056

rbuffat:
Thanks! @polarbearing

PR #1063, opened 2021-02-22
Poland geoportal update

Asteliks:
This PR solves issue https://github.com/osmlab/editor-layer-index/issues/1060 and adds three new sources form Geoportal 2:

- building layouts map,
- street names,
- address points

Issue #1039, opened 2021-01-31
Pull request / strict check workflow has trouble with some characters


rbuffat:
The current workflow to check new files has some problems if e.g. a filename contains a `(` or `)`. These characters should be escaped before the filenames are used.

E.g. 
```
python scripts/strict_check.py sources/europe/pl/GliwiceOrthophotomap2013(aerialimage).geojson sources/europe/pl/GliwiceOrthophotomap2017(aerialimage).geojson sources/europe/pl/GliwiceOrthophotomap2018(aerialimage).geojson sources/europe/pl/GliwiceOrthophotomap2020(aerialimage).geojson sources/south-america/br/Rio_2015.geojson
```

PR #1064, opened 2021-02-22
Poland: Outdated maps removal

Asteliks:
This PR is a result of months of checking which sources are still available and which we can use. In general the story goes as follows:  the data has been moved to the national system thus we no longer maintain it. Based on that I request that first the missing layers should be added from the national system (https://github.com/osmlab/editor-layer-index/pull/1063) and only then this PR can be merged as a form of clean up. As for the other sources highlighted by the [watchdog](https://rbuffat.github.io/eli_watchdog/) these still work / can be updated to work.


Asteliks:
@rbuffat  check failed probably due to: https://github.com/osmlab/editor-layer-index/issues/1039?

rbuffat:
Yes I think this is the reason that the tests fail. I hope that I will have time too look at it in the next days.

rbuffat:
@Asteliks I hope I found a fix for https://github.com/osmlab/editor-layer-index/issues/1039. Would it be possible to rebase / merge the current gh-pages in your branch to test it?

Asteliks:
@rbuffat seems to be working here

rbuffat:
@Asteliks Thanks for making the effort and going through all the sources!

PR #1075, opened 2021-03-06
Add building overlay for Norway

NKAmapper:
[JOSM entry](https://josm.openstreetmap.de/wiki/Maps/Norway#KartverketBuildingsoverlay)

[Permission](https://forum.openstreetmap.org/viewtopic.php?id=71994)

[Data protection policy](https://www.kartverket.no/en/about-kartverket/data-protection-policy)


rbuffat:
Thanks @NKAmapper !

Issue #1071, opened 2021-03-01
Swisstopo data


paulw1010:
Today the Federal Office of Topography (Swisstopo) in Switzerland released their data for open access use. ~~Many layers are now available for use in OSM~~ (**edit: discussions on use of data still ongoing**) including the national map, hi-res imagery, and hillshade. They cover both Switzerland and Liechtenstein.

## URL of the imagery
https://wms.geo.admin.ch/

## Two letter country code
CH and LI

## Desired layers

Here are some of the layers most useful for OSM. Others can be found [here](https://map.geo.admin.ch/). To make a WMS link, use the following format 

`https://wms.geo.admin.ch/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=<YOUR LAYER HERE>&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}`

- Swisstopo national map (1:10,000) ([WMS link](https://wms.geo.admin.ch/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=ch.swisstopo.landeskarte-farbe-10&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}))

- Swissimage (most recent) ([WMS link](https://wms.geo.admin.ch/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=ch.swisstopo.swissimage&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}))

- SwissALTI3D Multidirectional Hillshade ([WMS link](https://wms.geo.admin.ch/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=ch.swisstopo.swissalti3d-reliefschattierung&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}))

## License

No CC license or anything, just this disclaimer...

> The conditions of use allow free use for all purposes and oblige the user to indicate the source as “Source: Federal Office of Topography swisstopo” or “© swisstopo”.

[...]

PR #1080, opened 2021-03-07
Add europe/ch swisstopo swissimage + elevation hillshade

rbuffat:
Based on the feedback provided by @paulw1010 and @simonpoole in https://github.com/osmlab/editor-layer-index/issues/1071  I created a PR for the swissimage / swissalti and swisssurface datasets. 

Issues:
* At the moment the country code is only set to CH, but the data is also available for Lichtenstein.
* The swissSurface3D layer currently contains the geometry for Switzerland + Lichtenstein, but currently it is only available for the west and east. But in the next 1-2 years it will be available for the whole country. 

@simonpoole  Would it be good to have an OSM wiki page to track the current status of what swisstopo datasets can be used?


simonpoole:
@rbuffat I'm preparing another post on the matter with input from swisstopo, as soon as I have a sentence or two that can be quoted from swisstopo I'll add it to the wiki and give the go ahead here.

simonpoole:
See https://wiki.openstreetmap.org/wiki/Swisstopo_data_use good to go from my pov.

rbuffat:
Thanks @simonpoole @paulw1010!

PR #1101, opened 2021-03-27
Added Echtfarbenbild VoGIS2020

ghaefele:
Current photo for Vorarlberg (Austria),


rbuffat:
@ghaefele Thanks for this PR! 

It looks like the VoGIS data are now under the CC-BY 4.0 license. There are some compatibility issues with this license, which require explicit permission: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

There is a template available to obtain permission: https://drive.google.com/file/d/0B3PN5zfbzThqeTdWR1l3SzJVcTg/view

Would you be willing to contact the data owner to obtain permission? (= Asking the data owner by email if they would be willing to fill out the template)

ghaefele:
I got the permission from the data owner.
[VoGIS-Web-Services.pdf](https://github.com/osmlab/editor-layer-index/files/6270266/VoGIS-Web-Services.pdf)

rbuffat:
super!  Thanks for asking!

PR #1110, opened 2021-03-29
Fix dates of Frankfurt am Main Luftbild 2017

xuiqzy:
The start_date and end_date was the same as that of the Frankfurt am Main Luftbild 2016, but it really should have the dates mentioned here for the 2017 version: https://www.offenedaten.frankfurt.de/dataset/wms-luftbild-2017
Note the dates are in DD.MM.YYYY format in German.

The dates in the description key in the geojson are already correct.


rbuffat:
Thanks @xuiqzy !

PR #890, opened 2020-08-17
Fix broken "Hamburg (20 cm)" source

rbuffat:
The WMS server has changed and thus is currently broken: https://geoportal-hamburg.de/info/DOP_GBK_Anpassungen.pdf

> Digitale Orthophotos Hamburg
> WMS Digitale Orthophotos Hamburg
> Name alt: WMS Digitale Orthophotos 20cm Hamburg
> URL alt: https://geodienste.hamburg.de/HH_WMS_DOP20
> URL neu: https://geodienste.hamburg.de/HH_WMS_DOP

The layer did not have any license URL before. The [WMS server](https://geodienste.hamburg.de/HH_WMS_DOP?REQUEST=GetCapabilities&SERVICE=WMS) includes:

```
<Fees>Nutzungsbedingungen: Datenlizenz Deutschland - Namensnennung - Version 2.0; &lt;a href="https://www.govdata.de/dl-de/by-2-0"&gt;https://www.govdata.de/dl-de/by-2-0&lt;/a&gt;; dl-de-by-2.0; Namensnennung: "Freie und Hansestadt Hamburg, Landesbetrieb Geoinformation und Vermessung"</Fees>
<AccessConstraints>Es gelten keine Zugriffsbeschränkungen,Hamburgisches Geodateninfrastrukturgesetz (HmbGDIG)</AccessConstraints>
```

I found the following page: https://metaver.de/trefferanzeige?cmd=doShowDocument&docuuid=B2013F67-EB8A-4629-8E6D-55A5F74CE129&plugid=/ingrid-group:dsc-MV
-> Verfügbarkeit 

> Zugriffsbeschränkungen
> Es gelten keine Zugriffsbeschränkungen
> 
> Nutzungsbedingungen
>  Datenlizenz Deutschland Namensnennung 2.0
> 
> Quellenvermerk: Freie und Hansestadt Hamburg, Landesbetrieb Geoinformation und Vermessung

The link to the WMS server can be found under "Verweise" -> "Weitere Verweise" -> "Dienst "WMS Digitale Orthophotos Hamburg" (GetCapabilities)"

@Klumbumbus would it be possible that you have a look at the license?

PR #1114, opened 2021-04-09
Add DOP20 and DK5 layers for Hamburg, Germany

hauke96:
This PR re-adds the two layers for Hamburg, Germany as there are new information regarding the legal situation.

There was some discussion going on in PR #890 and so the community in hamburg asked the LGV (Landesbetrieb Geoinformation und Vermessung) for permission, which is the official owner and creator of the data. The answer was positive and we are allowed to use the data as long as we make clear what the sources are by adding a reference to the LGV. This can be done via changeset metadata with e.g. "HH LGV DOP20 2021" or similar.

I just copied the geojson data from the #890 PR, added the LGV reference in the name and fixed the URL of the "Hamburg 20cm" layer.

The complete mail transcript with the LGV (German only): https://wiki.openstreetmap.org/wiki/DE:Permissions/LGV_Hamburg_2021
The (moderated and not yet up to date) discussion in the forum: https://forum.openstreetmap.org/viewtopic.php?id=19227&p=2


rbuffat:
@hauke96 Thank you for contacting the LGV, this is much appreciated!

rbuffat:
Just a note:  as far as I'm aware the way OSM attributes data is using the Contributors Wiki page: https://wiki.openstreetmap.org/wiki/Contributors
The changeset comments are intended for mappers and not for attribution. 

See the discussion in JOSM and iD:
https://josm.openstreetmap.de/ticket/19911
https://github.com/openstreetmap/iD/issues/6300

PR #1119, opened 2021-04-15
Created md (Moldova) folder, added imageries and their properties

Oni-DOS:
These are the basical imagery used by Moldovan OSM Community. The Moldovan Agency for Land Relations and Cadaster of the Republic of Moldova permit using these imagery in OSM and other non-profit projects. They work in accordance with the law on the national spatial data infrastructure ( https://www.legis.md/cautare/getResults?doc_id=96633 ) and with the Regulation on the rules for the sharing of spatial data sets and related services between public entities and third parties ( https://www.legis.md/cautare/getResults?doc_id=102213 )


grischard:
First of all, congratulations to the Moldova community! It's always special when you get to create your folder in ELI :). I also see that this is your first pull request ever, so double congratulations.

The `:` in your source id is causing the CI to fail. The acceptable regex pattern is `^[-_.A-Za-z0-9]+$`.

Oni-DOS:
Hello! And thanks for your respond! I fixed the regex errors but I don't understand what is "ERROR:root:Exception: 'NoneType' object has no attribute 'findall'".

rbuffat:
> Hello! And thanks for your respond! I fixed the regex errors but I don't understand what is "ERROR:root:Exception: 'NoneType' object has no attribute 'findall'".

@Oni-DOS The error comes from the library checking WMTS sources. I agree this is not a helpful error message.  But I guess you meant to specify the type as `"type": "wms"`and not `"type": "wmts"` for Ortophoto_2016_RM_Reserve-Server.geojson.

sources/europe/md/Ortophoto_2016_RM_Reserve-Server.geojson
```diff
@@ -0,0 +1,1741 @@
+{
+    "type": "Feature",
+    "properties": {
+        "description": "2016 Ortophotos, Moldova, Scale 1:5000, 20cm, true colours, covers practically the entire territory of the Republic of Moldova, except some areas from the east part of Moldova. [Geoportal.md server]",
+        "id": "Geoportal-Ortofoto_2016_RM",
+        "name": "Moldova Ortophoto (2016) / Geoportal.md Reserve Server",
+        "type": "wmts",
```
rbuffat: I assume this should be `"type": "wms"`

Oni-DOS: Ah! Yeah, that's true, here should to be wms.

Oni-DOS: Thanks for the help!

grischard:
LGTM. @rbuffat?

grischard:
No complaints means we can merge :). Congratulations again. The layers will be in the next version of iD. For the other editors, don't forget to add the layers to the JOSM wiki.

Oni-DOS:
Thanks! In next day I will try to add these layers and to JOSM wiki. In the next weeks I will add and others old imagery (from 2014, 2011, 2007 and 2006). Thanks again for the help!

rbuffat:
@grischard  I had not yet time to look at the license terms, but you have way more experience than me, so you don't have to wait for my reaction!

PR #1121, opened 2021-04-20
Update PNOA Spain TMS to use JPEG format

jsanz:
I'm submitting this change in the format of the images requested to the PNOA TMS service for Spain as a request by the Spanish NMA (IGN/CNIG). `image/jpeg` requests are cached by their CDN as opposed to the current `image/png`, making the editing experience slower for everyone.

### DIFF
--- a/sources/europe/es/PNOASpain-TMS.geojson
+++ b/sources/europe/es/PNOASpain-TMS.geojson
@@ -6,7 +6,7 @@
             "required": true
         },
         "name": "PNOA Spain",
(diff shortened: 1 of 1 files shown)


rbuffat:
The test is failing because there are issues with the SSL certificate for the privacy url, but the URL is reachable. We need to relax this test, as we can't influence the presence of correct SSL certificates anyway.

rbuffat:
Also, thanks @jsanz for your PR!

jsanz:
> The test is failing because there are issues with the SSL certificate for the privacy url, but the URL is reachable. We need to relax this test, as we can't influence the presence of correct SSL certificates anyway. 

I'll share this with them to see if they can do anything to fix it.

Thanks! 🙌

rbuffat:
There seem to be some issues with SSL: https://www.ssllabs.com/ssltest/analyze.html?d=www.idee.es, but Chrome does not show any. It could also be just a problem with the setup of the test script.

PR #1123, opened 2021-04-21
Update Stockholm orthophoto to 2018

NKAmapper:
Also adds privacy policy.
Syncs with JOSM - [link](https://josm.openstreetmap.de/wiki/Maps/Sweden#StockholmOrthophoto).


rbuffat:
All failing tests are due to SSL errors, but the URLs are reachable. We need to relax these tests. 

Thanks @NKAmapper for this PR!

rbuffat:
If the 2016 imagery is still accessible, would it make sense to keep it in the index for some time?

NKAmapper:
> If the 2016 imagery is still accessible, would it make sense to keep it in the index for some time?

I do not think so. I could have a look at one of the more historic ones later, for example 1995 is available.

rbuffat:
There seem to be some issues with SSL: https://www.ssllabs.com/ssltest/analyze.html?d=www.idee.es, but Chrome does not show any. It could also be just a problem with the setup of the test script.

PR #1126, opened 2021-04-27
Add Turku topographic map

crimsondusk:
Previously I added `turku-orto-2018-true.geojson`. Here is the topographic map from the same WMS source.


rbuffat:
Thanks  @crimsondusk!

PR #1130, opened 2021-05-06
update french sources

Olyon:
add source from https://josm.openstreetmap.de/wiki/Maps/France


rbuffat:
@Olyon Thanks for this PR? 

The PR includes a lot of changes, would it be possible to split it into smaller ones? This would help reviewing the individual changes. 

I had a quick look at the changes and I had the following thoughts:
* Some sources seem to be quite old (> 10 years). Since the `BDOrtho IGN` imagery is available now, are some of these imagery identical to the IGN imagery, or are outdated?
* There seems to be a osmbased map. The data listed in this index is mainly used for tracing. Thus, does this map contain useful information for tracers or is it just a different style of the default map?

Olyon:
I have divided into several PR. 
#1131 
#1132 
#1133
#1134 
#1135

PR #1131, opened 2021-05-07
update french sources : Ortho HR

Olyon:
I add or update the French ortho HR sources.
Ortho HR 2013 to 2020 brings together the orthophotos of French départements published under the French open license OL/LO hosted on wms.openstreetmap.fr.
Ortho HR are the youngest of these orthophotos in France.
Ortho HR is often the same as BDOrtho IGN but is sometimes newer. 
more details on wms.openstreetmap.fr


rbuffat:
@Olyon As far as I'm aware openstreetmap.fr does not has a privacy policy (cookies, saved data, etc.). Thus, it is ok when the tests fail for this.
 
Ideally, there would be a license URL set for each source that documents the license of the imagery. I'm not sure if it is possible to get this information for all data on [openstreetmap.fr](https://wiki.openstreetmap.org/wiki/FR:Serveurs/wms.openstreetmap.fr).

rbuffat:
Thanks @Olyon

PR #1135, opened 2021-05-07
old photos of Nancy

Olyon:
Old photos of the city of Nancy in France.


rbuffat:
I'm a bit hesitant to merge this. The imagery seems to be old 1994-2008 and there seems to be already at least 3 different more recent imagery sources available.

Is there a particular reason why this imagery is better than the existing ones?

cicku:
You should check out https://openhistoricalmap.org/.

OSM is simply not for old stuffs.

Olyon:
I wanted to add all the photos present in Josm but I agree that these orthophotos are old and do not bring much for the contributors.

rbuffat:
It is not black and white. It could be that old imagery offers some benefits such as less vegetation or a better resolution. But in this case, it looks as there is enough newer imagery available. Thus I will close this PR. If somebody does not agree please comment.

PR #1136, opened 2021-05-07
Add PA Orthoimagery

cicku:
[PEMA Ortho](https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=5158) is public domain imagery.

image


rbuffat:
@cicku The metadata page does not seem to list the license. Would it be possible for you to create a OSM Wiki page with the screenshot? Then we could link this as license_url. So we would be independent of if this Github PR discussion will still be accessible in the future.

cicku:
Do you have any sample wiki page for reference?

rbuffat:
@cicku It does not need to be something fancy. I would just upload the picture and put it together with the name of the dataset and the URL describing the data https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=5158. 

Here you can find examples of existing entries: https://github.com/osmlab/editor-layer-index/search?p=1&q=wiki.openstreetmap.org

cicku:
Done.

rbuffat:
@cicku Great! Thank you for all your effort!

PR #1140, opened 2021-05-12
Fix tms max_zoom for europe/fr

rbuffat:
This PR sets the max_zoom to the metadata value of the TileMap.


rbuffat:
@don-vip I implemented a test for min/max zoom using the MapTile metadata (https://github.com/osmlab/editor-layer-index/pull/1138).  Based on this test I changed these max_zoom levels for wms.openstreetmap.fr sources. I assume you have a lot more experience regarding min/max zoom tests. Do you have any thoughts about the dangers of testing for min/max zooms using this method? One thing I noticed is that wms.openstreetmap.fr throws HTTP 504 for small zoom levels (0-3).  Thus, the test emits only a warning if the zoom level mismatches for min zoom.

rbuffat:
ok, the last few commits should have been in a new branch...

don-vip:
It seems fine, good job 👍

rbuffat:
One possible issue I can think of if the highest zoom available level does not offer a better resolution some might prefer to use a lower max_zoom. Lets see if there will be any issues. If so, the test can be disabled again.

PR #1150, opened 2021-05-25
Updated imagery in Croatia

Janjko:
Removed 2017, 2018 and 2019 satellite imageries, and added their composites, 2017-2018 and 2019-2020.


rbuffat:
Thanks @Janjko !

PR #1175, opened 2021-06-13
Adding two new hillshade layers for Slovakia

Mashin6:


arch0345:
Looks like these layers have been added in #1996

PR #1180, opened 2021-06-25
Maryland LiDAR imagery

cicku:
LiDAR imagery to assist mapping in forest


cicku:
Sorry, I submitted the PR in my phone and messed up with the title and a missing comma.

rbuffat:
The license information: https://data.imap.maryland.gov/datasets/019cd3b69bc64a878b5dc506b633df84/explore
states the following:

> The Spatial Data, and the information therein, (collectively the "Data") is provided "as is" without warranty of any kind, either expressed, implied, or statutory. The user assumes the entire risk as to quality and performance of the Data. No guarantee of accuracy is granted, nor is any responsibility for reliance thereon assumed. In no event shall the State of Maryland be liable for direct, indirect, incidental, consequential or special damages of any kind. The State of Maryland does not accept liability for any damages or misrepresentation caused by inaccuracies in the Data or as a result to changes to the Data, nor is there responsibility assumed to maintain the Data in any manner or form. The Data can be freely distributed as long as the metadata entry is not modified or deleted. Any data derived from the Data must acknowledge the State of Maryland in the metadata.

> Any data derived from the Data must acknowledge the State of Maryland in the metadata.

We should clarify with the data owner if e.g. an entry under https://wiki.openstreetmap.org/wiki/Contributors would be sufficient to meet this criterion.

Edit: there seems to be an entry for Maryland State Government in https://wiki.openstreetmap.org/wiki/Contributors

cicku:
I’d say all MD iMap data use this license. So if MD aerial imagery got accepted without issue, this one should be fine as well.

rbuffat:
> I’d say all MD iMap data use this license. So if MD aerial imagery got accepted without issue, this one should be fine as well.

Not necessarily. The license terms could have changed, or somebody could have overlooked this requirement.

cicku:
@talllguy Hi, would you please take a look at this?

cicku:
@rbuffat Looks like that you edited your comment there. Now I don't think there is any issue not merging it based on https://wiki.openstreetmap.org/wiki/Contributors.

talllguy:
Hey, I'm [an appointed member to the Maryland Council on Open Data](https://twitter.com/talllguy/status/476713251934576640) and [we lobbied for the legislature to pass a law](http://mgaleg.maryland.gov/mgawebsite/Laws/StatuteText?article=gsg&section=10-1501&enactments=false) to make this data free and available. All of the state's imagery is free from any fee but is subject to restrictions. Those terms are typical as a disclaimer because lawyers but it shouldn't prevent the use of state data on OSM.

I can bring it up to the council for clarity but don't expect a quick answer. We're in the middle of long process around _commercial_ use, which I don't think applies to OSM, but that takes precedence for now. I have [permission for use and its all documented on the wiki](https://wiki.openstreetmap.org/wiki/Contributors#United_States).

Now, all that said, I'm a frequent user of the LiDAR for trail mapping and have spoken about it at conferences. It is an excellent resource but it is also **highly subjective to advanced interpretation such that I would not recommend it be included** in the default layers for iD. Very often things that appear in LiDAR are not actually there, such as an old road that has returned to nature with no visible path (see example below). These require survey to verify and I've made mistakes myself adding things that are not actually out there. I'd say that lidar interpretation is a type of advanced mapping that is best for advanced users. The imagery can be added to JOSM directly by WMTS. If we do decide to include this, I think it should be put into the 'map' category, so it doesn't appear as a type of imagery, but rather as a map like the topo maps.

image
^ screenshot of Maryland LiDAR at Lake Roland, a park in Maryland. [Link to OSM location](https://www.openstreetmap.org/#map=18/39.38805/-76.65067).

cicku:
Maybe, but MassGIS was accepted.

image

cicku:
I'm going to close this one, will propose using USGS 3DEP later.

PR #1185, opened 2021-07-10
Create SouthTyrolOrthofoto2020.geojson

MarcelloPerathoner:
Year 2020 orthofoto with full coperture.


tyrasd:
It seems that this Layer is only available under the terms of the CC-BY (4.0) license, which normally requires a waiver to be used for usage in OSM mapping (see the discussion in #896). I have just sent an email to the responsible contact person to ask for a waiver for this layer. Let's wait until they reply.

MarcelloPerathoner:
Didn't know that CC-BY wasn't good enough. But it doesn't matter because CC-0 is granted to all "geografic and cartografic data" of the "Abteilung 28" by this writ: https://www.provinz.bz.it/natur-umwelt/natur-raum/downloads/Licence_CC0.pdf

MarcelloPerathoner:
I have switched the license to CC0. 
See the license writ: https://www.provinz.bz.it/natur-umwelt/natur-raum/downloads/Licence_CC0.pdf 
which is dated: 3/22/2021, 7:26:07 PM

After reading that writ IMO the data is at least double licensed under CC0 and CC-BY-4 and thus no attribution is necessary, because we can select the license which suits us best. Of course simply asking for permission may be the best way forward.

rbuffat:
Let's wait for an answer to the email. The data is still listed as cc-by: http://geoservices.buergernetz.bz.it/geonetwork/srv/ger/pdf?uuid=ae16961e-14e7-4c74-ba05-0cd407bc1fd3

tyrasd:
Good news: The 2020 layer will be available under the CC-0 from now on. The metadata on the website is not yet updated, so maybe let's wait for that. In the meantime, I've updated the layer extent slightly: it has coverage also for some areas outside of SouthTyrol, especially around its southern border.

tyrasd:
:heavy_check_mark: :smiley: 

image

PR #1201, opened 2021-07-14
Update Brussels UrbIS resources

jbelien:
- Delete UrbIS 2013 service that doesn't seem to exist anymore (and became irrelevant since there is up-to-date UrbIS layers)
- Add bilingual UrbIS layer
- Add aerial imagery 2020


rbuffat:
Thanks @jbelien !

PR #1203, opened 2021-07-19
Berlin areal images: Add 2021, update URLs for older images

tordans:
There are multiple thinks happening in this PR:

- Add 2021 images, mark as best …
Downgrade the 2020 images (not 'best' anymore).

- New infrared images from 2020, remove old 2016 images …
We still have the regular images for 2016 but now that there are new infrared images from 2020, I think we can cleanup the list a bit and only keep the new images.

- Update URLs to new server url structure …
The old URLs have a redirect, but it's better to have the current URL.

- Housekeeping: Cleanup spaces in .geojson files …
Empty end of line spaces + indentation to 2 spaces, like all the other berlin file.

- Update `max_zoom` to 20 after testing on https://maps.berlin.codefor.de/


polarbearing:
```
    "name": "Berlin/Geoportal DOP20RGB (2021)",
     "type": "tms",
     "category": "historicphoto",
```
2021 is not historic?

tordans:
> 2021 is not historic?

I thought about that too. But …
- there is no clear definition of the values in https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L46-L59 
- we always hat this value for all layers
- I don't think it makes a difference if we changed the latest to `photo`
- it is another think one needs to remember when adding the next layer next year …

polarbearing:
IIRC there is ```historicphoto``` for older ones, ```photo``` for the current and a boolean flag for the ```"best"``` one?
I'd assume an image consumer such as iD to show their user the best/current as default and the others on request. 
Thus I think it makes a difference.

tordans:
> I'd assume an image consumer such as iD to show their user the best/current as default and the others on request.

iD does just list all that are available and all other data consumers switched to the JOSM index. It might be that iD filters everything that is a map from the list. I think it's fine as is. Let's wait for feedback from the maintainers.

rbuffat:
JOSM uses the category `photo` for current imagery and `historicphoto` for outdated imagery.

tordans:
@rbuffat thanks for that info! I wonder, at what point is an areal image outdated? I consider all those Berlin images up to date enough to be used for mapping; they show different lighting situations and details each year. And of course the year indicates which images are newer. 

Also, since JOSM does not use ELI directly and we did not come up with an ELI-consumer who uses the key, I wonder what the effort is for.

----

But, while we are on the topic of JOSM: It would be great, if the URLs in the JOSM index where checked against this PR so the unnecessary redirects of the old URLs are removed from the index.

rbuffat:
I suppose most often "outdated" is interpreted as just that there is newer imagery available. Personally, while I think the category is something useful I'm also not too keen of the historic part, as this information can be determined automatically using the geometry and the start date / end date. But I might not be aware of important aspects. 

> Also, since JOSM does not use ELI directly and we did not come up with an ELI-consumer who uses the key, I wonder what the effort is for.

* It is good to keep the indexes as much in sync as possible
* There might be consumers in the future using the attribute

You should be able to edit https://josm.openstreetmap.de/wiki/Maps/Germany directly after you created an account. Otherwise, normally the changes in ELI are synced by the JOSM contributors within a few days/week.

PR #1206, opened 2021-07-21
Add South Carolina Aerial Imagery

cicku:


cicku:
@rbuffat I may need your help here regarding the second failed check.

rbuffat:
Both the [tms](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification) and [wmts](https://www.ogc.org/standards/wmts) services are similar in that they split a dataset into tiles that can be accessed using an URL. Although, wmts is a bit more complex compared to tms. 

For  `"type": "tms"` sources in ELI the URL attribute is the URL directly to the tiles. This allows clients to directly access the tiles by substitute the x,y and zoom parameters in the URL. For `"type": "wmts"` sources, the URL parameter should point to the "Capabilities XML" of the WMTS service. (e.g. https://tiles.arcgis.com/tiles/RvqSyw3diI7dTKo5/arcgis/rest/services/SC_2020_RGB/MapServer/WMTS)

If a WMTS service supports the right projections, a WMTS source can be defined as a tms source using the correct URL.
As not all OSM editors support WMTS this is actually the preferred way at the moment. 

 ```
 <TileMatrixSet>
<ows:Title>TileMatrix using 0.28mm</ows:Title>
<ows:Abstract>The tile matrix set that has scale values calculated based on the dpi defined by OGC specification (dpi assumes 0.28mm as the physical distance of a pixel).</ows:Abstract>
<ows:Identifier>default028mm</ows:Identifier>
<ows:SupportedCRS>urn:ogc:def:crs:EPSG::3857</ows:SupportedCRS>
<TileMatrix>
 ```

If I'm not mistaken this indicates that the default028mm TileMatrixset is in the [EPSG:3857](https://epsg.io/3857) projection.

rbuffat:
The URL you provided seems actually to be a WMTS service in tms format. Thus changing the type to tms and removing the supported_projections attribute should resolve the failing tests. 

Could you also state some information about the license of the data?

cicku:
I got in contact with South Carolina Revenue and Fiscal Affairs Office. They told me that the map should be free to use for general public as long as it is not used for resale.

However that is not very enough, I will try to obtain an official response as a record and get back later.

arch0345:
Have you gotten any word back?

cicku:
I don’t think we can use it. They do not reply = no approval.

PR #1218, opened 2021-08-09
fix coverage and tweak some SouthTyrol layers

tyrasd:
* fix coverage of highres "solartyrol" DTM/DSM layers
* use https links where possible
* use a more consistent naming scheme for file names


tyrasd:
before (left) vs. after (right):

image

rbuffat:
Thanks @tyrasd !

PR #1219, opened 2021-08-11
Add Code-Point Open tile server

Cj-Malone:
This is my first time, so you should check it properly, I used `OSMUK-Cadastral-Parcels.geojson` as a base.

I was confused with a few of the properties. Does `permission_osm` mean the tile server, or the data source? The [tile server](https://raggedred.net/codepoint/) is explicit for OSM, the [data](https://www.ordnancesurvey.co.uk/business-government/products/code-point-open) is implicit due to the license. `attribution/url` would presumably be to do with the license of the data, but then how is that different to `license_url` and how do I link to the tile server webpage?


rbuffat:
@Cj-Malone Thanks for your PR!

Would it be possible to remove the white space from the filename? Then the test script should find the file. 

`license_url`is typically used to point to a website that states the license of the data (e.g. its metadata page)
All properties of`attribution` are displayed by OSM editors if the data is shown. (But nothing more) 
`permission_osm` is used to indicate if the permission to use the data for osm is implicit or explicit: https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L73-L80 This is an optional property.

The [OGL 3.0 requires attribution](https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/):
```
acknowledge the source of the Information in your product or application by including or linking to any attribution statement specified by the Information Provider(s) and, where possible, provide a link to this licence;
``` 
Thus it should be checked if the entry in https://wiki.openstreetmap.org/wiki/Contributors#United_Kingdom is up-to-date.

Cj-Malone:
If [this](https://cj-malone.github.io/editor-layer-index/) is to be believed, there are still issues. Any ideas?

rbuffat:
> If [this](https://cj-malone.github.io/editor-layer-index/) is to be believed, there are still issues. Any ideas?

Not sure what you mean. 

> ERROR:root:URL validation error:  https://www.raggedred.net/tiles/codepoint/{zoom}/{x}/{y}.png

Not yet sure what is going on here. The URL looks fine to me. Maybe something wrong with the test. I will look into this.

boothym:
Just wondering if there's been any progress with this? Would be a useful overlay to have 😃

grischard:
The URL works in iD for me. Force merging this, feel free to revert if I've done anything stupid, @rbuffat :)

Issue #1181, opened 2021-06-27
Add CyclOSM layer


LySioS:
Hi everyone,
The requested feature is in the title.

Excuse me if this issue report is not good enough, this is my first time.

Thanks for your work.
Lysios

## URL of the imagery
I think that's the one > `https://{s}.tile-cyclosm.openstreetmap.fr/[cyclosm|cyclosm-lite]/{z}/{x}/{y}.png`

## Two letter country code
world

## Desired layer name
CyclOSM

## Licenses
[https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/LICENSE.md](https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/LICENSE.md)

## Is the license compatible with OSM?
(required)

## Attribution (text, URL, required?)
I do not know

## Privacy policy (URL)
I do not know

## Icon
[https://raw.githubusercontent.com/cyclosm/cyclosm-website/master/images/logo.png](https://raw.githubusercontent.com/cyclosm/cyclosm-website/master/images/logo.png)

## Bounding polygon
(if possible, provide a bounding polygon where the imagery is valid, e.g. as a GeoJSON or OSM file.)<br/>
See FAQ at https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#how-can-i-draw-a-bounding-polygon) for information about how to draw a bounding polygon.

Asteliks:
is this done?

PR #1226, opened 2021-08-16
Add CyclOSM layer

PanierAvide:
Related to #1181


rbuffat:
@PanierAvide 

Thanks for your PR! Would it be possible to include a license_url and privacy_policy_url? Also, it would be good to not include the imagery* files in your PR. These are automatically updated when the PR is merged.

PanierAvide:
OK, sorry I messed up branch management and had to open a new PR, this is this one #1238

PanierAvide:
Also, I'm not sure to have seen in docs that the two mentioned fields are mandatory, if this is true now, maybe we can update docs to avoid this pitfall to new contributors ? :wink:

Issue #1172, opened 2021-06-10
mapproxy.openstreetmap.se not reachable


rbuffat:
It looks as openstreetmap.se is not reachable for more than a year ([Archive.org has the last entry from April 7th 2020](https://web.archive.org/web/2020*/http://openstreetmap.se)).

There are currently 3 sources pointing to this server: https://github.com/osmlab/editor-layer-index/search?q=mapproxy.openstreetmap.se

Does somebody know if this server will come back from the dead eventually or has contact with the people that were operating openstreetmap.se?

@grischard @NKAmapper @don-vip

NKAmapper:
https://github.com/karlwettin/carto-style-hydda/issues/3

@karlwettin

rbuffat:
@grischard @simonpoole @cquest 
Is there some OSM infrastructure available to host imagery specific for OSM mapping/purposes?  I assume not all local OSM communities have the resources to host imagery on their own. 

That said, I also do not know if there is still a demand for these datasets. At least for Helsingborg, there seems to be newer data available in public domain: https://helsingborg.opendatasoft.com/explore/dataset/flygbilder/information/

simonpoole:
SOSM has plenty of resources for such things and continues to operate a mapproxy instance.

But I suppose we really need some feedback from the SE community if the layers are/were really useful.

grischard:
Yes there's a server to host imagery directly. A shared mapproxy would be possible too, but no one has ever volunteered the time to set it up.

rbuffat:
Closing this issue as there seems to be no interest.

PR #1234, opened 2021-08-22
Remove europe/sweden entries that are broken for a long time

rbuffat:


rbuffat:
See https://github.com/osmlab/editor-layer-index/issues/1172

I don't think that mapproxy.openstreetmap.se will be online anytime soon. The server is dead since April 2020. If this happens, this PR can be reverted.

PR #1235, opened 2021-08-23
Update New Jersey backgrounds to 2020 version

cicku:
This commit pulls in new 2020 imageries for the state of New Jersey. It does not bring in [leaf on imagery](https://njgin.nj.gov/njgin/edata/imagery/) which is published at the same place because it is literally the same from NAIP.

The old 2015 imageries should be removed because the newer is better.


rbuffat:
@cicku Thanks for this PR! I would suggest keeping the 2015 imagery. The old imagery allows seeing why something was mapped when it does not correspond to the current imagery. But I guess, there can be different opinions about this. 

Regarding WMS 1.3.0: with WMS 1.3.0 the SRS parameter is now called CRS. There is a jupyter notebook that can facilitate finding the correct URL and wms_projections: https://mybinder.org/v2/gh/rbuffat/eli-helper/master?urlpath=apps%2F/create_wms_template_appmode.ipynb (it can take a long time to load)
But as long as there is the URL and wms_projections specified work, there is a weekly bot that should update the URL / wms_projections if something changes. Thus you don't have to worry too much about this.

cicku:
Yes I understand the difference between 1.1.1 and 1.3.0.

The reason why I did not put 1.3.0 is also because of the unknown projections. So I just copied part of the 2015 file into the new one.

I have no objection keeping 2015, so when you decide to merge it, please feel free to ignore https://github.com/osmlab/editor-layer-index/pull/1235/commits/2c799d40d57f064b208f8134b36005d2657a2ad9.

cicku:
Looks like I misunderstood what max_zoom is. Apparently what I checked was the max zoom level of basemap tiles. I've made the change.

rbuffat:
The metadata claim a resolution of 1 foot per pixel. This should roughly correspond to a zoom level of 19: https://wiki.openstreetmap.org/wiki/Zoom_levels
Looking at the zoom check it looks that higher levels are just interpolation done by the wms server. Thus, I set the zoom level to 19.

Issue #1035, opened 2021-01-31
Rhode Island 2019 Spring Aerial Imagery


ZeLonewolf:
When **requesting the addition of an imagery layer**, please provide the following information. Otherwise, delete the text and type your issue.

## URL of the imagery
https://maps.edc.uri.edu/rigis/services/IMG/RI_201904_RGB_3in_web/ImageServer/WMSServer?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=0&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}

## Two letter country code
US

## Desired layer name
Rhode Island Spring 2019 Aerial Imagery

## License
~~CC BY 4.0~~ Released without license per state authorities

## Is the license compatible with OSM?
See: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data
This layer automatically loads in the layer list in JOSM.

## Attribution (text, URL, required?)
This dataset is provided 'as is.’ The producer(s) of this dataset, contributors to this dataset, the Rhode Island Geographic Information System (RIGIS) consortium, the State of Rhode Island, and the University of Rhode Island do not make any warranties of any kind for this  dataset, and are not liable for any loss or damage however and whenever caused by any use of this dataset.  Please acknowledge both RIGIS and the primary producer(s) of this dataset in any derived products.  Versions of the RIGIS logo suitable for both printed and web-based products are available at https://info.rigis.org/logos/.
https://www.rigis.org/pages/2019-spring-imagery

## Privacy policy (URL)
None listed

## Icon
https://data.rigis.org/assets/img/favicon.ico

## Bounding polygon
[RI.txt](https://github.com/osmlab/editor-layer-index/files/5899046/RI.txt)

rbuffat:
There seems to be no metadata page, but on https://www.rigis.org/ at the bottom it is linked to https://creativecommons.org/licenses/by/4.0/

@ZeLonewolf Thanks for creating this new issue.  As in https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data pointed out CC BY 4.0 requires a waiver.  Unfortunately, being included in the JOSM index does not guarantee the existence of such a waiver, as this is not checked in JOSM. Currently, I did not yet find an existing one. 

CC @Klumbumbus @don-vip @stoecker

ZeLonewolf:
[...]

PR #1237, opened 2021-08-23
Add Rhode Island Aerial Photo

cicku:
For #1035.

I'd prefer putting the copyright clarification into a single page.


cicku:
I also want to add a note that the latest imagery is from Spring 2021 but the website admin told me that they have no plan to enable WMS support.

The currently supported formats are: JSON, SOAP and WCS. Apparently there is no way to use them right now in iD.

rbuffat:
> I also want to add a note that the latest imagery is from Spring 2021 but the website admin told me that they have no plan to enable WMS support.
> 
> The currently supported formats are: JSON, SOAP and WCS. Apparently, there is no way to use them right now in iD.

@cicku Sometimes the [ArcGIS JavaScript View](https://maps.edc.uri.edu/rigis/rest/services/IMG/RI_202009_RGB_3in_web/ImageServer) can be turned into a WMS source compatible format (Using ESRIs rest API):

```
...
"type": "wms",
"url": "https://maps.edc.uri.edu/rigis/rest/services/IMG/RI_202103_RGB_3in_web/ImageServer/exportImage?f=image&format=jpgpng&imageSR=3857&bboxSR=3857&bbox={bbox}&size={width},{height}&foo={proj}",
"available_projections": ["EPSG:3857"],
...
```

Some suggestions for the 2020 imagery:
* Maybe it would be better to use the year of the imagery in the id and to use latest, if the imagery is not always the latest
* The max zoom looks a bit low. There is a juptyter notebook to check zoom levels you might want to try out: https://mybinder.org/v2/gh/rbuffat/eli-helper/master?urlpath=apps%2F/check_minzoom.ipynb

The imagery seems not to be orthorectified. It might be worthwhile to put this information in the description.https://rigis-edc.opendata.arcgis.com/pages/2021-spring-imagery

>While these aerial photographs are very detailed (3-in pixel resolution), they are not traditionally orthorectified. Their horizontal accuracy may vary throughout the state. Potential users are encouraged to carefully evaluate the suitability of these images before use.

cicku:
@rbuffat Thank you for the hints. I will definitely check it.

Meanwhile, is it possible, in the near future, to add a variable/field in the ELI, so iD will be able to alert user whether the imagery is orthorectified or not?

rbuffat:
There is the valid-georeference property: https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L196. But I do not know if iD uses this property or not.

cicku:
I will update RI geojson soon again with updated information.

PR #1241, opened 2021-08-30
Perfect the Rhode Island imagery based on #1237

cicku:
This commit:

1. Corrects the max zoom level.
2. Indicates that the image is not orthorectified.
3. Specifies the imagery date.

### DIFF
--- a/sources/north-america/us/ri/RIGIS_Aerial_Photo_21s.geojson
+++ b/sources/north-america/us/ri/RIGIS_Aerial_Photo_21s.geojson
@@ -8,7 +8,7 @@
             "required": false
         },
         "name": "Rhode Island Aerial Photo (Spring 2021)",
-        "url": "https://maps.edc.uri.edu/rigis/rest/services/IMG/RI_202103_RGB_3in_web/ImageServer/exportImage?f=image&format=jpgpng&imageSR={wkid}&bboxSR={wkid}&bbox={bbox}&size={width},{height}",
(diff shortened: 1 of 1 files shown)


cicku:
Is "{proj}" really needed? Or should I remove "wms"? @rbuffat

rbuffat:
@cicku It will depend on the editor, but to be sure, it would be better to add it. WMS typically requires the proj parameter, thus editors could expect the `{proj}` and fail if it is not available. 

In the future, it would be good to distinguish between wms and the esri API.

cicku:
@andrewharvey Feel free to merge this one since they are just some minor changes.

andrewharvey:
@cicku if the Fall 2020 is still available it would be best to create two sources so users can toggle between both in the editor, is the old one still available?

cicku:
The old ones are always available but because I committed the file not long before this PR, I would start doing as you said above from 2021 Spring and keep it when Fall comes out. The reason why I didn’t select 2021 imagery in #1237 is because I didn’t know iD supports reading from JSON. 

The imagery has not been officially included in any version of iD and it is not even orthorectified, so there isn’t much impact.

cicku:
@rbuffat I'm not inclined to keep the old imagery in this PR. I submitted the original PR 5 days before at #1237, was expecting to fix this as soon as possible. Then, you know...

andrewharvey:
I'll merge this PR because it's better to have newer imagery than older imagery, but it would be nice to have another PR which re-adds the Fall 2020 imagery as a new source. As a mapper it is useful to sometimes be able to use older imagery as well.

Issue #1213, opened 2021-07-31
Determine (automatically) which WMTS services support the OGC WMTS Simple Profile


simonpoole:
Not very many OSM editors support WMTS sources, however those WMTS servers that have a TileMatrixSet that conforms to the OGC WMTS Simple Profile provide an URL that can be used for tile access just like as for a OSM/google compatible tile server.

If the OGC WMTS Simple Profile is supported it could make sense to either automatically add a new field with the URL to WMTS ELI entries, or simply duplicate them as normal "TMS" entries, so that these can be used without clients having to implement full blown WMTS support. 

 But as a first step, it would simply be interesting to know how many of the existing entries have a conformant TileMatrixSet.

rbuffat:
This is a good idea.

TODOs would be:
- [ ] Query WMTS sources to check if they support the Simple Profile / urn:ogc:def:wkss:OGC:1.0:GoogleMapsCompatible
- [ ] Construct tms URLs from the service metadata
- [ ] Ideally, there should be support for mirrors in ELI (or wmts sources are deleted in favor of TMS)
- [ ] Add check for new wmts sources if simple profile is supported

rbuffat:
Currently, there are not a lot of WMTS sources in ELI. Two of those sources support EPSG3857 and could be added as tms, but for them already an alternative server with WMS is available. But it would be good to switch these to tms.

```
MCGIS-County-NAIP-Imagery-2019 https://mcgis.mesacounty.us/image/rest/services/Mosaic_Datasets/MesaCounty_2019/ImageServer/WMTS/1.0.0/WMTSCapabilities.xml
ID: default028mm tile_matrix_crs: urn:ogc:def:crs:EPSG::3742

Bern-AmtlicheVermessung https://map.bern.ch/arcgis/rest/services/AV_farbig/MapServer/WMTS?
ID: default028mm tile_matrix_crs: urn:ogc:def:crs:EPSG::2056

npi-svalbard-landsat https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Satellitt_Svalbard_WMTS_25833/MapServer/WMTS/
ID: default028mm tile_matrix_crs: urn:ogc:def:crs:EPSG::25833

NPI-Svalbard-ortofoto https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_25833/MapServer/WMTS/
[...]

PR #1243, opened 2021-08-30
WIP: check if layer of wmts sources support epsg 3857

rbuffat:
Towards implementing https://github.com/osmlab/editor-layer-index/issues/1213

Ammend strict check for new or updated sources to check if wmts sources contain layers in EPSG3857 that could be added as tms, and thus would be accessible to OSM editors not supporting wmts.


rbuffat:
Superseded by https://github.com/osmlab/editor-layer-index/pull/1377

PR #1245, opened 2021-09-01
Create DC Orthophoto

cicku:
The 2021 Ortho will be updated by the Thanksgiving.


cicku:
@rbuffat should I wait for few weeks to get 2021 imagery into here directly?

andrewharvey:
It's useful to have 2019 imagery even when 2021 becomes available

cicku:
Thank you, I will submit the PR when it’s ready. They told me a delay is expected.

PR #1246, opened 2021-09-01
Update NAIP imagery based on watchdog result

cicku:
This PR:

1. Exclude Nevada Test and Training Range from the boundary.
2. Add license and privacy fields.
3. Specifies start and end dates. Note that these two may need to be updated in the future upon new image acquisition is complete.
AREA 51


cicku:
I have not excluded other army facilities like Fort Bragg in this PR. Large places like this will be done in the future.

image

rbuffat:
Thanks @cicku

PR #541, opened 2018-08-21
Update NC Latest Orthoimagery

rivermont:
* Tagged as 'best' imagery for NC. It's nearly the highest quality in most areas and is ortho-rectified and thus has almost negligible offset throughout.
* More imagery was taken in 2015 and 2017, so added that to the layer description.

PR #1247, opened 2021-09-03
Update NC imagery

cicku:
This PR:

1. Uses the cached WMTS service
2. Corrects the boundary of NC Ortho and include NDVI imagery for east NC

My question:

1. The cached WMTS service has minor coverage differences (some highways close to military area) if comparing with the WMS service (currently in the repo but very slow if out of the state). Should I also exclusively add a 2020 dataset here? The NDVI imagery shares the same coverage with the 2020 dataset though.

image

image

2. Which type is preferred? WMS or WMTS/TMS?


rbuffat:
There is a tradeoff between the accuracy of the geometry and the resulting filesize.  It is sufficient that the geometry only roughly represents the imagery.  The updated geometry is too detailed.

Ideally, the number of points of the geometry is less than 100 or in the range of a few hundred points. Naturally, this depends on the complexity of the imagery.

cicku:
Should I drop smaller holes? I incline to set correct boundary (may not be detailed though) because the last time when I used the imagery in Indiana it took me a few minutes to realize that the boundary was wrong but not my computer network had problem.

cicku:
Another thought, since I intend to set the ortho as the best available. If I don’t add TN orthoimagery in the future, then the boundary could be a little bit rough.

Currently VA is already in the repo, SC is being added. TN is not available (TN has 2 different photos captured by different entities, one is currently restricted the other one is not).

I would not like to send users to the best imagery of another state…

cicku:
I have optimized them and they are much smaller now.

cicku:
@andrewharvey Whenever you have time feel free to merge this.

cicku:
@rbuffat They are smaller, do you have any suggestion to make it even smaller?

rbuffat:
@cicku There is indeed an issue for border regions. I think we just have to accept that this is a limitation of the expressiveness of the "best" flag.  Thanks for optimizing the shapes.

rivermont:
I guess this makes #541 moot then so I'll close it.

cicku:
Oh thanks for point out - I did not even notice that old ticket which was on the second page in issues.

PR #1254, opened 2021-09-04
Add Brazil-Sobral imagery

matheusgomesms:
We received explicit permission to use their imagery on OSM, which is of high resolution and still a bit recent.


matheusgomesms:
There are 3 issues with this PR:
1 - no Privacy Policy (I couldn't find one at all, unfortunately)
2 - somehow the strict check says that the geometry is wrong. I believe this is related to the 3 districts in this imagery. Maybe a multipolygon solves this?
3 - Although this imagery works normally on [JOSM](https://josm.openstreetmap.de/wiki/Maps/Brazil#SobralOrthophoto2017), I can't manage to make it work on iD (custom imagery). Not sure what happens.

@rbuffat could you please take a look into this? There's no rush, but this is a very good imagery! Would be nice to have it on iD.

willemarcel (approved):
@matheusgomesms It worked on iD for me. It's ok not having a privacy policy. I improved the geometry, but it spans over an area a bit bigger than the imagery bbox yet.

PR #1255, opened 2021-09-04
Add Brazil-Belo Horizonte Imagery

matheusgomesms:
2015 imagery from Belo Horizonte. According to [Brazil Open Data website](https://dados.gov.br/dataset/portal-bhmap), all BH Geo portal is on ODbL license. The Brazilian community also uses [their data](https://wiki.openstreetmap.org/wiki/Pt:Importa%C3%A7%C3%A3o_de_dados_do_BHMap) since a while ago.


cicku:
Don’t you think all photos from at least 3-4 years ago should be `historicphoto` at all?

matheusgomesms:
> Don’t you think all photos from at least 3-4 years ago should be `historicphoto` at all?

I personally don't mind if it's used "photo" or "historicphoto". I just used "photo" from the JOSM definition: _the historic variant of photo and map should be used either if the source shows very old data or if there is a newer variant of the same source from the same provider available._ 

Since there's no newer photos from the same source... But yeah, if you prefer I could switch it to "historicphoto" since it's a bit outdated, no worries from my side!

cicku:
Another question, the website is like this:

image

It does not specify the version, I'm assuming that Dados GEO uses 4.0 as it is the latest.

If that's true, a waiver/permission is needed: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/.

matheusgomesms:
@cicku That's true, my mistake! It seems it's on CC BY 4.0, so I'll talk to them to get permission. Thanks for the heads up! :D

matheusgomesms:
Hey guys,

When I opened this PR the data was on ODbL, but it was modified right in Sep 21.

Anyway, I sent them an e-mail and [they let us use their data again](https://wiki.openstreetmap.org/wiki/File:Autoriza%C3%A7%C3%A3o_BHGeo.png).

I guess we're good to add this layer now!

@cicku @willemarcel @rbuffat

andrewharvey:
Usually for CC BY we ask the entity to agree to the two standard waiver terms from https://drive.google.com/file/d/0B3PN5zfbzThqeTdWR1l3SzJVcTg/view?resourcekey=0-PzVtHArfxvbYidpW2-AVTg

Do you think you can get them to agree to those standard waiver terms?

I'm not across the legals enough to know if the email response you got is sufficient.

andrewharvey:
I think using `photo` is fine in this case, editors can already provide a warning that the imagery is old based on the data from the metadata even without `historicphoto`.

matheusgomesms:
> Do you think you can get them to agree to those standard waiver terms?

I don't think so, this kind of official permission is very hard to obtain.

However, given that:

1 - government in Brazil is very permissive for this type of situation (even enforced by law)
2 - the data was in ODbL before
3 - government GIS usually use OSM (as map base, mainly)
4 - almost all permissions we got in Brazil, to use data into OSM, was obtained that way (even some layers here already accepted in ELI)

I do not think we would have any problem using this source (especially because we obtained permission), even without the waivers.

Not sure what you guys think, however. If truly necessary, we can push for those waivers, but I doubt so.

cicku:
I don't have strong opinion on this.

My understanding is that if they could change from odbl to CC without notice, they can also change to whatever anytime in the future...

matheusgomesms:
The reason that they changed it to CC-BY was "because we thought the other one was more restrictive. [...] This is our goal, that data can be used and shared." (their words, on the linked e-mail on my previous message).

Therefore, I conclude that they wanted a more open-license, which is a great sign (not entering the discussion that which one is more open, just their feelings that CC-BY is more open than ODbL). So I don't think they will switch to another more restrictive license out of the blue (because again, Brazilian local governments are enforced by federal law to publish their data in an open way).

Also, we have an e-mail stating clearly that they want us to use their data. The usage of e-mails instead of official letters or more "official communication" is debatable, I know, but this is how it works in Brazil and LatAm (and the great majority of the other ELI layers were approved on that basis).

matheusgomesms:
2 years since they are still using CC-BY, and we have explicit permission to use their data. Can I (or anyone here) merge this PR?

andrewharvey:
@simonpoole Any thoughts on if https://wiki.openstreetmap.org/wiki/File:Autoriza%C3%A7%C3%A3o_BHGeo.png is sufficient in lieu of the waiver?

simonpoole:
I haven't had anything to do with such things for a while, but fwiw I'll give it a look later this week.

matheusgomesms:
Any updates, colleagues? Can I merge it? :)

matheusgomesms:
More than 2 years they gave us explicit authorization to use their imagery and mappers from a 2.7 million inhabitants city can't use this high-res imagery.

Still [CC-BY](https://geonetwork.pbh.gov.br/geonetwork/srv/por/catalog.search#/metadata/5687b0f1-bcc0-4a2c-aabb-dc4256297f6c). If we take more time, this won't be `historicphoto`, but `ancientphoto`.

I'm taking full responsibility and merging this. If anyone disagrees, please let me know, reopen this discussion, or revert this merge.

PR #1266, opened 2021-09-19
New URLs for Norwegian imagery

NKAmapper:
Plus one max zoom adjustment.
Also adds privacy policy url's.
Syncs with [JOSM maps Norway](https://josm.openstreetmap.de/wiki/Maps/Norway).


cicku:
Please check attribution url.

PR #1268, opened 2021-09-24
Add Mat-Su Borough (MSB) Orthoimageries

cicku:
1. MSB permits the public use of its aerial imagery products.
2. MSB captures by region on a yearly basis.
3. The 2021 imagery has not been published, it will be added later.
4. No privacy disclaimer can be found on their website.


rbuffat:
I currently unable to access any matsugov.us server. I will try again later.

cicku:
Thank you for commenting on this. It also took me a fair amount of time making contact with them because the Boro suffered a ransomware attack and got itself on BBC. They have a strict firewall as well as spam filter policy implemented.

https://www.bbc.com/future/article/20190108-the-cyber-attack-that-sent-an-alaskan-community-back-in-time

rbuffat:
oh wow.

Could you post an (anonymized) screenshot of your conversation or one of the terms to the openstreetmap wiki?

cicku:
I only called them and have no record on this. May send them email later when I’m free (busy now).

PR #1270, opened 2021-09-25
Update VA imagery

cicku:
1. Update date
2. Correct links
3. Load from origin instead of poor performance proxy
4. Update the boundary to avoid overlapping with DC, NC and MD


cicku:
The website (copyright disclaimer) is being updated and the file is not found on the server. The Google cache shows that it used to be there.

cicku:
VGIN personnel told me the file I mentioned above has been dated and removed recently.

However, the license has not been changed. Still public domain with attribution as #540 had discussed previously.

PR #1274, opened 2021-10-03
Fix ewidencja budynków geoportal

Asteliks:
It's that time of the year when I contribute to open source projects, meaning its Hacktoberfest time! Thus it would be great if this PR can be given the 'hacktoberfest-accepted' label.

Basically since there has been no change on the iD side I have removed the attribute best from the overlays as this makes them the default layer for Poland and because of that the first thing one sees when opening the editor is this:

image

Instead of something like this: 
image


Asteliks:
@rbuffat ready for merge

Asteliks:
@andrewharvey @grischard 
Hi, he seems to be unavailable. Can you guys help me out?

Issue #741, opened 2019-11-25
Remove PolandGeoportal.geojson and edit other


maro-21:
- [x] Please remove "PolandGeoportal.geojson"
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/pl/PolandGeoportal.geojson - it doesn't work anymore.

- [ ] edit  /sources/europe/pl/WrocawOrthophotomap2015(aerialimage).geojson
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/pl/WrocawOrthophotomap2015(aerialimage).geojson
and add
	- `"start_date": "2015-06-05",`
	- `"end_date": "2015-06-06",`
	- `"i18n": true"`

- [ ] edit /sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/pl/Geoportal2Orthophotomap(aerialimage).geojson
add
	- `"i18n": true"`

- [ ] edit sources/europe/pl/polskae-mapanetBuildings.geojson
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/pl/polskae-mapanetBuildings.geojson
add
	- `"i18n": true"`

grischard:
Hello @maro-21, and welcome! I see this is your first issue; thank you for reporting this.

Would you like to try creating a pull request yourself to change these? Happy to assist if you need a hand.

maro-21:
Second :)

Yes, please. But I don't use Linux now, so I don't think I can do it by myself.

SomeoneElseOSM:
Does https://www.openstreetmap.org/user/SomeoneElse/diary/391681 help?  That's talking about a different repository, but it should explain how to do a textual change "just online". There are also git clients for Windows etc. Of course.

andrewharvey:
Looks like this was closed by #1277

PR #1277, opened 2021-10-03
added "i18n": "true"

Asteliks:
It's that time of the year when I contribute to open source projects, meaning its Hacktoberfest time! Thus it would be great if this PR can be given the 'hacktoberfest-accepted' label.

This PR adds `"i18n": "true"` to all the layers available for Poland so that the Polish community won't bother me with this anymore. Also solves this: https://github.com/osmlab/editor-layer-index/issues/741
and all the other issues requesting `"i18n": "true"` for Poland

### DIFF
--- a/sources/europe/pl/BdzinBuildings.geojson
+++ b/sources/europe/pl/BdzinBuildings.geojson
@@ -3,7 +3,7 @@
     "properties": {
         "id": "Bedzin-PL-buildings",
         "name": "Będzin: Buildings",
-        "i18n": "true",
+        "i18n": true,
         "type": "wms",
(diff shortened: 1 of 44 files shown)


Asteliks:
@rbuffat ready for merge. It fails the check due to a missing license but that is not related to the essence of the change

Asteliks:
@andrewharvey @grischard 
Hi, he seems to be unavailable. Can you guys help me out?

tyrasd:
FYI: It would have been nice to first translate all strings into English before requesting translators to translate it into other languages. Especially the attribution strings are causing confusion now, as people are getting asked to translate text from Polish into Norwegian for example: https://www.transifex.com/openstreetmap/id-editor/translate/#no/$/195041368

//edit: One solution would be to translate the attribution strings of all Polnish layers with `"i18n": true` to English. Would that be possible?

PR #1281, opened 2021-10-05
Update PEMA imagery

cicku:
This commit adds some information recommended by PEMA Digital Director Katie Schroeder.

### DIFF
--- a/sources/north-america/us/pa/PEMA_Orthoimagery.geojson
+++ b/sources/north-america/us/pa/PEMA_Orthoimagery.geojson
@@ -1,13 +1,14 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "PEMA_Orthoimagery_2018_2020",
+        "id": "PEMA_Orthoimagery",
         "attribution": {
-            "url": "https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=5158",
-            "text": "PASDA",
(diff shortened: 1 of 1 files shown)


andrewharvey:
The attribution and icon changes are good. Can I ask though, because the ESRI service is still `PEMAImagery2018_2020` is this imagery updated beyond 2020 or will it always be 2018 to 2020 imagery? If there will be a new layer for 2021+ imagery then I think it's clearer to leave this one labelled as 2018-2020. However if this layer will be updated for new imagery then your changes look fine.

cicku:
https://www.pema.pa.gov/911-Program/Partners/NG911-GIS/Aerial-Imagery/Pages/FAQ.aspx

Yes. I thought about this. PEMA should roll out some new imageries by the end of the year per contract. The current acquisition cycle is 2018-2022. Few Central PA counties told me to wait for the update.

I just wonder if old_id is needed, otherwise I’m open to keep it as is because following PASDA page is nothing wrong here.

andrewharvey:
Usually either a provider will,
1. publish different tile services for each capture (eg. 2018, 2020, 2022 etc) or,
2. just have one tile service that composites all with latest on top (so you can't go back and compare imagery)

If they do 1, then it's useful to have all of these in the index with different ids, but if they do 2 then we can only have one and it makes sense to remove the year from the name/id and instead rely on the start/end data metadata.

cicku:
I’m 100% confident that PASDA hosts the data as option 2.

andrewharvey:
In that case, this change makes sense.

PR #1282, opened 2021-10-06
Minor changes to NY imagery

cicku:


cicku:
The strict check can be safely ignored because of the sea area.

PR #1283, opened 2021-10-06
Remove useless TX NAIP imageries

cicku:
These NAIP layers should not be added. Instead, mappers are recommended to use NAIP imagery directly via https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/north-america/us/USDA-NAIP.geojson.


rbuffat:
It is quite common to have older imageries also included in the index.  Old imagery can be useful to understand why something was mapped in the past if something has changed and the current imagery does not correspond with the mapped data. There is the risk that people use old imagery while new imagery is available, but I think the risk for this is rather low when the date of the imagery is visible in its name.  Naturally, really old imagery should be removed.

cicku:
These NAIP imageries are never considered as reliable because of the 1 meter resolution.

If you really need to include them, consider including certain acquisition areas of specific year. NAIP does not only update Texas in 2018, 2017, 2016 or even older. All other states within the acquisition area should be included as well.

PR #1290, opened 2021-10-11
europe/ch: switch from wms to tms (wmts with epsg3857)

rbuffat:
Switch swisstopo layers to tms (using a wmts server). This should be faster compared to wms.

CC @simonpoole


simonpoole:
Hi rbuffat there are two issues with the change, one minor there is no guarantee that the images cached from the WMS server have the same tile size as those from the wtms server (for example vespucci gets 512x512 pixel images for efficiency purposes), unluckily there is no way to indicate to users that they should clear the cache.

The bigger issue is that the max zoom is set to 18, if this is the maximum zoom swisstopo has tiles for, this is a significant resolution reduction, as the nominal resolution for the images is supposed to be 10cm/pixel which is roughly zoom level 20. In testing this is very noticeable (typical editing zoom levels are 21-23, so your typical editor will do in app overzoom). Naturally if there are higher zoom level tiles we can set the max zoom to whatever it is.

rbuffat:
@simonpoole Thanks for looking into it! 

I was not aware of the cache issue. The issue here is if the server changes but the id stays the same? Thus changing to a new id would solve this issue (this is also relevant for other imagery, I wasn't thinking about this problem). 

The max_zoom is a good point. The swissalti/swisssurface datasets have a resolution of 0.5 m (the server has a max_zoom of 18), swissimage has a resolution of 0.1 m (max_zoom 20).  For Berne, max_zoom 18 should correspond to 0.408 m / pixel, max_zoom of 20 to a 0.102 meter / pixel. (I'm already a bit tired, could be that I made an error in the script). Thus, switching only the swissalti / swisssurface layers should not result in a loss of resolution? 

```
Zoom: 10                m/Pixel (x): 104.761            m/Pixel (y): 104.116
Zoom: 11                m/Pixel (x): 52.264             m/Pixel (y): 52.059
Zoom: 12                m/Pixel (x): 26.132             m/Pixel (y): 26.059
Zoom: 13                m/Pixel (x): 13.066             m/Pixel (y): 13.022
Zoom: 14                m/Pixel (x): 6.533              m/Pixel (y): 6.513
Zoom: 15                m/Pixel (x): 3.267              m/Pixel (y): 3.256
Zoom: 16                m/Pixel (x): 1.633              m/Pixel (y): 1.628
Zoom: 17                m/Pixel (x): 0.817              m/Pixel (y): 0.814
Zoom: 18                m/Pixel (x): 0.408              m/Pixel (y): 0.407
Zoom: 19                m/Pixel (x): 0.204              m/Pixel (y): 0.204
Zoom: 20                m/Pixel (x): 0.102              m/Pixel (y): 0.102
Zoom: 21                m/Pixel (x): 0.051              m/Pixel (y): 0.051
Zoom: 22                m/Pixel (x): 0.026              m/Pixel (y): 0.025
Zoom: 23                m/Pixel (x): 0.013              m/Pixel (y): 0.013
```

Script: 

```
import mercantile
from pyproj import Transformer, transformer
from pyproj.crs import CRS

pixel_tile = 256.0

crs_from = CRS.from_string("epsg:4326")
crs_to = CRS.from_string("epsg:2056")

transformer = Transformer.from_crs(crs_from, crs_to, always_xy=True)

for zoom in range(10, 24):
    bds = mercantile.bounds(mercantile.tile(7.43866, 46.95110, zoom))
    tl = list(transformer.transform(bds.west, bds.north))
    br = list(transformer.transform(bds.east, bds.south))
    pixel_x = (br[0] - tl[0]) / pixel_tile
    pixel_y = (tl[1] - br[1]) / pixel_tile
    print(f"Zoom: {zoom}\t\tm/Pixel (x): {round(pixel_x, 3)}\t\tm/Pixel (y): {round(pixel_y, 3)}")
```

simonpoole:
> @simonpoole Thanks for looking into it! 
> 
> I was not aware of the cache issue. The issue here is if the server changes but the id stays the same? Thus changing to a new id would solve this issue (this is also relevant for other imagery, I wasn't thinking about this problem). 

The problem is that changing the ID makes the cache for the layer unreachable, at least for Vespucci, but I suspect it might be a general issue. This could be fixed with a pass over the existing caches at startup, which might be a good idea in any case.

> The max_zoom is a good point. The swissalti/swisssurface datasets have a resolution of 0.5 m (the server has a max_zoom of 18), swissimage has a resolution of 0.1 m (max_zoom 20).  For Berne, max_zoom 18 should correspond to 0.408 m / pixel, max_zoom of 20 to a 0.102 meter / pixel. (I'm already a bit tired, could be that I made an error in the script). 

Maybe I was too tired and mixed up the layers?

simonpoole:
> Maybe I was too tired and mixed up the layers?
Yeah, was looking at the wrong layer, zoom 20 is naturally fine. I think for this iteration keeping the id the same makes most sense, but it is something we should look at going forward.

rbuffat:
@simonpoole I just noticed that the WMTS server includes the swisstopo watermark for the swissimage dataset. There seems to be no watermark for the hillshades. Also, the max_zoom for these layers should be larger than the imagery resolution. What do you think about just switching them to WMTS?

The reason for this PR was that I noticed that in rare cases with iD not all tiles are loaded. I suspect this is due to too many requests to the WMS server and I then noticed that there is WMTS server with epsg:3857.  Also, I assume that WMTS server should be a bit faster (but actually I haven't tested it, I just assume this because of how WMS/WMTS works).  But as the WMS server works too, there is no need to merge this if it causes more issues than it solves!

simonpoole:
I've noted the "slow" part of that though the WMTS doesn't seem to be much better (I suspect overload by the usual suspects scraping). IMHO no problem with switching. I'll talk to our contacts at swisstopo about the watermark, that doesn't seem to make sense if they want people to preferably use the WMTS server.

Issue #616, opened 2019-01-19
Franklin County, Ohio


1ec5:
The [City of Columbus GIS agency](https://www.columbus.gov/Templates/Detail.aspx?id=68426) publishes 3-inch orthoimagery of Franklin County, Ohio, that was collected as part of the Ohio Statewide Imagery Program (#593).

## URL of the imagery
http://maps.columbus.gov/arcgis/rest/services/Imagery/Imagery2017/MapServer

## Two letter country code
US

## Desired layer name
Columbus 2017 Orthophotography

## License
Public domain
(as part of OSIP)

## Is the license compatible with OSM?
Yes

## Attribution (text, URL, required?)
Columbus GIS, [Ohio Statewide Imagery Program](https://ogrip.oit.ohio.gov/ProjectsInitiatives/StatewideImagery.aspx)
(not required)

## Privacy policy (URL)
https://www.columbus.gov/Privacy/

## Icon
[<img src="https://www.columbus.gov/favicon.ico" width="16" alt="">](https://www.columbus.gov/favicon.ico)

## Bounding polygon
[GeoJSON](https://gist.github.com/1ec5/83834773000d2949973d34d8fefa18b1)

The coverage area overlaps with OSIP (#593).

EarthAdam:
According to one of the GIS Analysts at the City of Columbus, OSIP imagery for the City of Columbus is public domain.
![image](https://user-images.githubusercontent.com/8453197/51555828-8cc86c80-1e46-11e9-91f8-2d2d84f6fe1f.png)
They couldn't say for anything outside of the city though as that's governed by OGRIP.

1ec5:
Fixed in #1298.

PR #1298, opened 2021-10-14
Add Franklin OH orthoimagery

cicku:
Public domain.


sources/north-america/us/oh/Franklin_OH_2019.geojson
```diff
@@ -0,0 +1,280 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "Franklin_OH_2019",
+        "attribution": {
+            "url": "https://auditor-fca.opendata.arcgis.com/",
```
1ec5, line 6: #616 suggests an icon for attribution, not required though.

cicku: That logo in #616 appears to be off-topic with Franklin County...

The one used here is perfect.  
image

cicku: https://github.com/osmlab/editor-layer-index/commit/e4232ebda55d59b9a8a05e08de0d9b42c91f4af3

1ec5: Thanks, that makes sense. The same imagery is available from the City of Columbus, hence the confusion.

PR #1303, opened 2021-10-18
Add Miami-Dade FL imageries

cicku:
Per Florida State Law, all information published on the government website are public domain.

I've confirmed this with the county, hopefully I won'be questioned again here.

image


sources/north-america/us/fl/Miami-Dade_Ortho_2021.geojson
```diff
@@ -0,0 +1,266 @@
...
+        "type": "wms",
+        "available_projections": [
+            "CRS:84",
+            "EPSG:4326",
+            "EPSG:6438"
+        ],
+        "start_date": "2021",
+        "end_date": "2021",
+        "description": "The 2021 orthoimagery for Miami-Dade County of the State of Florida",
+        "category": "historicphoto",
```
andrewharvey, line 27: is this 2021 imagery historical?

cicku: My thought was because they have a most recent layer, all other layers of specific year should be considered as historical. The county has adequate funding which can be used to update aerial every year.

Correct me if I’m wrong.

andrewharvey: I can't see any other us/fl imagery already in the index, which is the more recent layer?

cicku: It’s in this PR…

andrewharvey: ah I see, thanks.

PR #1307, opened 2021-10-19
Add Oakland MI orthoimagery

cicku:


andrewharvey:
Assuming this is public domain

PR #1310, opened 2021-10-20
modifying and verifying privacy policy urls for sources/world

jeffreyameyer:
Not sure why the head repo here is the OHM ELI, but submitting against gh-pages in the hopes all is actually usable...


jeffreyameyer:
Also... included a few entries of '"privacy_policy_url": "false"' in anticipation of support for this in an updated strict checker...

andrewharvey:
> sources/world/OpenPTMap-overlay.geojson → sources/world/OpenPTMap-overlay.geojson.del 
 
 Did you mean to rename that file or delete it?
 
 > sources/world/.OpenTopoMap.geojson.swp 
 
 shouldn't be commited

andrewharvey:
Apart from my comments, everything else looks good! Thanks.

> Also... included a few entries of '"privacy_policy_url": "false"' in anticipation of support for this in an updated strict checker...

We need to document if this should be `"false"` or `false` or `null`. I think `false` is best so it can be String or Boolean. Though I'm happy to circle back to these later once the schema and documentation has been updated.

rbuffat:
@jeffreyameyer Thanks for searching for the privacy policies! The situation that there is no way to document that there is none is not satisfying.  I'm a bit hesitant to introduce something that might break something for a data consumer without any notification. Would it be an option to just leave the `'"privacy_policy_url": "false"'` out for now?

andrewharvey:
Agreed @rbuffat we can leave it out for now, and follow up with a separate schema change giving data consumers notice/checking with them about using `false` to indicate there is no privacy policy.

PR #1314, opened 2021-10-24
Update Indiana imagery

cicku:
Now reading the imagery directly from the state instead of the university to have a seamless experience. The state already has integrated data from 2019.

Next cycle data starting from 2021 will be pushed by the end of the year.


andrewharvey:
Can the older imagery still be retained or is it not available anymore? Older dates can be useful for mappers to see what's changed and when.

cicku:
Sorry I just got back to work and trying to catch up here. Old imageries can be found on Indiana University website as usual.

The state updates 1/3 of the state each year. The imagery proposed here is a latest available composite of all three in the past + 2019 new acquisition. It is not a statewide imagery with all area updated in 2019. Using this imagery will also reduce the work on JSON polygon maintenance (`891 lines` vs `3435 lines`).

I understand your concern and would see if keeping 2016-2018 is still needed.

cicku:
I'm going to close this one and keep 3 others whilst adding it as "Latest". Let 2016-2018 be gone when the imagery program enters the second cycle.

PR #1318, opened 2021-10-29
Add more VCGI imageries

cicku:
- LiDAR + Infrared imageries of the state of Vermont
- Put them underneath subfolder vt/


andrewharvey:
I'll merge this because it looks like the errors are due to a bad ssl library in the checker, as accessing directly via the web works for me.

PR #1326, opened 2021-11-05
add TMS-compatible variant of pl/Geoportal 2 "latest" imagery

### LINKED ISSUE #1324: Geoportal 2 imagery: Longer loading times than expected
@rbuffat @andrewharvey 

Something seems to be wrong with iD editor. In Poland the Geoportal 2 layers are loading painstakingly long. At first I thought that this is something with the provider but the exact same layers load instantly in JOSM. Thus I'll need some help concerning this topic.

tyrasd:
closes #1324


Asteliks:
I see, your already doing this. Just to make it simpler note that there are other layers form the geoportal server used here.
Here you can find all the available sources: https://www.geoportal.gov.pl/uslugi/usluga-przegladania-wms

tyrasd:
for reference: https://github.com/osmlab/editor-layer-index/issues/1324#issuecomment-961787952 explains why this PR only addresses the "latest" imagery layer.

@Asteliks feel free to work on this further if you feel it is necessary also for the rest of the layers.

Asteliks:
> for reference: [#1324 (comment)](https://github.com/osmlab/editor-layer-index/issues/1324#issuecomment-961787952) explains why this PR only addresses the "latest" imagery layer.
> 
> @Asteliks feel free to work on this further if you feel it is necessary also for the rest of the layers.

Sure, I will. Thanks for your help. This will serve as a good example of TMS use. 

As a side note. How can I get collaborator status for this repo? I know that I can make PR without it and I already have but you know the deal...

tyrasd:
> As a side note. How can I get collaborator status for this repo?

Good question. I haven't done too much work on this repo lately, so can't tell how things are currently done here. In the past it used to be very easy to get write access, but I can see that only giving write access to users who actually are helping with reviewing/managing PR and issues can also make sense.

matkoniecz:
I propose to revert https://github.com/osmlab/editor-layer-index/pull/1326/commits/1737638a98b5a1cbb368c5e3bf197869c53352c9 - "may be old and outdated" was not misleading and is an important info

PR against this branch is on #1327

@Asteliks Why it was supposed to be misleading?

tyrasd:
> I propose to revert 1737638 

I think we should work on a kind of "compromise" solution here since the current title does IMHO also have its flaws (copying some of my comments from https://github.com/osmlab/editor-layer-index/commit/f9ce24c6104e963c5b60ad49ecc6db82d799e78d#commitcomment-59529745):

* For other composite layer which include imagery of varying vintage (e.g. Bing) where we also don't put a warning in the layer's title.
* Also, at least in iD there is not enough space to display the warning/description part of the title: 
    image
* The description of the layer (in Polish) already mentions the fact that the layer does not show the "latest" available imagery. This description is shown by e.g. by iD as a tooltip.
* //edit: FWIW, JOSM also doesn't include such a warning in the layer name: https://josm.openstreetmap.de/wiki/Maps/Poland#Geoportal2:HighResolutionOrthophotomapaerialimage

Especially the second point above needs to be addressed if the consensus is to keep having a warning in the title.

matkoniecz:
> This description is shown by e.g. by iD as a tooltip.

I would expect that this info will reach approximately noone.

> For other composite layer which include imagery of varying vintage (e.g. Bing) 

The problem is that it has "high resolution" in name and is actually of high resolution, what appears to make it superior from alternative layer from the same source

> Also, at least in iD there is not enough space to display the warning/description part of the title

what about

`Geoportal 2 aerial - High Resolution, may be older`

`Geoportal 2 aerial - latest imagery`

as names?

> FWIW, JOSM also doesn't include such a warning in the layer name

In case some good name will be reached here I will edit/propose editing also JOSM layers.

rbuffat:
@tyrasd Thank you for creating this PR and leading the discussion. Except for the name change, it looks good to go. 

Regarding the name issue, I think ultimately it would be best if the Polish community can decide on this.

Personally, if there is a layer `Geoportal2 Latest (aerial image) ` and a layer `Geoportal2  Highest resolution (aerial image)` I would automatically assume that the imagery in the `Highest resolution` layer may be outdated. (Other Polish layers seem to use (aerial image)) Especially if you zoom out and see that one covers the whole country, and one only parts of it. But I can't speak for others on this matter.

In any case, it would be good to ensure that we have a Polish translation for the name and the description. See https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#translations

rbuffat:
> > As a side note. How can I get collaborator status for this repo?
> 
> Good question. I haven't done too much work on this repo lately, so can't tell how things are currently done here. In the past it used to be very easy to get write access, but I can see that only giving write access to users who actually are helping with reviewing/managing PR and issues can also make sense.

I don't have permission to add collaborators. I'm also not sure who exactly has this permission.

bhousel:
> I don't have permission to add collaborators. I'm also not sure who exactly has this permission.

~~I was just about to add him, but looks like someone beat me to it.~~
Sorry I was mistaken - I thought this was @tyrasd asking (who does have access).

tyrasd:
I've split off the name change issue into a separate PR: #1333

PR #1332, opened 2021-11-11
Update PDOK-Luchtfoto-Beeldmateriaal-25cm-latest.geojson

AllroadsNL:
New high resolution 7,5 cm aerial imagery available from PDOK for the Netherlands.

Question: 
"type": "tms", is this correct, it was already there and working.
Because: SERVICE=WMTS
Should that be "type": "wmts"?

The icon is still the same.


AllroadsNL:
Update PDOK-Luchtfoto-Beeldmateriaal-25cm-latest.geojson
The name is not correct 25cm must be  7,5cm

andrewharvey:
> Should that be "type": "wmts"?

Because the URL has parameters `{zoom}&tileRow={y}&tileCol={x}` then it's tms.

AllroadsNL:
Thanks.

How can we change the name of the geojson file in 7,5cm?

PDOK-Luchtfoto-Beeldmateriaal-7,5cm-latest.geojson

This is more correct as this is a high resolution imagery.
or us instead of 7,5cm, Ortho_HR

andrewharvey:
@AllroadsNL you can go to the committed file at https://github.com/AllroadsNL/editor-layer-index/blob/gh-pages/sources/europe/nl/PDOK-Luchtfoto-Beeldmateriaal-25cm-latest.geojson then click the pencil "Edit this file" button, then rename the file and click "propose changes" which will submit a new PR.

PR #1333, opened 2021-11-12
improve naming of pl/Geoportal 2 "high res" imagery

tyrasd:
This was split of from #1326, as this should be discussed and decided by the Polish community (see https://github.com/osmlab/editor-layer-index/pull/1326#issuecomment-962191576):

> Regarding the name issue, I think ultimately it would be best if the Polish community can decide on this.

@Asteliks can you perhaps coordinate this?

### DIFF
--- a/sources/europe/pl/Geoportal2OrthophotomapHighQuality(aerialimage)_WMS.geojson
+++ b/sources/europe/pl/Geoportal2OrthophotomapHighQuality(aerialimage)_WMS.geojson
@@ -2,7 +2,7 @@
   "type": "Feature",
   "properties": {
     "id": "Geoportal2-PL-HighResolution-aerial_image_WMS",
-    "name": "Geoportal 2: High Resolution Orthophotomap, may be old and outdated (aerial image)",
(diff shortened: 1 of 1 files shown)


Asteliks:
Every Friday the Polish community has a meeting on discord. I’ll do my best to attend today if I manage to finish my work

matkoniecz:
I created https://forum.openstreetmap.org/viewtopic.php?pid=847077#p847077 that has wider audience

matkoniecz:
And in general outcome was that there was no real support for what I proposed.

So while I still think it would be a good idea, this is clearly not shared by others.

tyrasd:
@matkoniecz So the PR can be merged as is, right?

matkoniecz:
yes

rbuffat:
@matkoniecz @Asteliks Thank you for taking the time and effort to discuss this issue and come to a conclusion! ❤️

Asteliks:
To be honest @matkoniecz did most of the work. I just reminded everyone about this topic something like last week. Thanks @matkoniecz for driving this all the way to the end!

PR #1339, opened 2021-11-14
Update Europe / Moldova imagery

Oni-DOS:


Oni-DOS:
Is the "available_projections" parameter important?

rbuffat:
@Oni-DOS Yes, as editors can choose to ignore sources if no supported projections can be found. There is a tool that helps to create the correct WMS parameters:

https://mybinder.org/v2/gh/rbuffat/eli-helper/master?urlpath=apps%2F/create_wms_template_appmode.ipynb

You should be able to paste the URL (including the parameters in {}) in the form and then you have to enter the number of the layer you want so select.

PR #1351, opened 2021-11-23
change category of Waymarked Trails layers to "osmbasedmap"

tyrasd:
I'm not sure why the Waymarked Trails overlays got categorized as `other` in #761, but I think `osmbasedmap` fits better.

@lonvia @simonpoole @grischard @Klumbumbus – what do you think?


simonpoole:
Fairly sure this was to not create a diff to the JOSM categorization.

tyrasd:
My question was more about your opinion on the change itself. A quick :+1: / :-1: would be nice. :blush: 

PS: I would (of course) also propose to change the categorization on the JOSM side as well.

grischard:
No one has complained, so merging :)

PR #1352, opened 2021-11-26
Set best to false of NJ imagery

cicku:
https://github.com/openstreetmap/iD/issues/8777


bhousel:
Thanks! This was driving me nuts too..

bhousel:
Actually might be better to just remove the `best` prop.  I don’t know what `best: false` will actually do.

cicku:
I noticed that Go Map is having trouble understanding best = false though.

rbuffat:
We could drop `best: false` in the script that generates the imagery.geojson?

PR #1354, opened 2021-11-29
proposal to add instructions about the icon and category props

tyrasd:
Addresses suggestions by @tordans in https://github.com/openstreetmap/iD/pull/8826#issuecomment-980842315.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,28 @@ To specify imagery taken sometime in 2019, use `"start_date": "2019"`,
 Implementations *must not* round down the end date (e.g. consider `2013` the same as the
 start of `2013`. Note that this is the opposite of what we did before, and layers before 2015 could have overly wide date ranges.
 
+
+##### Categories
+
+Sources should be classified into one of the following categories of different types of layers:
+
+* `photo` – For satellite and aerial imagery.
+* `historicphoto` – Same as above, but for `photo` layers which are either quite old[^1] or when there exists a more recent `photo` source for the same region from the same imagery provider.
+* `map` – A map which is not mostly based on OSM data.
+* `historicmap` – Same as above, but for `map` layers which are either quite old[^1] or when there exists a more recent version of the map for the same region from the same provider.
+* `elevation` – A map focused on displaying terrain cartography, such as elevation contour lines or shaded relief maps.
+* `osmbasedmap` – A map which is mostly based on OSM data.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -71,6 +71,28 @@ To specify imagery taken sometime in 2019, use `"start_date": "2019"`,
...
+
+Sources should be classified into one of the following categories of different types of layers:
+
+* `photo` – For satellite and aerial imagery.
+* `historicphoto` – Same as above, but for `photo` layers which are either quite old[^1] or when there exists a more recent `photo` source for the same region from the same imagery provider.
+* `map` – A map which is not mostly based on OSM data.
+* `historicmap` – Same as above, but for `map` layers which are either quite old[^1] or when there exists a more recent version of the map for the same region from the same provider.
+* `elevation` – A map focused on displaying terrain cartography, such as elevation contour lines or shaded relief maps.
+* `osmbasedmap` – A map which is mostly based on OSM data.
+* `qa` – A map or overlay with the main intent
```
simonpoole: .. with the main intent ??? the end of the sentence is missing here.

tordans:
Thanks for adding this documentation! For me, that makes it a lot clearer already.

---

Should we maybe reference this section at https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L46-L59. Something like.

```diff
- "description": "A rough categorisation of different types of layers",
+ "description": "A classified of sources. See https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#categories",
```

---

I think this would be a good time to reconsider, if the two `*historic` categories are really needed. My take is, that ELI should only have useful maps. If a map is not useful anymore, we should remove it. If it is still useful, the capture-year is a much more important value to have. Editor can always sort, filter or limit/cap based on the capture-year.

Can we maybe just re-categorize those layer to remove the historic?

grischard:
I think "a classified of sources" is missing at least one word @tordans?

The categories are the way they are to stay compatible with JOSM. Any changes to them have to be agreed with the JOSM maintainers first. Editors can and do filter on capture years and categories.

CONTRIBUTING.md
tyrasd: oops, the rest of the sentence was missing (and the description for `other`), sorry for the oversight. It should be fixed now.

tyrasd:
> Should we maybe reference this section at https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L46-L59

:+1: I've added the link in 7a73e7f

PR #1363, opened 2021-12-07
fix South Tyrol hillshade layers

tyrasd:
Apparently the WMS endpoint was changed to a new URL and the layers were renamed.

This also fixes the incorrect vintage of the 2.5m resolution "lowers" DTM/DSM layers, which has data from 2006 (not 2013).

### DIFF
--- a/sources/europe/it/SouthTyrol-Hillshade_DSM-2006-lowres.geojson
+++ b/sources/europe/it/SouthTyrol-Hillshade_DSM-2006-lowres.geojson
@@ -27,10 +27,10 @@
         "country_code": "IT",
         "type": "wms",
         "category": "elevation",
-        "start_date": "2013",
-        "end_date": "2013",
-        "id": "South-Tyrol-DSM_2013_2_5m",
-        "name": "South Tyrol DSM Hillshade 2013 (2.5 m)"
+        "start_date": "2006",
+        "end_date": "2006",
+        "id": "South-Tyrol-DSM_2006_2_5m",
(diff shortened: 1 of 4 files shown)


rbuffat:
Cool! The `https://geoportal.buergernetz.bz.it/geodaten.asp` link is not really meaningful, as it currently just states `Falls nicht anders angegeben, werden die Daten des Geoportal Südtirol mit der Lizenz "Creative Commons Zero" (kurz CC0) ausgegeben.` Would it be possible to use the PDF links from the geokatalog as license url, e.g. http://geoservices.buergernetz.bz.it/geonetwork/srv/ger/pdf?uuid=p_bz:Elevation:DigitalTerrainModel-2.5m ?

tyrasd:
> The https://geoportal.buergernetz.bz.it/geodaten.asp link is not really meaningful

Yeah, I thought that as well. But it is the only place where the actual license is mentioned. The PDFs don't really contain any license information at all for the CC0 layers, except if you interpret the `gmd:useLimitation – no conditions apply` line as such?! //edit: I've changed the links to the metadata pdfs now

rbuffat:
@tyrasd True. If I remember correctly they had previously included more license information. `gmd:useLimitation – no conditions apply` is also not optimal. It's best if you decide!

tyrasd:
let's keep the PDF's. There are enough ways to find the https://geoportal.buergernetz.bz.it/geodaten.asp page, while the metadata PDFs are otherwise well hidden.

PR #1368, opened 2021-12-12
Update URL of Hamburg DK5 layer

hauke96:
The LGV (operator of this layer) recently changed the layer name from "1" to "DK5".


rbuffat:
Thanks @hauke96 !

PR #1370, opened 2021-12-13
Updates to NYSDOP_Latest

ZLima12:
The 2021 imagery batch is now live, and these changes will enable it.

Additionally, I noticed that this repository requests tiles as JPEGs, whereas JOSM requests PNGs with transparency (see [here](https://josm.openstreetmap.de/wiki/Maps/USA%20States)). I went ahead and switched this to the transparent PNGs, which should be advantageous.


cicku:
Please do not switch to PNG, it pulls more traffic.

ZLima12:
Alright then, I'll drop that commit.

cicku (approved):

PR #1377, opened 2021-12-16
Refactor libeli and check scripts

rbuffat:
Work in progress. It will need some time until it is finished. 

Refactor libeli and scripts to hopefully increase the maintainability and readability of the code. 

- Use type hints wherever possible
- Introduce helper modules to parse WMS and WMTS Capabilities as well as TileMapResources.
- Port strict_check.py and wms_sync.py to new helper modules
- New check for WMTS if layers could be added as tms. URL suggestions are given, e.g. : 
```
Layer ch.swisstopo.swisstlm3d-eisenbahnnetz could be represented as type 'tms' and the URL: https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swisstlm3d-eisenbahnnetz/default/current/3857/{zoom}/{x}/{y}.png

```
- Format everything with black. Add pyproject.toml for configuration.
- Use pip-tools to configure Python requirements
- Add unit tests for libeli using pytest
- Add dependabot to update Python dependencies and github actions
- Encode XML according to `encoding=X` XML attribute. 
- wms_sync: Remove transparent=true for image formats that are not gif or png (other formats do not support transparency AFAIK)


rbuffat:
Should be "finished".

@grischard  @cicku @Asteliks @willemarcel I assume there will be some issues I'm not yet aware of. Please be extra cautious if the script fails for PRs as it can be due to not yet detected bugs.

Resolves https://github.com/osmlab/editor-layer-index/issues/1213

PR #1385, opened 2021-12-20
Add rights of way tiles for GB

systemed:
These tiles are generated from the Open Government Licence data collated at rowmaps.com by Barry Cornelius.


systemed:
Anything I can do to help get this merged?

boothym:
Any chance Scotland could be removed from this polygon? AFAIK the ROW maps only cover England and Wales.

grischard:
Yes, you can edit [the geojson](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/gb/LocalAuthorities_RightsOfWay.geojson) in josm or on geojson.io and submit it as a pull request here.

The geojson poly is way too detailed for the imagery boundary, and the many points over the sea, where there's no mapping, can be simplified.

grischard:
Actually I did it in #1445

PR #1390, opened 2021-12-22
Create IDECA-orthophotomosaic.geojson

angoca:
One imagery for Bogota, Colombia


rbuffat:
@angoca Sorry for the long delay! I had a quick look at your PR. I have some comments. `"license_url"` typically points to an URL where there is information of what license the data is and not directly to a license. This allows other people to easily verify the license of the data. 

If the license of the imagery is CC-BY, then the license requires attribution. Thus, we need to ensure that there is an entry attributing that contributions to OSM used this data in https://wiki.openstreetmap.org/wiki/Contributors and we need to document that the data owners are fine with attributing the use of their data in this way.

angoca:
Ok, I corrected the link, now it points to the documentation where it says the documentation.

angoca:
I already changed the link.

arch0345:
As @rbuffat mentioned, we need to confirm with IDECA that listing them on the Contributors page is sufficient for attribution under CC BY 4.0. This can be done with a waiver: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

PR #1401, opened 2022-01-05
Add be/Wegenregister

pietervdvn:
This pull requests adds a new source for Flanders (Belgium).

Wegenregister is a source containing roads and paths in Flanders and Brussels. They are published by 'Agentschap Informatie Vlaanderen' (which also publish the GRB and aerial imageries for this region which already are in the ELI).

The license is the same as GRB and already existing sources.

I've reused the outline from one of the Hillshade-files.


rbuffat:
@grischard could you have a look at this PR?

grischard:
I will probably tomorrow @rbuffat

grischard:
Merged, thank you and sorry for the delay.

@pietervdvn does the footprint really extend outside of Flanders for a few km, or is that just because you reused the hillshade footprint?

pietervdvn:
@grischard I reused another outline, so I should check the actual outline.

grischard:
@pietervdvn for the coast you can then pull a straight line through the sea - no one maps in the water so it doesn't matter

PR #1402, opened 2022-01-06
Update Budapest XI. district

urbalazs:


rbuffat:
@urbalazs Thanks for your PR and sorry for the long delay. Could you give more information about the license of this dataset? The translated page https://wiki.openstreetmap.org/wiki/WikiProject_Hungary/Ortofotók#Budapest_XI._ker.C3.BClet seems only to state that somebody? got the information about the license of that data but not what license the data actually is. 

> We have known about the public GIS system of Újbuda for a long time, we contacted the Local Government, but we did not receive an answer for a long time. Finally, we received the license to use Cartographia / funiQ.hu , and we thank you again.

urbalazs:
@rbuffat nothing has changed since the [last pull request](https://github.com/osmlab/editor-layer-index/pull/736). What else do you need for the current one?

rbuffat:
> What else do you need for the current one?

What license is this data and is this license compatible?

Asteliks:
@urbalazs could you update your PR with a link to the proper license used for this service? We need implicit information that we may use this source before merging

grinapo:
> What license is this data 

The data is licensed to the OSM community to be used for adding/updating OSM objects (in legal lingo "to create derivative works based on it to be used for updateing OpenStreetMap data"). It cannot be used for anything else. It shall be attributed as the wiki page states.

> and is this license compatible?

It is compatible with OSM ODbL since it allows to create derivative works and does not limit the usage of those, apart from the attribution tagged on the objects. It is not compatible with any random OSA license since the data cannot be used for anything else outside adding to OSM.

This is pobably the same as almost all orthographic photo source since they all want to prevent people to use the imagery for random purposes.

rbuffat:
If it is a requirement to add tags on modified objects, then I see it critical to include this imagery here as I'm not aware of a mechanism that this would be enforced by OSM editors.

AFAIK the only place to acknowledge used data is https://wiki.openstreetmap.org/wiki/Contributors
But I'm happy if somebody with more experience could prove me wrong. 
@grischard @simonpoole

grinapo:
> If it is a requirement to add tags on modified objects, 

No. We asked how to attribute them and they suggested a text. That's all what's been written.

We tag this text on objects, or changesets, which is our way to give credit. 

> AFAIK the only place to acknowledge used data is https://wiki.openstreetmap.org/wiki/Contributors But I'm happy if somebody with more experience could prove me wrong.

I can link the Hungarian permission page if anyone insists but it's in Hungarian and I see no good way to keep it in sync in a translation, so I rather wouldn't create disjunct language versions.

BathoryPeter:
I have copies of the permission request emails. Tagging was not a requirement of use. I'm copying the machine translated version of the permission request here:

> We would like to obtain the notary's permission for the following.
>
> We would like to request permission to use *orthophotographic layers* visible in the district's public GIS system for the purpose of creating and specifying map objects for the volunteer map editing community of *OpenStreetMap*.
>
> OpenStreetMap is currently created on the basis of much lower resolution satellite images, gps tracks and personal tours. It would be a great help for more detailed and accurate mapping if we could use the orthophotos available at http://terinfo.ujbuda.hu/. OpenStreetMap builds the map with volunteer work, so it does not ask for money for the work or the end result, they can use it for free.
>
> We would only need a written consent from the municipality or the rights holder of the orthophotos.

And the answer:
> Dear xxx, I give permission.
>
> dr. Kremzner Zsuzsanna
> notary

Issue #1340, opened 2021-11-15
Indicate that no privacy policy is available


rbuffat:
Currently, it is not possible to indicate that there is no privacy policy: for sources without a `"privacy_policy_url"` there can be either no privacy policy or nobody bothered to look if there is one.

There are different possibilities to define that "somebody looked but there is no privacy policy":
1. Using `"privacy_policy_url": null` for cases when somebody looked for a privacy policy, but none was found. 
2. Using a defined keyword that does not start with "http" to define that no privacy policy was found, e.g. `"privacy_policy_url": "none"` or `"privacy_policy_url": "none_found"` or `"privacy_policy_url": "false"` etc.

A missing `"privacy_policy_url"`would then indicate that nobody looked yet for a privacy policy.

I don't think it matters too much what is chosen, it is just important that there is an agreement on which way is chosen and all data consumers are aware of this. 

@simonpoole @tyrasd @bhousel @bryceco @Klumbumbus @don-vip @grischard @andrewharvey

andrewharvey:
I would have thought `"privacy_policy_url": false` is best, it's the most explicit declaration there is no privacy policy, using `"privacy_policy_url": null` could be considered the same as no `privacy_policy_url` key in the Object, which is "nobody looked yet for a privacy policy".

tyrasd:
+1 for `false`: to me it seems to be both quite intuitive and concise. But I can also very well live with any other solution.

rbuffat:
I created a PR that implements `"privacy_policy_url": false` to indicate that there is no privacy policy available: https://github.com/osmlab/editor-layer-index/pull/1414. One thing to discuss is that with this approach privacy_policy_url is either a string or a bool. For untyped languages such as javascript or python, this is not an issue. But I'm not sure how much this will be an issue/break things for typed programming languages such as Java. @simonpoole what would be your perspective regarding Vespucci?

bryceco:
I can accept false, null or empty string. There’s no issue for typed languages.

simonpoole:
@rbuffat I noticed that @systemed wasn't included in the mentions, afaik P3 uses ELI too.

Due to the wonders of google not actually implementing what they say their methods do in their GSON library, using the (boolean) false will not cause any issues than displaying _false_ instead of an URL, which is bearable behaviour. So, OK with me.

rbuffat:
> @rbuffat I noticed that @systemed wasn't included in the mentions, afaik P3 uses ELI too.

[...]

PR #1414, opened 2022-01-19
Allow false for privacy_policy_url

rbuffat:
Implements https://github.com/osmlab/editor-layer-index/issues/1340


cicku:
Looks good.

rbuffat:
@cicku  Thanks for looking at this PR.

PR #1420, opened 2022-01-25
Latvia, orthophoto 2008: add new map

soshial:


rbuffat:
Thanks @soshial !

soshial:
Recently, Latvian cadastre map has been published under open data lisense. This WMS service has a lot of layers, hence it's not really possible/convenient to add only one layer. @[rbuffat](https://github.com/rbuffat), may I create a layer with [GetCapabilities URL](https://lvmgeoserver.lvm.lv/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities&layer=publicwfs%3AKadastra_karte)?

rbuffat:
Theoretically yes, but then only a few osm editors will be able to use the data. The WMS GetMap request allows to include multiple layers using `LAYERS=1,2,3,4,5,6&STYLES=,,,,,`.  Maybe this is an option?

soshial:
Thanks, I managed to do it in this PR: https://github.com/osmlab/editor-layer-index/pull/1449

PR #1423, opened 2022-01-26
Add Marion County Imagery 2021

arch0345:
This PR adds the newest imagery for Marion County, Oregon. I've gotten permission from Marion County GIS for [this imagery](https://marioncounty.maps.arcgis.com/home/item.html?id=d37104a0d271405d82a21184948650a7) to be used to improve OSM and included a pdf copy of the email. I simplified the geometry a bit, but let me know if this should be simplified further.


cicku:
I'm afraid that 99% of the imageries from Eagleview (Pictometry) cannot be added, especially for those still hosting at `pictometry.com` but not at county.

In most cases, Eagleview will sign the agreement with county that the aerial imagery can be used **ONLY** on county website and services. That's the reason you got the reply "Marion County owns the rights to display those layers in our maps and applications", they are right. However, Marion County did not mention that if they have the permission to share the imagery with OSM, especially sharing something not even hosted on their server, and I'm afraid the answer is **NO**.

To be honest, if Eagleview is really friendly to us, I should have added tons of imageries from them including this one. If you really wish to work on this, you need to contact Eagleview and get a waiver. Good luck if you can get yourself out of their backlog, it is been half a year and I have received no updates with their legal team.

arch0345:
Just got a reply from Pictometry, and unfortunately it sounds like we don't have permission to use this, see their reply below. Because of this, I'm going to close this PR.

> No, you cannot use our WMTS endpoint in your own applications to perform your feature mapping, and we are not able to sign your requested agreement.  This is not in compliance with our licensing agreement with Marion County.  They are licensed to use that service as a basemap in their public-facing applications and provide the URL to their internal departments, but they are not able to publicly provide the source URL.  We have reached out to them and requested that they modify the sharing permissions for that AGOL service.

cicku:
Not surprised at all. Thank you for the update.

Issue #1421, opened 2022-01-26
[Watchdog] Imagery "Ministerio de Agricultura, Ganadería y Pesca (WMS)": sources/south-america/ar/Ministry-of-Agriculture_wms.geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"Ministerio de Agricultura, Ganadería y Pesca (WMS)" -> [sources/south-america/ar/Ministry-of-Agriculture_wms.geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/south-america/ar/Ministry-of-Agriculture_wms.geojson)

Reason:
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.
Error: Exception: Could not parse XML.

CC contributors to this imagery:
@Zalitoar

Zalitoar:
Hi @rbuffat that service was moved or deleted. The ministry responsible of it didn't announced a replacement or a new URL. I could remove it from this index until further notice.

rbuffat:
Imagery seems to work again.

PR #1058, opened 2021-02-19
Argentina: adding new services and updating / removing old ones 

Zalitoar:
New base maps and mosaics from National Geographic Institute of Argentina added, existing files renamed, fallen services removed and old URLs updated.

License applying on sources are defined globally for all government public data (included in 'license_url' property): https://datos.gob.ar/acerca/seccion/Marco%20legal

Issue #1048, opened 2021-02-05
Argentina sources are broken


rbuffat:
The following sources are broken:
* south-america / ar / IGN orthophoto Posadas (WMTS)
* south-america / ar / IGN topographical map (TMS)
* south-america / ar / IGN topographical map (WMS)

https://www.ign.gob.ar/NuestrasActividades/InformacionGeoespacial/ServiciosOGC lists https://www.ign.gob.ar/descargas/tyc1.html as terms and conditions, but the servers report the data as CC BY 4.0: https://wms.ign.gob.ar/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities

```
<AccessConstraints>CC BY 4.0 https://creativecommons.org/licenses/by/4.0/</AccessConstraints>
```

@don-vip You seem to have looked at these datasets in https://josm.openstreetmap.de/wiki/Maps/Argentina?sfp_email=&sfph_mail=&action=diff&version=106&old_version=105&sfp_email=&sfph_mail Do you know more about the CC BY issue?

don-vip:
> @don-vip You seem to have looked at these datasets in https://josm.openstreetmap.de/wiki/Maps/Argentina?sfp_email=&sfph_mail=&action=diff&version=106&old_version=105&sfp_email=&sfph_mail Do you know more about the CC BY issue?

Sorry I don't know anything about any source outside of France. All of the modifications I do for other countries result from lookups for similar content when an URL breaks. I don't even remember this one.

ignaciolep:
Perhaps @Zalitoar could help. He tried to update the sources before (PR #676)

ignaciolep:
PR #1058

Zalitoar:
Hello world, I submitted a PR that fixes all the fallen services. 

I added privacy_policy_url and updated the license parameters also for all the services. Those new sources are highly used by the argentinean community and other folks mapping here, and provide a lot of very useful maps and photos published by the IGN (National Geographic Institute).

Thanks in advance! :)

Zalitoar:
Hi Vincent and René, thanks for your feedback.
I made a PR which fixes the broken services and replaced others very old from the IGN. 

We also made a change in those services in order to reflect in capabilities documents the appicable legal framework of the argentine government about public data, the 27275 Law ('Right of access to public information').
This law rules the access and nature of public data produced by all the government and associates (seventh article):
[...]

PR #1425, opened 2022-01-26
Removing a disabled service and translating names to Spanish

Zalitoar:
This service is no longer available. See issue #1421


Zalitoar:
We also discussed in the argentinean community (the Telegram group) the convenience of the names and it was decided that they will be better for us in Spanish and summarized.

rbuffat:
@Zalitoar Thanks for this PR. Sorry for the long wait.

-  ✓Removal of defunct imagery
-  The license for the added [Buenos Aires Data](https://data.buenosaires.gob.ar/dataset/) source is CC-BY-2.5-AR[1]. This license requires attribution. We need to ensure that there is an entry in https://wiki.openstreetmap.org/wiki/Contributors and should document that it is ok for the owner of the data to attribute the data this way.
-  Regarding the translations, have you considered translating them with i18n? I have to say I'm not 100% familiar with how it is implemented (and it might be that there is something broken at the moment) @grischard @bhousel @tyrasd maybe somebody can give more insights here. 
Based on my limited knowledge I assume there should be for each source an entry in https://github.com/osmlab/editor-layer-index/blob/gh-pages/i18n/es.yaml and each source with an entry should have the `"i18n": true` property set. Using i18n we would ensure that the imagery is accessible to more people. 

[1] https://creativecommons.org/licenses/by/2.5/ar/

ignaciolep:
Hi René,

> * The license for the added [Buenos Aires Data](https://data.buenosaires.gob.ar/dataset/) source is CC-BY-2.5-AR[1]. This license requires attribution. We need to ensure that there is an entry in https://wiki.openstreetmap.org/wiki/Contributors and should document that it is ok for the owner of the data to attribute the data this way.

Buenos Aires Data already has its attribution entry on https://wiki.openstreetmap.org/wiki/Contributors#Argentina (See "Datos Abiertos del Gobierno de la Ciudad Autónoma de Buenos Aires")

> * Regarding the translations, have you considered translating them with i18n? I have to say I'm not 100% familiar with how it is implemented (and it might be that there is something broken at the moment) @grischard @bhousel @tyrasd maybe somebody can give more insights here.
>   Based on my limited knowledge I assume there should be for each source an entry in https://github.com/osmlab/editor-layer-index/blob/gh-pages/i18n/es.yaml and each source with an entry should have the `"i18n": true` property set. Using i18n we would ensure that the imagery is accessible to more people.

I think that right now it would be better to implement the changes in this PR as it is and leave i18n for later.

By the way, I collaborate with the imagery translation into Spanish and I have seen many entries in languages other than English (for example, Polish). IMHO those layers shouldn't have the `"i18n": true` property set.

Zalitoar:
Hi @rbuffat adding to @ignaciolep response I would say there's a misconception about licensing from the publisher side, because the citiy government approved in 1998 the "Law 104 on access to public information" which says:

> "It will imply the freedom to access, request, receive, copy, analyze, reprocess and redistribute information in the custody of the obligated subjects, with the only limitations and exceptions established by this law."

As we understand, this is compatible with OSM license. 

Also, the law 27,275 "Right of access to public information" (a national law) remarks in its 2nd article the following:

> "The right of access to public information includes the possibility of freely searching, accessing, requesting, receiving, copying, analyzing, reprocessing, reusing and redistributing the information"

So, the supoused incomptaibility of licensing types between argentinean sources and the OSM license doesn't exist due that mentioned laws apply always over any other text, permission or license as creative commons in these cases, since sources are goverment offices. 
I'm not a lawyer but as far as I get it, no text or waiver can overrule or replace a law.

This topic was discused in #1058 and #1048.

Ask me any other thing you need. Thanks in advance!  :)

ignaciolep:
Hi, We haven't had any new comments since February. Could someone kindly let us know if there are any other doubt about this PR? Thanks in advance!

Zalitoar:
Hi [cicku](https://github.com/cicku) 
I saw you made commits recently. I'm wondering if you could help us with this PR?

ignaciolep:
Hello @grischard, Could you kindly help us in merging these commits? We responded to René's comments 6 months ago, but he appears to have stopped responding since then.
Thanks!

ignaciolep:
Hi René @rbuffat , I hope this message finds you well. I wanted to touch base regarding the Buenos Aires city layer license that we discussed in February 2022. I was wondering if you still had any questions or concerns regarding the license, or if you are now able to move forward with the merge. Thanks!

matheusgomesms:
I am not a lawyer, and although there is a large discussion in the mentioned PRs, I think there is some differences between Argentina and Buenos Aires cases. I also agree that federal law has more importance than anything, and that it looks to me (without much research, I admit) that the _Plan Nacional de Apertura de Datos_ is compatible with OSM.

However, CABA is explicitly saying that the data is in [CC-BY 2.5](https://data.buenosaires.gob.ar/dataset/tiles-fotografia-aerea), while the metadata of the [TMS service has no license at all](http://servicios.usig.buenosaires.gob.ar/mapcache/tms/1.0.0/fotografias_aereas_2017_caba_3857), so we can infer that this data is indeed using CC-BY 2.5, which is not compatible to OSM.

I can happily merge this PR, but I don't feel comfortable with this licensing situation. I would highly recommend sending some e-mail to CABA (datosabiertos@buenosaires.gob.ar) asking for explicit permission. Given the Open Data situation in Argentina, I think this should not be hard (in Brazil happens the same situation, we have a very good open-data federal legislation, but municipalities are still not proper following for diverse reasons).

Also, I would recommend separate PRs to not slow down the situation. Let's deal separately with updating names and adding a new layer.

ignaciolep:
Thanks for the thorough review, @matheusgomesms. Appreciate your diligence on licensing concerns. We've reached out to CABA (datosabiertos@buenosaires.gob.ar) and obtained explicit permission. You can find the permission statement [here](https://wiki.openstreetmap.org/wiki/File:Solicitud-de-permiso-para-usar-datos-de-BA-Data-en-OpenStreetMap.png) and also linked on our [contributors page](https://wiki.openstreetmap.org/wiki/Contributors#Argentina). Feel free to proceed with merging the PR. Thanks!

grischard:
@ignaciolep @Zalitoar do you know how to resolve the merge conflicts?

PR #1431, opened 2022-01-31
Ohio Statewide Imagery Program III

### LINKED ISSUE #1407: Ohio Statewide Imagery Program III
For some reason, the Ohio Statewide Imagery Program never updated their “[OSIP 6in Imagery Most Current Available](https://github.com/osmlab/editor-layer-index/blob/bd09c22eff54bbe6334c6f109de5b81420e4b8b5/sources/north-america/us/oh/Ohio_OSIP_6in.geojson?short_path=52718e8)” layer to include the OSIP&nbsp;III imagery collected in 2020 in several southeastern Ohio counties. This imagery wound up in a separate layer instead. Unfortunately, the separate layer is not WMS-enabled, so we’ll need to rely on the `exportImage` verb as in #611.

## URL of the imagery
https://geo1.oit.ohio.gov/arcgis/rest/services/OSIP/OSIPIII_6in/ImageServer/
or
https://geo1.oit.ohio.gov/arcgis/rest/services/OSIP/OSIPIII_2020_6in/ImageServer/ (seems to be identical)

## Two letter country code
US

## Desired layer name
OSIP III 6in

## License
[Public domain](https://ogrip.oit.ohio.gov/Portals/0/PDFs/OSIP%20Program%20Description.pdf#page=2)

Yes

(not required)

## Privacy policy (URL)
https://ogrip.oit.ohio.gov/privacy.aspx

## Bounding polygon
[Overpass query](https://overpass-turbo.eu/s/1fB6)

1ec5:
Added an OSIP III 6-inch layer that only includes the flyovers in 2019 and 2020, as opposed to the existing 6-inch layer that includes flyovers from 2010–2014 (OSIP II) and 2017–2018 (OSIP III). There is some intentional overlap between the two 6-inch layers. The new layer unfortunately does not have WMS enabled, so we’re relying on the `exportImage` endpoint, which does not perform as well.

I’ve updated the other two OSIP layers to reflect the fact that they’re no longer the “most current available”. Now that OSIP III has concluded, the coverage area of all these layers is stable and we can more confidently indicate their coverage areas. I’ve refined the 6-inch layers’ geometries to encompass just the counties that they cover, instead of the whole state.

Fixes #1407.


1ec5:
In the course of refining the geometries, I put together [this Observable notebook](https://observablehq.com/@1ec5/editor-layer-geometry-lab) that makes it easier to derive a simplified, ELI-compatible Polygon feature and JOSM-compatible Imagery XML entry from a list of boundary relation IDs.

1ec5:
[r278](https://josm.openstreetmap.de/wiki/Maps/USA%20States?action=diff&version=278) and [r279](https://josm.openstreetmap.de/wiki/Maps/USA%20States?action=diff&version=279) implement the same changes as this PR in the JOSM Wiki.

cicku:
✔️

Good to go.

cicku:
Is i18n really needed?

1ec5:
I wasn’t sure, but I recently had to translate a large number of layers from Polish to Vietnamese due to #1277, so I thought we were taking the same approach as the OSM community index and making as much localizable as possible. (Hopefully I understood what the Polish descriptions were trying to say.) I can remove the `i18n` flag if we aren’t ready to make so much of the project localizable.

PR #1435, opened 2022-02-01
Add MassGIS 2021 orthoimagery

gdt:
The 2019 entry was used as a starting point; the layer basically
differs from 2019 only in URL and capture date.

The licensing and privacy policies are the same as the 2019 imagery, which has already been checked/approved.


cicku:
Sorry you are late on this one (1 hour approx.).

https://github.com/osmlab/editor-layer-index/commit/6aea8f03fb6dffe70b0cbc11366cd3c7e5d18438

gdt:
Thanks - I will submit a fixup pull request as your version is missing some metadata and doesn't follow the naming scheme used by the previous imagery -- this is newer imagery but the same upstream source and workflow.

cicku:
Sure. But I don't think there is any naming scheme we need to follow here.

gdt:
Not any ELI doctrine, but people in MA are used to how we have been naming MassGIS layers and having the name mostly match except for they year will lead them to the obvious, correct, conclusion.    I should note that MassGIS themselves are not super consistent with their URL fragment names.

PR #1437, opened 2022-02-01
US/Massachusetts: Merge two MassGIS 2021 orthoimagery descriptions

gdt:
I and @cicki did this essentially simultaneously and each of us had
details the other didn't.  This merges the two versions and aligns
them to the style of the other 3 MassGIS layers (which have common
logical ancestry), both key ordering and naming.


cicku:
`coordinates` change is not needed because it does not affect the final index file. However, I do not think there is anything I need to merge except for `start` and `end` date, please explain:

1. `name` change: As the name comes from MassGIS and no one should change it. You should not name something different from official name.
image

2. `id` change:

Normally `id` should match the file name. I do not think there is any naming scheme we need to follow at ELI. ELI works as an index to serve iD primarily and you do not need to touch anything like this. For JOSM, that's completely different and we don't care because JOSM does not prefer ELI.

gdt:
If that's your view, then yes the start/end date is the big deal.

cicku:
I do not perceive that there is any "view" here. I'm just expressing my view/opinion and you are welcome to challenge it. Note that I have never said who is correct and wrong, but your PR is basically rewriting the file in a different format.

However, merging your PR simply means the ELI ecosystem has to accept the "view" which is allegedly only used by a small group of people, and is it a really needed change that will help the majority users? ELI tends to be neutral and it should just follow official names as many as possible. The way you name such in the past, `MassGIS {year} Orthos`, is problematic because:

1. The core value of OSM is to not use the abbreviation. Orthos inevitably shows that it is an abbreviation only used by a small group and the public does not know what it means. Is it Orthoimagery? Orthophotography? Ortho image web services (like MassGIS uses)? Orthostatic hypotension (WTF is that)? I'm able to see that you changed the name in your PR, but does it follow the "convention" previously used? If not, what exactly is your naming scheme here?
 
2. To be honest, `id` does not matter, I don't have trouble accepting a PR changing it. But your `id` shows something similar to the `name`, which is the above question, that is a name change really needed?

If you feel like MassGIS uses a name that you don't prefer, you can write to them and ask to clarify which one is better, maybe they will even reduce the inconsistencies. ELI takes no responsibility to make such kind of action but we are always open to PR which can really improve the project. We really appreciate your effort checking the imagery capture date, but for other changes, at least I'm not able to see the point here (I fully respect every community and as you can see I did not change the file of 2019).

cc @rbuffat for further look.

rbuffat:
First, thanks @gdt for your PR! 

- `id`: There are no other constraints for the id other than that it is unique and does not contain some special characters. Ideally, it is somewhat descriptive of the imagery. If it does not cause extra effort it makes sense to use the same id in the JOSM index and ELI. But there is no requirement for it. 
- `name`: There is no strict name scheme in ELI. Ideally, it is the most helpful name for the people actually using the imagery.  Without having any data to back this up, `ortho imagery` is more a technical term understood by the spatial community while `aerial imagery` might be more understood by non-expert users. But in the end, I guess most people will identify what the imagery is with both names.
- `formatting`:  The "only issue" with formatting changes is that the github tool to visualize differences cannot handle them. Thus it is harder to spot the real changes of a PR. As we do not require a specific JSON formatting style, this happens quite a lot.

gdt:
Thanks - I've been contemplating and not getting around to being coherent.   A few thoughts:
  * As far as I know we in MA have been trying to get things into ELI and let the semi-automated ELI->josm catch up, and it's my impression that everyone (almost?) thinks that is a reasonable workflow.
  * I agree that using the names from the imagery provider is best.  So that means, combined with issuer name "MassGIS 2021 Aerial Imagery", which is how it is.  I'm come around to agreeing with that.
  * I think there should be consistency in naming, relative to layers that come from the same publisher, so therefore we (I?) should adjust the 2019 imagery (btw, 2019 is still relevant for non-historical mapping as many things are more visible in the 2019 than the 2021).
  * Probably tax parcels and LIDAR also need renaming, to align with the "use upstream names" and "avoid jargon that normal people won't follow" doctrines
  * While id is arbitrary, I think it should be aligned with layer names.
  * I do not think any particular formatting rule is super important, but I think it's reasonably important for files that describe layers with some common origin differ (in the 'diff -u' sense) only when there is an actual semantic difference.   In prior contributions I tried to copy a file that was close and then change only what needed changing, to let people looking at the set of MA files quickly understand.  That property no longer holds with the 2021 file, and I'd like to adjust file names, whitespace scheme, wrapping geo or not, tag ordering, to match.

I understand the issue with git and commits that change whitespace.    So if I make a PR that has separate that aligns everything as above, with whitespace changes in a commit that is documented to only change whitespace, are you ok with that?  I'm happy to help, but if that's not welcome, I'd rather not start.

Separately, I'd suggest that you do have guidelines for formatting and tag ordering in new submissions, even if you don't enforce them, so that there is a definition of correct.  My memory is that it seemed obvious to me that tags should be in the same order as the schema, but that was long ago when I added LIDAR.

PR #1449, opened 2022-02-08
Latvia: add cadastral map

soshial:
Would be cool to have a designated category for cadastral maps [here](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json).


rbuffat:
> Would be cool to have a designated category for cadastral maps [here](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json).

Thanks @soshial !
I agree, this could be helpful category.

PR #1452, opened 2022-02-09
Stricht check: propose corrected geometry

rbuffat:
Follow up to https://github.com/osmlab/editor-layer-index/pull/1450
Do not only warn if the geometry is not valid but also propose a correct geometry.

### DIFF
--- a/.github/workflows/test_strict_check.yml
+++ b/.github/workflows/test_strict_check.yml
@@ -26,12 +26,12 @@ jobs:
           python -m pip install -U pip wheel
           python -m pip install -r requirements.txt
 
-      - name: Basic check
-        run: find sources -name \*.geojson | xargs python scripts/check.py
-
       # Rudimentary check if strict_check.py works by applying it to a small subset of sources.
       # This is expected to fail as sources might fail the strict check.
       # However, there should be no Python errors or obvious wrong checks present.
       - name: Strict check
         run: |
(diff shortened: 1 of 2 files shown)


grischard:
Does this take care of clockwise/counterclockwise direction?

rbuffat:
Not sure. The geometries were corrected using the following algorithm: https://libgeos.org/doxygen/classgeos_1_1operation_1_1valid_1_1MakeValid.html#ae93e2478fc05c9d331c1a0974f98f2df

I need to check if this also fixes orientation.

rbuffat:
I assume it takes care of Polygons that are converted to MultiPolygon. But it does not ensure that all existing Polygons have a fixed orientation, as Shapely's `is_valid` seems not to look at orientation:


I used the following code to convert the geometries:
```
    if not geom.is_valid:
        print("Not valid: {}".format(filename))
        reason = explain_validity(geom)
        print("Reason: {}".format(reason))
        geom = make_valid(geom)

```
But it should be fairly easy to convert all geometries to a given orientation using shapely: https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.polygon.orient

> Returns a properly oriented copy of the given polygon. The signed area of the result will have the given sign. A sign of 1.0 means that the coordinates of the product’s exterior ring will be oriented counter-clockwise and the interior rings (holes) will be oriented clockwise.

grischard:
Fantastic!

PR #1457, opened 2022-02-13
Fix NYS Orthos to accomodate for new layer IDs.

ZLima12:
Currently, this imagery source will error if you make a tile request, since the current URL asks for a layer that no longer exists. This fixes the issue.

### DIFF
--- a/sources/north-america/us/ny/NYSDOP_Latest.geojson
+++ b/sources/north-america/us/ny/NYSDOP_Latest.geojson
@@ -1379,7 +1379,7 @@
         "max_zoom": 19,
         "name": "NYSDOP Orthoimagery - Latest",
         "type": "wms",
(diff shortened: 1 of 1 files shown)


1ec5:
Even though this PR doesn't touch the feature’s geometry, [the build is failing](https://github.com/osmlab/editor-layer-index/runs/5171403501?check_suite_focus=true) on this geometry check:

> ERROR:root:39.14% of geometry is outside of the layers bounding box. Geometry should be checked
[11](https://github.com/osmlab/editor-layer-index/runs/5171403501?check_suite_focus=true#step:8:11)

cicku:
It is safe to ignore that warning. NY is a little bit special here.

grischard:
It's not a warning, it's an error. Why can't the bbox be adjusted to match the layer?

cicku:
I don't understand what you meant here. Is there any need to introduce nodes to make the imagery file larger?

image

rbuffat:
@cicku @grischard I suspect there is a bug in the script to check for the boundaries. 

I assume here we need to check the geometry against the union of all WMS layer bounding boxes and not separately for each layer. https://github.com/osmlab/editor-layer-index/blob/gh-pages/scripts/strict_check.py#L98-L125

cicku:
Of course there is :)

Issue #1446, opened 2022-02-07
How long should we keep outdated imagery in ELI


rbuffat:
**Problem:** 

For some regions, there is new imagery available in regular intervals (e.g. each year). The question is, how long should we keep old imagery in ELI. 

Keeping old imagery in ELI has some negative aspects:
- Old imagery is less relevant
- The size of the imagery.geojson is unnecessary increased if we keep not relevant imagery 
- The maintenance work is increased
- Mappers could not be aware that they are using old imagery

But having older imagery has also some positive aspects:
- Having old imagery helps people to identify what has changed if they encounter mapped features that do not correspond with current imagery

**Possible actions:**

- There is a tradeoff between keeping old imagery in the index. Are there some opinions about general recommendations we should adopt? E.g. not more than 3 imagery for the same regions? Not older than X years after the most recent imagery?

- If there is a start / end date set in the imagery, it should be possible to automatically calculate an ordering of recentness between imagery sources. This could be helpful for editors to sort/filter the imagery. 

- geojson is a good format to exchange geodata, but it is not optimized for size. We could look into different spatial formats that are more optimized regarding file size. E.g. [flatgeobuf](https://github.com/flatgeobuf/flatgeobuf) seems to allow spatial filters without requiring to download the whole dataset.

cicku:
My opinion is no more than 2 imageries should be allowed for the same region in ELI (composite is an exception).

It would be better for iD to support reading from ESRI metadata or Bing or...to tell users more. If they cannot, then it is their "fault" to trick users using obsolete things.

If people wish to get something from 10 years ago here, they should dump to OpenHistoricalMap.

simonpoole:
From a consumer of ELI POV such layers should be moved to the historicphoto category to make it clear that they have been superseded and shouldn't be shown except when the user is explicitly looking for them.

coolultra1:
I personally think, all imagery should be kept in ELI. This is an index of what exists, and not what is useful. As long as old imagery is classified historicphoto, there is no reason not to keep it.

rbuffat:
> there is no reason not to keep it.

Not true:

> The size of the imagery.geojson is unnecessary increased if we keep not relevant imagery
> The maintenance work is increased
> Mappers could not be aware that they are using old imagery

simonpoole:
> > The size of the imagery.geojson is unnecessary increased if we keep not relevant imagery

This is a general issue naturally. 

[...]

PR #1466, opened 2022-02-17
Add Worms imagery

coolultra1:
https://wiki.openstreetmap.org/wiki/DE:Permissions/Luftbild_Worms

Not a lot to say, plz merge


cicku:
Please remove imageries prior to 2012 unless you really have a good reason.

coolultra1:
There is no reason not to include the old imagery. Not only for OHM, it can
be used to see features long gone or obstructed by trees in the latest
aerial imagery.

cicku:
I don't think so. Long gone features should not be added

coolultra1:
I still think, there is no reason not to add it.

coolultra1:
> Please remove imageries prior to 2012 unless you really have a good reason.

Actually, you should be the one giving me the reason to remove. This project is meant to be a complete index, not just what might be needed.

rbuffat:
@coolultra1 The topic of how long imagery should be included in ELI is a bit controversial. See https://github.com/osmlab/editor-layer-index/issues/1446 for an ongoing discussion. 

As you probably know these imageries the best, what do you think how useful imagery older than 2012 will be for mappers mapping features that currently exist on the ground? I guess the newest imagery is the most useful, for older imagery there might be certain features that are better visible. Do you think there will be features that are not visible in either the imagery of 2020, 2016 or 2012 that still exist today?

cicku:
You are right, we are still discussing it. https://github.com/osmlab/editor-layer-index/issues/1446.

```
 create mode 100644 sources/europe/de/Worms-2003.geojson
 create mode 100644 sources/europe/de/Worms-2008.geojson
 create mode 100644 sources/europe/de/Worms-2012.geojson
 create mode 100644 sources/europe/de/Worms-2016.geojson
 create mode 100644 sources/europe/de/Worms-2020.geojson
```

Something from 10 years ago could be/still be useful, but 20 years? I don't think so. Because you have something from 2 years ago submitted, it is hard to make me believe that newer is not better. It is enough to compare between 2012, 2016 and 2020.

coolultra1:
So? @cicku @rbuffat? Is this mergable this way?

coolultra1:
@cicku, @rbuffat, this has unfortunately seemed to have caught up in discussion, but that shouldn't be a reason to delay merging this. People are still using old imagery around here, please merge this!

And if its really, really necessary, delete the imagery, my god, I don't care.

andrewharvey:
In my view we should accept this until there is a decision around #1446 and subsequent project policy.

sources/europe/de/Worms-2003.geojson
```diff
@@ -0,0 +1,29 @@
+{
+    "type": "Feature",
+    "properties": {
+        "name": "Worms 2003",
+        "type": "wms",
+        "url": "https://geoportal-worms.de/ogc/wms/luftbild-2003?LAYERS=D763F3C28EC14D8BAB4C307D33306FAF&STYLES=&FORMAT=image/png&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap",
+        "license_url": "https://wiki.openstreetmap.org/wiki/DE:Permissions/Luftbild_Worms",
```
andrewharvey, line 7: I'm not really in a position to assess the legalities here, but this sounds like a clear okay that they allow derived data like traced features to be owned by you, just we can't publish the actual imagery.

andrewharvey (approved):
This change looks good to me, I'm happy to accept it. @rbuffat did you have any opinion?

sources/europe/de/Worms-2003.geojson
coolultra1: @andrewharvey Yeah, I wrote the city myself and, as described on the Wiki page, they clearly stated, derived features may be used.

coolultra1:
> This change looks good to me, I'm happy to accept it. @rbuffat did you have any opinion?

Seems @rbuffat has got no opinion on this? Please, I'm begging anyone to merge this.

PR #1480, opened 2022-03-18
Fix URL for ACTmapi December 2021 Imagery

davidhicks:
data.actmapi.gov.au doesn't contain the latest December 2021 imagery but
data4.actmapi.gov.au does, so use data4.actmapi.gov.au instead.

### DIFF
--- a/sources/oceania/au/act/ACTmapi-Imagery202112.geojson
+++ b/sources/oceania/au/act/ACTmapi-Imagery202112.geojson
@@ -13,7 +13,7 @@
         "available_projections": [
             "EPSG:3857"
         ],
-        "url": "https://data.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery202112gda2020/ImageServer/exportImage?f=image&format=jpeg&imageSR=3857&bboxSR=3857&bbox={bbox}&size={width},{height}&foo={proj}",
+        "url": "https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery202112gda2020/ImageServer/exportImage?f=image&format=jpeg&imageSR=3857&bboxSR=3857&bbox={bbox}&size={width},{height}&foo={proj}",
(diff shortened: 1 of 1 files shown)


andrewharvey:
Ah interesting, it looks like a few of the older years will also need to use the other URL, did you want to update those too?

davidhicks:
I've just done a comparison between the CC-BY imagery listed at https://www.actmapi.act.gov.au/imagery.html and the ACTmapi imagery in editor-layer-index.

We're missing:
1. [March 2015](https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery2015mga2020/MapServer)
2. [March 2016](https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery2016mga2020/MapServer)
3. [Oct/Nov 2019](https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery201910mga2020/MapServer)

I'm happy to add these layers when I get a chance to trace the boundaries.

Additionally I recommend that ACTmapi-Imagery2019.geojson be renamed ACTmapi-Imagery2019**02**.geojson (and the ID from "ACT2019" to "ACT201902") but I wanted to check that renaming a source file is OK and won't break any guarantees to editor-layer-index users about imagery layer names being the same?

andrewharvey:
I meant that https://data.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY now only contains June 2020, Feb 2021, Apr 2021 however https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY contains these plus the rest. However in our index here, all the URLs are using data.actmapi so most will be broken currently so these will all need to be updated to data4.actmapi.

I think the 2019 file was added when there was only one, but it's okay to update the name and ID.

I'll merge this PR now, and then I'll fix the data/data4 URLs, but then feel free to add the other missing older sources.

andrewharvey (approved):

PR #1482, opened 2022-03-19
update au/act urls due to server changes

andrewharvey:


andrewharvey:
Further changes are needed here, the ArcGIS server using using MapServer for most of these however requesting tiles in EPSG:3857 is only working for low zooms at high zooms it returns a checkerboard image.

WMTS tiles work for JOSM but these aren't supported in iD due to the projection.

cc @davidhicks

davidhicks:
Since the ACTmapi imagery was [added in 2018](https://github.com/osmlab/editor-layer-index/issues/297), it looks like they've updated the imagery from using EPSG:4326 and EPSG:28355 projections to the new projection of EPSG:7855 (GDA2020). In the process perhaps they've been removing the old projection imagery from data.actmapi.gov.au and placing the new EPSG:7855 projected imagery on data4.actmapi.gov.au instead.

iD Editor doesn't support EPSG:7855 meaning that the imagery layers will be up to 3m misaligned at present. I also couldn't find a published simple approximation (via block shift vector) that could be implemented at https://github.com/openstreetmap/iD/blob/develop/modules/renderer/background_source.js#L139 and I assume a grid transformation approach would only be possible if editor-layer-index bundled some very approximate and simplified grids that are loaded by iD editor alongside each imagery layer needing an approximate/simple transformation of projections.

andrewharvey:
> Since the ACTmapi imagery was added in 2018, it looks like they've updated the imagery from using EPSG:4326 and EPSG:28355 projections to the new projection of EPSG:7855 (GDA2020). In the process perhaps they've been removing the old projection imagery from data.actmapi.gov.au and placing the new EPSG:7855 projected imagery on data4.actmapi.gov.au instead.

Ah that makes sense

> iD Editor doesn't support EPSG:7855 meaning that the imagery layers will be up to 3m misaligned at present.

It's more than that, iD can't even use the existing tiles at all, https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY/imagery2017mga2020/MapServer it can only make a WMS request for the tiles in EPSG:3857 but there is the checkerboard issue per my comment.

> I also couldn't find a published simple approximation (via block shift vector) that could be implemented at https://github.com/openstreetmap/iD/blob/develop/modules/renderer/background_source.js#L139 and I assume a grid transformation approach would only be possible if editor-layer-index bundled some very approximate and simplified grids that are loaded by iD editor alongside each imagery layer needing an approximate/simple transformation of projections.

Maybe if https://github.com/openstreetmap/iD/issues/4843 is done then it could all be handled on the iD side, until then it's not looking good for these layers in iD...

grischard:
Hi @tyrasd, what do you think ELI should do here?

simonpoole:
Isn't the simple solution for OSM Oceania to run a maproxy instance that reprojects the WMS layers? 

I would offer the SOSM instance, but it really doesn't make a lot of sense to send everything around half the globe and then back again.

andrewharvey:
> Isn't the simple solution for OSM Oceania to run a maproxy instance that reprojects the WMS layers?

Good suggestion

rbuffat:
@andrewharvey 
The services I checked under https://data4.actmapi.act.gov.au/arcgis/rest/services/ACT_IMAGERY seem now to support a "real" WMS server supporting EPSG:4326/CRS:84. E.g. https://data4.actmapi.act.gov.au/arcgis/services/ACT_IMAGERY/imagery2017mga2020/MapServer/WMSServer?request=GetCapabilities&service=WMS

FrakGart:
Hi,

Just to let you know, you can probably close this PR. The ACT Government is retiring their server (data4) and moving to ArcGIS hosted services. data4 is due to go off the air on 31 May 2024.

Thanks.

andrewharvey:
Thanks for the tip @FrakGart , looks like we also have 2023 imagery now https://actmapi-actgov.opendata.arcgis.com/search?tags=Imagery were you planning a PR to update this, or should I?

FrakGart:
I wasn't going to propose a PR (as I don't have the foo). See  #2340 for more details about what's available.

andrewharvey:
> Isn't the simple solution for OSM Oceania to run a maproxy instance that reprojects the WMS layers?

@simonpoole I attempted this but couldn't get it to work.

simonpoole:
The configuration tends to be a bit finaky (because yaml) but there shouldn't be a fundamental reason for it not to work. I might give it a go when I have a bit of time.

andrewharvey:
this is what I tried

```yaml
services:
  tms:
    use_grid_names: false
    origin: 'nw'

layers:
  - name: ACT_Aerial_Imagery_202309
    title: ACT_Aerial_Imagery_202309
    sources: [ACT_Aerial_Imagery_202309_cache]

caches:
  ACT_Aerial_Imagery_202309_cache:
    format: image/jpeg
    grids: [GLOBAL_WEBMERCATOR]
    sources: [ACT_Aerial_Imagery_202309_gda2020mga55_cache] 

  ACT_Aerial_Imagery_202309_gda2020mga55_cache:
    format: image/jpeg
    grids: [gda2020mga55]
    disable_storage: true
    sources: [ACT_Aerial_Imagery_202309]
                                                                                                                                                                                                                                               
sources:
  ACT_Aerial_Imagery_202309:
    type: tile
    url: https://tiles.arcgis.com/tiles/E5n4f1VY84i0xSjy/arcgis/rest/services/ACT_Aerial_Imagery_202309/MapServer/tile/%(z)s/%(y)s/%(x)s
    grid: gda2020mga55

grids:
  webmercator:
    base: GLOBAL_WEBMERCATOR
  gda2020mga55:
    srs: 'EPSG:7855'
    bbox: [658034.3343965416, 6065374.9924620865, 735626.203736689, 6114136.442014559]
    bbox_srs: 'EPSG:7855'
    origin: 'nw'
    min_res: 156543.03392800014
    max_res: 0.07464553541190416
    num_levels: 22


globals:
```

PR #1249, opened 2021-09-04
Fixed and Improved Brazil's IBGE sources

matheusgomesms:
Fixed issue #983 (updated all URLs when possible), added privacy policies, and did minor fixes on them, as discussed previously on PR #1233.

PR #1484, opened 2022-03-20
Addition of Dourados-Brazil Ortophotos

matheusgomesms:
2018-2020 high-res imagery from Dourados. We have [explicit permission](https://wiki.openstreetmap.org/wiki/File:Opendata-dourados-ms.jpg) to use their data.

I would also kindly ask to review all other Brazilian sources added that we have permission, PRs opened months ago (PR #1255  and #1249).

PS.: I can never make the geometry check work, although it seems this geometry is pretty fine.

Best,
Matheus


andrewharvey:
Your geometry looks fine (https://github.com/osmlab/editor-layer-index/blob/0ec4eaf05c6d9053c74a539e0e54a334a22810bf/sources/south-america/br/Dourados.geojson) it could be either the check is buggy or the server is serving data metadata. Happy to ignore it since it looks okay.

Did you check the 2020 imagery? I can only see the 2018 imagery loading.

matheusgomesms:
Thanks for taking a look!

Perhaps the metadata is missing, not sure. About the images, 2018 is from the main city and 2020 from the districts. I just checked on -22.1816,-54.6708 the 2020 district layer on iD and it seems it is working fine.

andrewharvey:
Thanks I can see it in iD but not QGIS, maybe the metadat is wrong on the server, and QGIS is choosing not to show it, but iD just using the data url will try and load and find the data. All good then.

andrewharvey (approved):
CI check seems like it may be due to buggy server metadata, which we can ignore as manually checked configuration here to be okay.

matheusgomesms:
Thanks, Andrew! Much appreciated!

Issue #1400, opened 2022-01-05
[Question] About creating an Issue or plural Issues for multiple datasets


nyampire:
I recently discovered that a number of GeoTIFFs are available under an open license as part of a project called Plateau, run by the Japanese government.
(This is a multiple license under the Japanese government's open data license agreement, CC-BY4.0, ODC-BY or ODbL)

OSM Foundation Japan branch is currently converting the data into tiles and making them available for use in OSM mapping and quality assurance work for Plateau data importing.
There are 24 cities in the GeoTIFF data set, but the cities are far from each other, so there will be multiple layers for each data set.
We are concerned that the number of issues will be too much because there are 24.

Is it acceptable to submit issues one by one for each layer?
Or is it better to create multiple configuration geojson-files and register them as a single issue?

cicku:
What kind of REST service they use to publish these tiles? WMS? WMTS? JSON?

nyampire:
It will be xyz TMS, and will be created for each city.

cicku:
If each place will be assigned an unique TMS URL then you have to submit 24 different files. You can submit one by one or all in one, it is up to you and there is no issue as long as they are free to go.

However, if 24 cities are published under one TMS URL then it is not needed, GeoJSON supports multiple places together in one file.

nyampire:
Thanks a lot, I understood! And I'm glad to know we can use multipolygon feature to indicate multiple places.

I'll discuss this to our team which approach could be better. (to use, to operate and/or to update in future and so on.)

nyampire:
(This issue could be closed)

nyampire:
Hello team, sorry to add re-open comment on this issue. @cicku 

I'm writing the geojson to register 24 cities of Japan.
(Plateau project starts TMS service as official finally!)

It could be `FeatureCollection` I guess, but `FeatureCollection` could not include `properties`.

Could you give me a hint to write properties for this layer-index as valid format?

My WIP file is here. 
https://github.com/nyampire/editor-layer-index/blob/Add_JP_plateau/sources/asia/jp/JapanMLIT-PlateauOfficial.geojson

cicku:
An example is here:

https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/north-america/us/ar/ADOP_2017_6in.geojson

nyampire:
Thank you so much! @cicku 
I have successfully created the geojson file. Separately, I would like to submit a PR.

PR #1499, opened 2022-04-11
Add Plateau official tile dataset, Japan

nyampire:
Hello team!

I got the permission (waiver) to use Plateau Tile distribution by MLIT, a Japanese ministry. So I would like to add the URL to the index.

With great help by @cicku  (Thank you!) in #1400 , hopefully I could made the geojson file with probably valid format.

(behind the scene) Plateau Project is a project led by MLIT. They are publishing opendata for 56 cities currently. It contains CityGML, GeoTIFF and so on. They starts official tile distribution for 24 cities.

The detailed information is available here. (sorry, available in Japanese.)
https://wiki.openstreetmap.org/wiki/JA:MLIT_PLATEAU
https://wiki.openstreetmap.org/wiki/JA:MLIT_PLATEAU/tiles_official


cicku:
Hello,

Following checks here, you need to replace {z} with {zoom} in all geojson files.

nyampire:
@cicku Thank you for checking. I've replaced the "z" to "zoom" at the file.

Please take a look.

cicku:
Hi,

You can check errors at the `Details`:

image


First, does it have a privacy policy? I think the data is hosted on AWS so maybe this one:

- https://aws.amazon.com/privacy/

nyampire:
Thank you!
I found it had 3 errors.

1. Privacy Policy
The Plateau project itself [does not have](https://www.mlit.go.jp/plateau/site-policy/) their own Privacy Policy.
But as you mentioned above, the data is hosted by AWS. I'll follow your opinion.
Probably this is fixed.

2. `ring orientation does not correspond to GeoJSON RFC7946`
I've converted the file via [Right Hand Rule GeoJSON Fixer](https://mapster.me/right-hand-rule-geojson-fixer/).
But it still has same errors...

3. `please consider using corrected geometry: `
Is this error related to 2?

cicku:
I think 2 and 3 are related. CC @rbuffat for a review.

rbuffat:
I'm sorry I had lately not much time available for this project, thus my late reply. 

Regarding 1: if there is no privacy policy available it is since a few weeks ago accepted to set the value to `false`.
Regarding 2 and 3: the geojson specification [1] specifies an ordering of rings for Polygons. This check is also quite new.  The check shout output the Polygon in the correct ordering according to the geojson standard. It should be sufficient to just replace the existing geometry in the geojson file with the output of the check script.  If not, then there is a bug in the script. 


Regarding 

[1] https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6

PR #1500, opened 2022-04-24
Update name of stuttgart luftbild

david-sawatzke:
It seems like the gis server is delivering updated imagery from 2021 (for example, you can see the recently completed transit station at https://www.openstreetmap.org/edit#map=19/48.78349/9.18779).

On https://gis6.stuttgart.de/maps/index.html the 2021 and previous "Luftbilder" can be seen.

Due to permission being granted for the future captures, this shouldn't be an issue, but I don't know the capture dates.


rbuffat:
@david-sawatzke Thanks for your PR!

What would you think about using 
```
        "start_date": "2021,
        "end_date": "2021",
```
instead of removing the entries?

Also "category": "photo" could be added. The current check script is more extensive compared to when the imagery was added the first time, thus now this "error".

david-sawatzke:
Makes sense, didn't think of that

rbuffat:
@david-sawatzke Thanks!

PR #1508, opened 2022-05-04
Remove osmdata.asitvd.ch hosted tiles as they will be retired on 30.06.2022

yblatti:
[ASIT ](https://asit-asso.ch/)(previously ASIT VD) will retire hosted tiles on 30.06.2022.

Reasons provided by ASIT are : 
- Service is not used by mappers anymore (based on access logs)
- Tiles are old (flight year 2012-2016)
- Switzerland is now fully covered by Open Data imagery ([SWISSIMAGE](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ch/swisstopo_swissimage.geojson)), fresh (2020+ for the same regions), with higher resolution (10cm)

This PR removes osmdata.asitvd.ch hosted tiles.
This also fixes some Solothurn imagery which mistakenly pointed to ASIT VD privacy_policy_url (bad copy pasta).


yblatti:
End of service message : http://osmdata.asitvd.ch/

grischard:
Thank you @yblatti !

Issue #691, opened 2019-08-10
Add the Esri imagery as "best" for Estonia (EE)


SviMik:
Currently, the "best" imagery for Estonia is:
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ee/EstoniaOrtho%28Maaamet%29.geojson

However, it is in EPSG:3301, which is not supported by the iD editor.
After some [discussion](https://github.com/openstreetmap/iD/issues/6736) with iD guys we came to a possible solution that more that one imagery can be marked as "best".

I did a quick check, and found that Esri imagery look exactly same as `EstoniaOrtho(Maaamet).geojson` but has a projection that works in iD. Could you add Esri to EE and mark it as "best" please?

grischard:
We can't currently mark one source as 'best' by region - it's all or nothing, everywhere or nowhere, best or not. How clients should handle more than one 'best' layer is undefined.

Am I understanding what you're suggesting incorrectly?

bhousel:
@grischard - see https://github.com/openstreetmap/iD/issues/6736#issuecomment-520155838
For some reason there are 2 "Estonia Ortho (Maaamet)" sources - A TMS one and a EPSG:3301 one.

They mostly just want the `best: true` moved to the TMS one so iD can use it.  We can probably remove the EPSG:3301 one entirely, if it's the same imagery.

SviMik:
The TMS one looks really bad in iD for some reason! I don't know if it's just me, but it looks like jpeg with 0% quality, barely can see anything. I would consider removing this completely if other people see the same thing.

> We can't currently mark one source as 'best' by region

It's really sad then. Any suggestion how to solve the problem? iD is showing Bing in Estonia by default, and it's a really awful source for Estonia comparing with Esri. If really nothing can be done, maybe just clone Esri, name it Esri (Estonia), and put it as best for this region?

grischard:
Well ideally I'd like to have something where editors request only local imagery, see #586. But that doesn't help you out a lot here.

I don't want to fragment the ELI global layers into potentially hundreds of local layers. The global geojson is already approaching 2 MB, making editing expensive and slow for editors on limited connections.

If Esri is the same as Maaamet but better, we could actually remove Maaamet completely.

SviMik:
I'm afraid removing Maaamet wouldn't solve the problem. Let's not remove anything if it doesn't solve anything.

And we still need ideas how to make Esri the default source for Estonia.

[...]

PR #1519, opened 2022-05-26
Update Maa-amet WMS URL to use EPSG:3857

Pikse:
The "best" source for Estonia previously used projection that isn't supported in iD editor (see #691).

Also doing the following (minor) adjustments to Maa-amet layers:
1.  make distinction between two orthophoto layers (WMS and TMS) more clear, and harmonize other three source names accordingly,
2. adjust two of these layer names (Basic Map, Cadastral Map) to Maa-amet's own translation,
3. enable i18n as these are translatable names,
4. correct capitalization in Estonian-language institution name (Maa-amet)
5. remove hillshading layer which has been unfunctional via given source for a while.
6. add mandatory properties to comply with new checks


grischard:
Thanks!

PR #1540, opened 2022-06-16
Add property 'support_cors' and set it to false for ESRI Clarity & Mapnik BlackWhite

Binnette:
Hi,

This is a simple PR that add a new property "support_cors". It indicates wether the imagery tile server support CORS or not.

I set this property to false for:
- ESRI Clarity
- Mapnik Black & White

In fact, those 2 tiles server to not send the header "Access-Control-Allow-Origin: *" and  because of that tools using those imagery will get a CORS error.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -252,7 +252,11 @@
                 "minimum-tile-expire": {
                     "description": "minimum expiry time for tiles in seconds. The larger the value, the longer entry in cache will be considered valid",
                     "type": "integer"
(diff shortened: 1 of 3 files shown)


cicku:
I'm not able to accept this PR because it introduces an undocumented property which is not widely used in this project.

Binnette:
Ok I will create an issue instead 😇👌

PR #1559, opened 2022-07-26
Create Berlinaerialphotograph2022.geojson

tordans:
Same as the previous years: The city recently published new aerial imagery which the community re-published as TMS source.

I just copied and updated one of the previous files inside Github. Is there an easy way to check if this will work in iD?


tordans:
@polarbearing FYI since you worked on those images before. A review of this PR would be great.

grischard:
Passes the checks and no big difference from 2021, looks good to me! You've added it to JOSM too?

tordans:
> You've added it to JOSM too?

no, but I asked the local community to please take care of it.

Issue #1530, opened 2022-06-07
Request: Turku aerial imagery 2021


larmax:
Hi, could the layer called "Turun Ilmakuva 2021" (Turku aerial imagery 2021) be added? It shouldn't replace the old one as that covers a larger area. Not sure how to get the bbox though.

_Originally posted by @larmax in https://github.com/osmlab/editor-layer-index/issues/751#issuecomment-1145998425_

**URL of the imagery**

https://opaskartta.turku.fi/TeklaOGCWeb/WMS.ashx

Description page: https://www.turku.fi/turku-tieto/kartat-ja-paikkatieto/karttapalveluiden-rajapinnat

**Two letter country code**

fi

**Desired layer name**
Turun Ilmakuva 2021

**License** (required)

CC-BY 4.0 with OSM waiver (http://www.turku.fi/avoindata/jhs189)

**Is the license compatible with OSM?**

Yes

**Attribution** (text, URL, required?)

Mention in "[Contributors](https://wiki.openstreetmap.org/wiki/Contributors)" wiki page enough

**Privacy policy**

https://www.turku.fi/turku-tieto/tietosuoja/kayttotiedot-ja-evasteet

PR #1567, opened 2022-08-16
Add Turku ortophoto 2021

crimsondusk:
City of Turku has a new, updated ortophoto from 2021 available. This is from the same wms service as my previous pull requests.


grischard:
Thank you! Do you want to rename the file with no year in it, to make life easier for 2023 crimsondusk?

larmax:
Should [this issue](https://github.com/osmlab/editor-layer-index/issues/1530) be closed as it's being handled here now?

cicku:
1. Logo should be: https://www.turku.fi/sites/default/files/styles/site_logo/public/sites/all/themes/custom/driveturku/images/sprites/logo.png
2. WMS is using PNG, can you change to JPEG if there isn't any huge difference? Because PNG consumes a lot more bandwidth.

crimsondusk:
Updated the icon.

The WMS appears to only serve PNG. GetCapabilities only lists image/png, and it does not work if I request JPG.

I decided to not rename the ```turku-orto``` one because it does not have the year in the WMS layer, but I thought that having the year in the title is still a good idea. In case the WMS layer is ever updated, we need to update the title.

PR #1575, opened 2022-09-01
Add CT shaded relief

Mashin6:
Adding shaded relief tiles from CT ECO.
image


cicku:
Please include year in the name.

sources/north-america/us/ct/CT_ECO_Shaded_relief.geojson
```diff
@@ -0,0 +1,468 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "CT_ECO_Shaded_relief",
```
cicku: Year

Mashin6: Does the file name have to be changed as well?

cicku: I'd recommend doing that, both id and filename

cicku:
Merged, thanks!

PR #1577, opened 2022-09-03
new 2021 imagery for Cook County, Illinois

stevevance:
Cook County published the 2021 imagery in August 2022


cicku:
Interesting, I checked it last week and did not find any updates.

Thanks!

PR #1593, opened 2022-09-20
Slovenia: GURS DOF025 orthophoto updated up to 2021

stefanb:
In August 2022 orthophoto in central part of Slovenia was updated by @uprel with 2021 recordings, see https://site.geo-portal.si/2022-08-17-vkljucitev-dof-21.html


stefanb:
Updated also in JOSM: https://josm.openstreetmap.de/wiki/Maps/Slovenia?action=diff&version=70

PR #1603, opened 2022-10-01
Update Israel sources

zstadler:
Based on URLs published at
https://github.com/IsraelHikingMap/Map#tile-urls

### DIFF
--- a/sources/asia/il/IsraelHiking.geojson
+++ b/sources/asia/il/IsraelHiking.geojson
@@ -4,7 +4,7 @@
         "id": "Israel_Hiking",
         "name": "Israel Hiking",
         "type": "tms",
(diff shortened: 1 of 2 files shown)


cicku:
Please consider removing `/Hebrew` from the URL, because `Israel_Hiking` is not broken.

For MTB, isn't this one working?

`https://israelhiking.osm.org.il/mtbTiles/{zoom}/{x}/{y}.png`

zstadler:
We, the [Israel Hiking Map team](https://github.com/orgs/IsraelHikingMap/people) - the maintainers of these sources, prefer that the official URLs will be used.

For backward compatibility, the equivalent legacy URLs, without the `Hebrew` part, are supported for the time being..

PR #1608, opened 2022-10-08
Helsinki Region Ortophoto update

houtari:
Helsinki Region Ortophoto update - New permission asked for the 2021 photo. Will create a github issue for this task.


houtari:
Here's [the issue](https://github.com/osmlab/editor-layer-index/issues/1609) I created for this topic.

cicku:
Hi,

Can you put the screenshot on OpenStreetmap wiki and then create a page so you can use that page for `license_url` then?

Asteliks:
@cicku that will be needed at the very least

cicku:
Sorry I don’t get what you said here.

If an imagery can be used for openstreetmap, the license/authorization can be put on openstreetmap website to let more people know.

Of course, you can put it here, but I’d like to hear the opinion from @houtari but not anyone else.

Asteliks:
@houtari & @cicku I have analyzed the hsy.fi website and have found the link to the terms of use. There it clearly states that their services are under the CC 4.0 BY licence thus there is no need for this email confirmation. It all can be found here: https://www.hsy.fi/en/air-quality-and-climate/geographic-information/open-map-service/ 
image

cicku:
Thank you.

I think this PR can be closed as it is not needed.

houtari:
Wait a minute! How come you all of sudden don't need an authorization to include CCBY40 licensed data into OSM? What about [ODbL Compatilibility](https://wiki.openstreetmap.org/wiki/Import/ODbL_Compatibility)? See row "CC-BY 4.0 International" where it clearly says _**likely incompatible attribution requirements and other terms; additional waivers required for reasonable attribution and unrestricted distribution**_. See also [this blog post](https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/) about using CCBY40 licensed data.

Until now this kind of waiver/authorization has every single time been required when using CCBY40-lisenced ortophotos with OSM. See eg. [these discussions](https://github.com/osmlab/editor-layer-index/issues/339), so why not in this case? Has something changed lately with the licenses and their interpretation?

Asteliks:
It would seem that I have misread the intention of this PR. I thought that you were confirming the license. My mistake. We will need the things confirmed:

- [ ] Is the person answering your email empowered legally to make this decision?
- [ ] Could you use the [official waiver](https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Cover_letter_and_waiver_template_for_CC_BY_4.0). We need the concessions to be fully stated and implicitly confirmed. We should not accept authorization such as "same way as 2017" without it referring to a legal document. Unless you can provide a higher-level legal document stating otherwise.
- [ ] Since we are online and, in the EU, please make sure to have the document properly signed with [XAdES](https://en.wikipedia.org/wiki/XAdES) or [PAdES](https://en.wikipedia.org/wiki/PAdES). This is referring to the [eIDAS](https://en.wikipedia.org/wiki/EIDASl).

houtari:
Hello!

I've explained the situation in more detail in this [issue](https://github.com/osmlab/editor-layer-index/issues/1609).

The person answering my email is indeed empowered legally to make this decision. I mailed my question to avoindata@hsy.fi and HSY (Helsinki Region Environmental Services) is as I've explained behind the issue link above the owner and procurer of the orthophotos and avoindata (opendata in english) is a common email they have regarding data issues and questions.

An official waiver? Is it really necessary? I must say I'm a bit confused by these different requirements that seem to change from time to time and depending on who is handling these issues on github. No offence intended, but I would really appreciate some consistency in these requirements as I really made an effort out of preparing this authorization process and ask for permissions according to the requirements issued last time these Helsinki Region images were updated in this github repo. Discussions with officials can be time consuming and even frustrating at times, so I try to keep it at mimimum. :-)

In this last process it was perfectly fine to just ask for permission by email and document this email discussion - that was it. No other requirements. See [Add Orthophotograph of Helsinki #339](https://github.com/osmlab/editor-layer-index/issues/339#issuecomment-410677315)  for further details.

Has something dramatically changed since this last time so that a properly (XAdES, PAdES & eIDAS) signed official waiver really is needed?

Asteliks:
Thought it may not seem so I understand your pain. The truth is that not all of us have the necessary law knowledge but in case something goes horribly wrong we will have to answer for our deeds. When it comes to me I know the Polish law and know what is allowed here. Unfortunately I do not know your local laws thus I’m basing my requirements on the EU law. This is why I ask for a bounding document that can clearly cover all the bases those being:
- having a document saying **implicitly** that we may use the source (the official waiver is great for that);
- making sure that an authorized person has signed it;
- making sure that the signature can be confirmed (PAdES covers me here).

Otherwise just as well somebody could create a fake email and ask me to add the source. Should something go wrong I could get into real trouble. Probably the odds of that are quite low but do I really want to risk that?

houtari:
Thanks for answering. I do appreciate it even though I didn't quite like the outcome of it. I do also understand your point.

Working for the public transport authority in the region I do know that creating these kind of documents can take time. Let's hope it will not be a show stopper in this case. I do know that for some local municipalities and their orthophotos that I have had my hopes on it surely will be.

Being a part of this regional orthophoto procurement process (we have already the next photos planned for 2023) I'll se what I can do to promote an official waiver being used in the future.

About this particular issue I'll probably take a break from it for a while and go back to basics ie. OSM editing. Let's see if someone else wants to continue with this case. This is supposed be a hobby and now suddenly too much resembles what I do for work. :-)

Asteliks:
I know that I shouldn’t say that but technically there are two alternative ways. You could ask someone else to check your PR or you could ask to become a contributor and absorb the risk

simonpoole:
@houtari I think everybody realizes that the situation is a pain, but the underlying reason is an unsuitable licence being chosen that in the end likely doesn't even do what the licensor intended. And yes, the conclusion is that they should change their licence and all the friction will go up in smoke.

simonpoole:
> It would seem that I have misread the intention of this PR. I thought that you were confirming the license. My mistake. We will need the things confirmed:

Note that the CC BY 4.0 waiver is only required in scenarios where 3rd party data (or derivatives of it) is being imported in to OSM or the licensor is claiming that they have rights in traced data from imagery (and that those are licensed on CC BY 4.0 terms*). Normally we would be looking to receiving one of the declarations here https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers for imagery.

&ast; I personally would suggest that we shouldn't be using anything from an licensor that claims that to start with, but that is OT here

houtari:
Thanks @Asteliks for your comments and honest approach. Let's see what happens. I do clearly understand the benefit of local knowledge in this case. Thanks also to @simonpoole for insights into technical licensing details and especially for the aerial imagery waiver template. That was a new document for me. About changing the license - easier said than done. At the moment I don't have the motivation to start that kind of a "fight". :-)

Issue #1429, opened 2022-01-30
[Watchdog] Imagery "Ruda Śląska: Orthophotomap (aerial image)": sources/europe/pl/RudalskaOrthophotomap(aerialimage).geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"Ruda Śląska: Orthophotomap (aerial image)" -> [sources/europe/pl/RudalskaOrthophotomap(aerialimage).geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/europe/pl/RudalskaOrthophotomap(aerialimage).geojson)

Reason:
Error: WMS -: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?SERVICE=WMS&REQUEST=GetCapabilities Error: Could not parse XML.
Error: WMS 1.3.0: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.3.0 Error: Could not parse XML.
Error: WMS 1.1.1: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.1 Error: Could not parse XML.
Error: WMS 1.1.0: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.0 Error: Could not parse XML.
Error: WMS 1.0.0: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.0.0 Error: Could not parse XML.
Error: WMS -: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?service=WMS&request=GetCapabilities Error: Could not parse XML.
Error: WMS 1.3.0: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?service=WMS&request=GetCapabilities&version=1.3.0 Error: Could not parse XML.
Error: WMS 1.1.1: URL: https://rudaslaska.geoportal2.pl/map/wmsorto/wms.php?service=WMS&request=GetCapabilities&version=1.1.1 Error: Could not parse XML.
[...]

PR #1610, opened 2022-10-08
Delete RudalskaOrthophotomap(aerialimage).geojson

Vieja:
Linked to the issue https://github.com/osmlab/editor-layer-index/issues/1429

Checked the source, the orthophotomap layer is unavailable and will no longer be. File removed.
@Asteliks please verify.


Asteliks (approved):
I can confirm that this service is no longer available

PR #1611, opened 2022-10-08
Added licence and privacy policy urls for Poznan orthophotomaps

Vieja:
URLs were updated fot PoznarthophotomapBiezaca and added for older layers 2014 and 2016.


Asteliks:
@rbuffat could you take a look at this error. Unfortunately, I'm having trouble understanding it

rbuffat:
@Asteliks there is a bug in the WMS GetCapabilities parser. I will have a look at it tomorrow.

Issue #1205, opened 2021-07-21
[Watchdog] Imagery "Bytom: Fotoplan 2014 (aerial image)": sources/europe/pl/BytomFotoplan2014(aerialimage).geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"Bytom: Fotoplan 2014 (aerial image)" -> [sources/europe/pl/BytomFotoplan2014(aerialimage).geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/europe/pl/BytomFotoplan2014(aerialimage).geojson)

Reason:
Error: WMS None: Timeout for: http://sitplan.um.bytom.pl:11180/isdp/gs/wms?service=WMS&request=GetCapabilities
Error: WMS 1.3.0: Timeout for: http://sitplan.um.bytom.pl:11180/isdp/gs/wms?service=WMS&request=GetCapabilities&version=1.3.0
Error: WMS 1.1.1: Timeout for: http://sitplan.um.bytom.pl:11180/isdp/gs/wms?service=WMS&request=GetCapabilities&version=1.1.1
Error: WMS 1.1.0: Timeout for: http://sitplan.um.bytom.pl:11180/isdp/gs/wms?service=WMS&request=GetCapabilities&version=1.1.0
Error: WMS 1.0.0: Timeout for: http://sitplan.um.bytom.pl:11180/isdp/gs/wms?service=WMS&request=GetCapabilities&version=1.0.0

CC contributors to this imagery:
@rbuffat, @grischard

rbuffat:
Imagery seems to work again.

PR #1613, opened 2022-10-08
Bytom WMS update

Vieja:
Linked to https://github.com/osmlab/editor-layer-index/issues/1205

Updated existing layers with new url, license and policy. Added new orthophotomaps of the region.


Asteliks:
@rbuffat we may have a similar bug here

PR #1641, opened 2022-10-15
Fixed Brasília-Brazil sources

matheusgomesms:
Hi,

Brasília (Federal District) updated their URLs, so they were broken on ELI. I've updated them, and I also added an icon and newer imagery (2021) from same source.

Best,
Matheus


matheusgomesms:
Wille fixed these layers in #1786, so I'm adding the 2021 imagery here with some small changes as well.

PR #1534, opened 2022-06-08
Update OS OpenMap Local to April 2022

Cj-Malone:
Untested

PR #1663, opened 2022-11-06
Update OS OpenMap Local to October 2022

UKChris-osm:
Update OS OpenMap Local to October 2022

Used #1534 as a reference.

Re: [b6a8604](https://github.com/openstreetmap/chef/commit/b6a86046121d80e5c5076ef8482e6b13c35ecae2)

### DIFF
--- a/sources/europe/gb/OSOpenMapLocal.geojson
+++ b/sources/europe/gb/OSOpenMapLocal.geojson
@@ -1,12 +1,12 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "OS-OpenMap_Local_2022_04",
-        "name": "OS OpenMap Local (April 2022)",
-        "url": "https://{switch:a,b,c}.os.openstreetmap.org/layer/gb_os_om_local_2022_04/{zoom}/{x}/{y}.png",
+        "id": "OS-OpenMap_Local_2022_10",
+        "name": "OS OpenMap Local (October 2022)",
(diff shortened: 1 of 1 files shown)


FloEdelmann:
Is the older version of the imagery still available? If so, why not keep them all in the index instead of only the newest one?

UKChris-osm:
The older imagery is available at [https://os.openstreetmap.org](https://os.openstreetmap.org/) although personally I think there aren't a huge amount of changes to make having older versions available within iD, and all it would do is make the selection list unnecessarily big.

boothym:
@FloEdelmann the main reason for this layer is to have the latest available OS open data, which contains any new roads and street names etc. This makes it very useful when trying to map new roads or updating construction sites which aren't available on any other imagery.

So I'm not sure why anyone would need to have a layer with data from 6 months ago, I can't see a use case for that when mapping? There is an older style version of the layer from 2016 available on ELI if required though.

FloEdelmann:
Some features may be seen better in one image or the other (e.g. because of tree leaves blocking the view in summer, but not in winter). Having another relatively recent option available would be helpful in that case.

Another use case would be comparing the imagery in one place to see *when* a feature changed on the ground.

boothym:
That's correct and I agree if this was aerial imagery, but it isn't 😉 And it's published every 6 months going back to 2010 so that would be a lot of layers in the list!

FloEdelmann:
Ah, then I agree it doesn't make too much sense. The only use case would be comparing a feature at different points in time, and that is not the normal use case for OSM editors.

PR #1695, opened 2022-12-06
Add AGIS 2022 imagery

simonpoole:
This adds 20cm nominal resolution imagery from mid 2022 for the Canton Aargau.

Thanks to @rbuffat for pointing the update out.


rbuffat:
@simonpoole Thanks!

PR #1723, opened 2022-12-10
imagico.de osmim 2022 update

imagico:
Changes:

* adding new Antarctic layers
* changing some URLs to https
* changing some older images from `photo` to `historicphoto`

As mentioned before - the metadata, including additional info like sun direction, is also available in tilejson format under for example https://imagico.de/map/osmim_tiles.php?layer=S2B_R121_S7X_20220301T085959 - list of layers under https://imagico.de/map/osmim_tiles.php?layer=


cicku:
Hi @imagico, can you amend the commit and trigger a new check here? I'm afraid it is impossible for us to merge anything while the check has failed.

imagico:
Frankly i have no idea what the issue is here.  The changes submitted are script generated using the same process i have used before.  I see nothing in https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md, https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md or https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json that would explain what has changed in the requirements.  GDAL/OGR is able to read the files without problems.

If there are more strict requirements for the layer specifications here it might be worth to consider generating files complying with the strict specifications from a more tolerant submission format to make it easier to contribute.

Alternatively you could of course also pull in the layer information from metadata supplied by the layer provider - like i linked to.

grischard:
Closing and reopening to trigger CI checks

grischard:
@imagico the CI checks have fixes for the polygons (clockwise vs counterclockwise), and warn of missing tiles at certain zoom levels: https://github.com/osmlab/editor-layer-index/actions/runs/5257332513/jobs/9499952749?pr=1723

imagico:
Thanks, that makes it a bit clearer.

The winding order requirement for polygons is a new one since the previous update in 2020 and it seems undocumented.  I can change that in the scripts but considering by default OGR does not meet this requirement when generating GeoJSON output i would politely suggest not to make this a requirement on contributors and this way create an additional hurdle for contributions.  If you want to still make this a requirement i suggest to document it clearly.

The missing tiles seems likewise based on undocumented requirements.  The empty tile rules of my layers are explained in https://github.com/osmlab/editor-layer-index/pull/1042#issuecomment-772124528.  These are originally developed w.r.t. JOSM being able to gracefully deal with empty tiles and avoiding unnecessary traffic.  If this project has different requirements please say so.

Note many of the high latitude layers have a variable maximum zoom level - taking into account the highly variable scale of the Mercator projection at high latitudes.  Hence it is natural that tile requests within the bounds of the layer return a 404 at or below max zoom level.  This just means that at the latitude requested this zoom level would be beyond the available resolution and the editor should use and overzoom a lower zoom level tile instead.

imagico:
Superseded by #2005

PR #1727, opened 2022-12-19
Add Lublin BDOT500 layer

maciej-zuk:


rbuffat:
@maciej-zuk Thanks for your PR. Sorry for the long response time! https://otwartedane.lublin.eu/dataset/dane-geodezyjne-w-ramach-iip indicates that the license is cc-by. This means that attribution is required. In this case, we should check that it is OK for the data owners to attribute the usage of the data with an entry in https://wiki.openstreetmap.org/wiki/Contributors You can find a waiver template here: https://wiki.osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates

See also 

>  [Creative Commons with condition (CC BY, CC BY-NC, CC BY-SA)](https://creativecommons.org/share-your-work/licensing-types-examples/) imagery sources are most likely compatible with tracing on OpenStreetMap, but to ensure no objection, we ask you confirm with the owning organization by using the [creative commons waiver](https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#where-can-i-find-an-imagery-license-waiver).

From https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#what-imagery-licenses-are-compatible-with-this-index

CC @Asteliks

maciej-zuk:
i have sent them an email but no response yet

PR #1749, opened 2023-01-21
Latvia: add latest orthophoto layers

soshial:


cicku:
The suggestion will be if you submit a layer that is considered as “latest”/“most recent”, then all others of the same area should have `category": "historicphoto"`

soshial:
Fixed it.

BTW, do we have "cadastre" category?

cicku:
It depends on our downstream, like iD or other editors.

soshial:
Could you please review my changes, @cicku @rbuffat?

rbuffat:
@soshial Would it be possible for you to link to the license information of this data?

https://www.lgia.gov.lv/en/node/857 seems not to state any license information. Or I overlooked it.

Davis-Klavins:
@rbuffat LVM GEO "license"should be applied. They don't have a specific one, just a statement that "published as open data available for any user". See https://www.lvmgeo.lv/en/data. If you have any doubts, contact them via lvmgeo@lvm.lv.

PR #1751, opened 2023-01-23
Spain: Add IGN Raster layer

Crashillo:
Include the official spanish IGN (national geographic institute) raster map tiles available for the spain region


cicku:
Please explain what you are doing here.

Crashillo:
Updated title and description

Crashillo:
Is it possible you guys are reviewing the PRs in a LIFO stack-way? I'm noticing that the latest open pull-requests are being merged _before_ than the older ones. Any reason to act like that?

rbuffat:
@Crashillo Thanks for your PR! Would it be possible for you to link to the metadata of this data / where the license of this dataset can be found?

Crashillo:
What do you mean, @rbuffat? The GeoJSON is including the required fields, _license_url_ and _privacy_policy_url_.
Besides, the [official spanish orthophoto imagery](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/es/PNOASpain.geojson) (already published) has exactly the same fields, as the organization is the same also (IGN), why should be this one different?

Crashillo:
Could you please give this PR a boost, @cicku? I didn't get any feedback yet, and the label is misleading

PR #1752, opened 2023-01-23
Spain (Basque country): Add geoEuskadi orthophotos

Crashillo:
Include the official geoEuskadi aerial imagery available for the basque country (spain) region


cicku:
Please explain what you are doing here.

Crashillo:
Updated title and description

Crashillo:
I'm noticing the checks are failing due to the geometry, but I got it from download.geofabrik.de .poly file. What's wrong with that? :thinking:

cicku:
Does it have a year mention or always latest? Also, there are 3 coordinates with different decimal place at the bottom.

Crashillo:
> Does it have a year mention or always latest?

Nope, all the layers I'm providing displays always the most recent photos.

> there are 3 coordinates with different decimal place at the bottom

Does it matter? I edited the polygon bounds to fit inside the bbox, so I guess geojson.io includes better precision.

cicku:
It does not matter much because the bot may fix it later.

PR #1755, opened 2023-01-25
rotate access token for mapbox satellite layer

tyrasd:
as requested by mapbox

### DIFF
--- a/sources/world/MapBoxSatellite.geojson
+++ b/sources/world/MapBoxSatellite.geojson
@@ -18,7 +18,7 @@
         "name": "Mapbox Satellite",
         "type": "tms",
         "category": "photo",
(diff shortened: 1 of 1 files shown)


simonpoole:
Not for this time around, but perhaps going forward we should probably replace this with the {apikey} placeholder and supply the apikey to app devs directly (maybe via the EWG?).

tyrasd:
> Not for this time around, but perhaps going forward we should probably replace this with the {apikey} placeholder and supply the apikey to app devs directly (maybe via the EWG?).

yes. :+1: 

//edit: and what we probably also should do is to issue dedicated api tokens for different consumers, i.e. a dedicated one for each editor / tool (iD, JOSM, maproulette,  mapcomplete, osmineditor, …). That would allow us to reduce disturbance when a token has to be revoked because of abuse.

PR #1757, opened 2023-01-27
Add JP GSI airphoto imagery updates

nyampire:
I found Japanese government has published their imagery updates with separate tiles.

They are slowly merged into the main imagery, but others are not.

So it would be better to have those with another entry.

I've checked those files with GeoJSON lint. Hopefully, they don't have any right-hand errors...


cicku:
Did you generate the file via JOSM?

nyampire:
No, those are distributed by GSI (JP government) in KML format. And I've converted those with QGIS, geojson.io and lint.

https://psgsv2.gsi.go.jp/koukyou/kihonsatsuei/index_photo_area2.html

Their data is distributed under their OpenData license (it covers the whole site), so it is not against the copyright. However, I might need their attribution somewhere.

cicku:
You may need to modify the file based on the check result and I will merge it then.

PR #1758, opened 2023-01-28
Spain (Catalonia): Add ICGC orthophotos

Crashillo:
Include the official ICGC aerial imagery available for catalonia (spain) region


cicku:
Questions:

1. ICGC is not a valid name.
2. Is it an imagery being frequently updated or not? If yes, you can keep the `best` flag. If not, you will need to specify the year (possibly `category`) and remove the `best`.
3. Generally speaking, PNG will consume more bandwidth, have you tried JPEG and is it acceptable to change to it?
4. Is it orthorectified? If yes, you need to specify via `valid-georeference`.
5. Does the provider have any icon/logo to be used?
6. Max/min zoom levels?

Crashillo:
> ICGC is not a valid name.

Updated.
> Is it an imagery being frequently updated or not? If yes, you can keep the best flag. If not, you will need to specify the year (possibly category) and remove the best.

Yep, it's the most recent one.
> Generally speaking, PNG will consume more bandwidth, have you tried JPEG and is it acceptable to change to it?

It was changed.
> Is it orthorectified? If yes, you need to specify via valid-georeference.

It's not.
> Does the provider have any icon/logo to be used?

No.
> Max/min zoom levels?

No restriction.

PR #1761, opened 2023-02-02
Create ideib_balears

marcboschmatas:
added geojson of Balearic islands with service


rbuffat:
@marcboschmatas Sorry for taking so long to answer. Would it be possible for you to add the `.geojson` extension to the file? Then the test script should also pick up your changes.

PR #1285, opened 2021-10-08
add Nerima Ortho imagery, Tokyo Japan

nyampire:
I would like to add an Orthoimagery resources to the repository.

The information wiki page is [on OSM wiki](https://wiki.openstreetmap.org/wiki/Nerima_ortho).

If this could be valid to merge, I would like to make more resources in Japan region.

This is the first time to add, so please forgive me if it has some errors. I've tested at [Imagery index](https://ideditor.github.io/imagery-index/index.html) and got valid result.

PR #1765, opened 2023-02-08
JP imageries by OSMF Japan

nyampire:
OSMF Japan is publishing imageries from OpenData and "Kokyo-Sokuryo" dataset.

OpenData imageries are available with the waiver process.

And "Kokyo-Sokuryo" dataset is defined in Japanese Survey law (測量法). They are created by municipalities, and available by application to them.

OSMFJ applied them and got the certification to publish, and use (mainly for tracing and machine learning sources) in OSM. It could be even under CC-BY license.

See detail in: https://wiki.openstreetmap.org/wiki/OSMFJ/ortho-tiles
Sorry this is available in Japanese, if any request you have, I'll happily to translate it


nyampire:
Hmmm, CI says it does not have `Privacy policy`. It is correct. But this is just a tile file storage by OSMFJ. (same approach as https://github.com/osmlab/editor-layer-index/pull/1285#issuecomment-945000448)

PR #1771, opened 2023-02-12
Replace Turku 2018 True ortho with Turku 2022 True ortho

larmax:
The City of Turku has released new true ortho imagery for which the city has granted permission for use in OSM.


rbuffat:
@larmax Thanks for your PR! Would it be possible to keep the 2018 imagery? Or would you be strongly against keeping it in?

cicku:
Please keep the 2018 imagery as @rbuffat has mentioned, and review the issues that I have pointed out.

arch0345:
The longitude/latitude for the 2022 geometry seems to be flipped.

PR #1786, opened 2023-03-08
Update URLs of GeoPortal DF Imagery

willemarcel:
URLs got outdated and were not working.


willemarcel:
@rbuffat it seems there is firewall blocking the access to the attribution url (https://www.geoportal.seduh.df.gov.br/) from outside Brazil. Is it ok merging it with a failing test?

rbuffat:
@willemarcel  Sorry for not replying earlier.  I would say yes. Ideally, if somebody is located within the country with access this person could run the test script manually on their computer.

PR #1787, opened 2023-03-10
Add cadastre layers for NRW, Germany

jomo:
This adds:

- ALKIS layer for North Rhine-Westphalia (NRW)
- ALKIS buildings overlay for NRW (same data source, but only renders yellow building outlines)
- ALKIS layer for Städteregion Aachen, NRW, Germany

The license for all these is [Data licence Germany - Zero - Version 2.0](https://www.govdata.de/dl-de/zero-2-0), which is essentially a translation of CC0. This repository already includes data using this license. Might be worth to list this in the [compatible licenses](https://github.com/osmlab/editor-layer-index/blob/gh-pages/FAQ.md#what-imagery-licenses-are-compatible-with-this-index)?

The data includes PNG renderings of government cadastre data (ALKIS) of buildings, roads, etc. It is usually of very high quality, as the data ultimately comes from official government data and land surveying and has an accuracy of at most a few centimeters.

The data for Aachen is mostly the same as for NRW (and eventually contributes to the Aachen part of the NRW data) but the Aachen data is updated daily (very useful to add new buildings) whereas the NRW data is only updated twice a year, and won't be updated in 2023 due to a data migration.


rbuffat:
Thanks @jomo !

crackwitz:
If anyone happens to notice the tiles being slow to download... that's because the server redraws them for every request, which is especially costly for the ALKIS/cadastre layers. There is no caching. The server even sets `Expires: Mon, 1 Jan 1900 00:00:00 GMT`. Caching could be trivial with a little bit of `nginx`.

PR #1789, opened 2023-03-11
Add TIGER 2022 roads overlay

### LINKED ISSUE #1777: TIGER Roads 2022
The U.S. Census Bureau released a new roads shapefile in October 2022. It can be downloaded [over the Web](https://www.census.gov/cgi-bin/geo/shapefiles/index.php?year=2022&layergroup=Roads) or [faux FTP](https://www2.census.gov/geo/tiger/TIGER2022/ROADS/). It would be great to have an updated version of the TIGER Roads 2021 overlay.

https://github.com/osmlab/editor-layer-index/blob/041610b632428c9d29504eb384edb2f5cef35ae6/sources/north-america/us/TIGER2021RoadsOverlay.geojson?short_path=4805311#L551

/ref #1467
/cc @iandees 🙏

iandees:
Fixes #1777


cicku:
Need a removal of old layers, we do not need to carry so many there.

iandees:
> Need a removal of old layers, we do not need to carry so many

They're useful for comparison.

1ec5 (approved):
I agree that it can be useful to compare the TIGER overlays from different years, since the dataset incorporates corrections and additions every year. These are pretty much the only overlays for the U.S., so I think the list in iD is still manageable. But eventually editors like iD could implement more scalable UIs or filter out older overlays if desired.

cicku:
Things of 2017 isn't useful. You can compare with 2020 and 2021, and that's fine.

I will remove 2017-2019 later this year, we do not need to carry 6 layers for the same dataset here.

iandees:
> do not need to carry 6 layers for the same dataset here

It is explicitly not the same data set. The point is that it changes every year.

cicku:
Not "exact", but "mostly".

> The point is that it changes every year.

That's my point as well, you need fresh data, and ditch the garbage of 6 years ago.

1ec5:
Looking at what’s already in OSM, you don’t know if it matched TIGER from back then or not unless you can easily compare it to the older overlay. Today’s TIGER can be garbage too, just of a different variety. There can even be regressions. Perhaps this kind of comparison would be easier if every street had a `source:name=TIGER 2017` on it, but that’s pretty rare.

If we purge older TIGER layers, I’d suggest a nonsequential pattern, for example every other year plus the last two years (2017, 2019, 2021, 2022). I assume a given road’s name would seldom flip-flop multiple times.

cicku:
We can discuss later when I open the PR, no worry.

Issue #1790, opened 2023-03-13
Layer "Kartverket N50 topo" contains unlicensed data


noenandre:
The layer titled "Kartverket N50 topo" contains data we don't have permission to use in Openstreetmap. The layer show «N50» data on lower zoom levels, which we do have a permission to use. But on higher zoom levels, past zoom level 15, the layer show more detailed map data from the Norwegian Mapping Authority, which we don't have a permission to use.
Users of may not be awhere, and may draw or trace of a unlicensed source.
On solution is to limit this layer to zoom level to zoom level 15. Another solution is switch out source of this layer to a WMTS layer called "kartdata3" at the same endpoint as before.

## URL of the imagery
https://opencache{switch:,2,3}.statkart.no/gatekeeper/gk/gk.open_gmaps?layers=kartdata3&zoom={zoom}&x={x}&y={y}

## Two letter country code
no

## Desired layer name
Kartverket N50 topo

NKAmapper:
I disagree with this statement. The map in question is licensed. Some of the vector data at the detailed zoom levels are not available for us for downloading, but that is a different thing. I think you should discuss this with the local community first rather than here.

PR #1791, opened 2023-03-13
Replaced map layer from Norwegian Mapping Authority

noenandre:
Replace layer with id "kartverket-topo4", with a similar new layer which don't contain unlicensed data.
See #1790


NKAmapper:
As commented in #1790, I disagree with this PR.

rbuffat:
@noenandre @NKAmapper It would be good if you/the local mapping community could find common ground and then either close this PR or propose how to go forward!

PR #1798, opened 2023-03-18
add ORKa.MV source layer

Strubbl:
This commit adds the ORKa.MV as a source layer. It is an OSM-based layer combined with cadastre layers.
The website can be found at https://www.orka-mv.de/

Further details can also be found in the OSM wiki at https://wiki.openstreetmap.org/wiki/Mecklenburg-Vorpommern#ORKA


grischard:
Please fix this issue:

ERROR:root:sources/europe/de/Mecklenburg-Vorpommern-ORKa.MV.geojson has no license_url set

Strubbl:
Thanks @grischard, i fixed this issue. I also fixed it in the documentention with #1803

PR #1801, opened 2023-03-22
Adds Rio's Mosaico 2022 with the same details as the previous two

nighto:
Adds new imagery source for Rio de Janeiro city, with the same area and license as the previous two existent sources.


matheusgomesms:
Lovely! Arlindo, what about also removing best=true from 2019 imagery and maybe delete 2015 imagery (it seems it's not working and is outdated, compared to the other images)?

nighto:
@matheusgomesms I thought about changing the `best` parameter, but turns out that the 2019 imagery is much better than the 2022 (this one has some clouds etc). It is of course much useful to see recent changes. Here's an example of some clouds on the imagery.
image
I will double check with my contact on the city hall if the 2015 imagery is gone for good.

nighto:
btw thanks for pointing out about the `best` parameter, I had duplicated 2019 and changed the name and url, just removed it from 2022 file.

matheusgomesms:
Seems great! Thanks for your work!

willemarcel (approved):

PR #1808, opened 2023-03-29
Germany: Add basemap.de source layer

der-eismann:


Marc-marc-marc:
it's wanted to keep it as a draft ?

matkoniecz:
@der-eismann What is missing here in this PR? Do you plan to finish it?

der-eismann:
Sorry guys. The thing is I'm unsure about the whole licensing stuff. The maps work fine, but I don't really get their licensing:
https://sgx.geodatenzentrum.de/web_public/gdz/lizenz/deu/basemapde_web_dienste_lizenz.pdf

danieldegroot2:
## License

> (3) Die mit den Diensten übermittelten Daten dürfen nur in Verbindung mit dem jeweiligen Dienst genutzt und nicht gespeichert werden. Eine Entnahme oder Weiterverwendung von Daten außerhalb der Dienste ist nicht erlaubt.

'(3) The data transmitted with the services may only be used in connection with the respective service and may not be stored. **Extraction or further use of data outside the services is not permitted**.'

It seems like tracing or otherwise deriving data is prohibited. Based on (4), only prints and custom rendering styles seem allowed.

> (4) Von den präsentierten Daten dürfen Druckausgaben und/oder abgeleitete Darstellungen (Präsentationsausgaben) erstellt werden.

'(4) **Printouts** and/or derived representations (**presentation** outputs) may be created from the presented data.'

Similarly, license of potentially other `basemap.de` data may have usage restrictions.

## Discussion

> Ministerium für Landesentwicklung und Wohnen:
> > Nach derzeitigem Stand wird es lediglich eine einzige länderübergreifende Basemap-Sicht für alle geben. Daher kann ich Ihnen aktuell keine Sonderkonditionen wie zu Maps4BW-Zeiten anbieten

'As things stand at the moment, there will only be one cross-country basemap view for all. Therefore, I cannot currently offer you special conditions as in Maps4BW times.' 
( Ref https://community.openstreetmap.org/t/maps4bw-und-basemap-de/6762/46 )

There seems to be currently no official permission to use `basemap.de` for OSM.

Sidenote: Not to be confused with `basemap.at`.

cc: @cicku @tyrasd

cicku:
3 is contracting with 4 here if the translation is correct.

This is very common…3 is often created to prohibit redistribution/reselling. 4 means that you can trace based on the provided data, or add some markers for a hard copy.

Not a native speaker so I won’t comment much but I’m sure the license clauses here are confusing right now.

tyrasd:
My understanding is that the licensing of the basemap.de product is not compatible with OSM and while there had been some efforts to come to an agreement (at least for the state of Baden-Württemberg, in analogy to an existing agreement on a "precursor" of basemap.de which is now defunct), these efforts have not been successful for the time being.

Sadly, this means that the source is not suitable to be included in ELI at the moment. :unamused:

der-eismann:
What a pity, but I kind of expected that. Thanks for chiming in!

PR #1859, opened 2023-05-15
Add BEV INSPIRE Orthofoto Layer

hlfan:
Add an ortho imagery layer from Austrias *Federal Office of Metrology and Surveying* which receives updates more frequently than the basemap.at or geoimage.at orthophoto layers.


hlfan:
[OSM-Community-Tread](//community.openstreetmap.org/t/bev-amap-luftbild-nutzung-fur-osm/98892/14) debating the licensing (in german)

PR #1880, opened 2023-05-25
Update OS Map Local to April 2023

SK53:
Update to latest (April 2023) release. Not sure if the {switch..} is still required, but works OK.

### DIFF
--- a/sources/europe/gb/OSOpenMapLocal.geojson
+++ b/sources/europe/gb/OSOpenMapLocal.geojson
@@ -5,7 +5,7 @@
         "name": "OS OpenMap Local (April 2023)",
         "url": "https://{switch:a,b,c}.os.openstreetmap.org/layer/gb_os_om_local_2022_10/{zoom}/{x}/{y}.png",
         "max_zoom": 18,
-        "min_zoom": 1
+        "min_zoom": 1,
         "end_date": "2023-04",
         "country_code": "GB",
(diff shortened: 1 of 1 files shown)


sources/europe/gb/OSOpenMapLocal.geojson
```diff
@@ -1,12 +1,12 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "OS-OpenMap_Local_2022_10",
-        "name": "OS OpenMap Local (October 2022)",
+        "id": "OS-OpenMap_Local_2023_044",
```
UKChris-osm: Should the “044” be “04”, as I think this is the reference to the month.

Marc-marc-marc: or drop any "date" reference ?

SK53: @UKChris-osm yes, should be 04 (I have a problem with the "4" key on my keyboard, which means I sometimes get nothing or it repeats".

@Marc-marc-marc we support all historical layers too, so it is best to label them explicitly.

Jez-C:
It looks like only the name and date were changed here because the layer showing in iD editor is actually the October 2022 release. In any case, the October 2023 release is already on os.openstreetmap.org so hopefully this index can be updated too?

Issue #1893, opened 2023-06-04
[Watchdog] Imagery "Trafikverket Road Network": sources/europe/se/trafikverket-nvdb-main.geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"Trafikverket Road Network" -> [sources/europe/se/trafikverket-nvdb-main.geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/europe/se/trafikverket-nvdb-main.geojson)

Reason:
Error: WMS -: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?SERVICE=WMS&REQUEST=GetCapabilities Error: WMS service exception
Error: WMS 1.3.0: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.3.0 Error: WMS service exception
Error: WMS 1.1.1: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.1 Error: WMS service exception
Error: WMS 1.1.0: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.0 Error: WMS service exception
Error: WMS 1.0.0: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.0.0 Error: WMS service exception
Error: WMS -: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?service=WMS&request=GetCapabilities Error: WMS service exception
Error: WMS 1.3.0: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?service=WMS&request=GetCapabilities&version=1.3.0 Error: WMS service exception
Error: WMS 1.1.1: URL: https://geo-netinfo.trafikverket.se/mapservice/wms.axd/NetInfo?service=WMS&request=GetCapabilities&version=1.1.1 Error: WMS service exception
[...]

Issue #1892, opened 2023-06-04
[Watchdog] Imagery "Trafikverket Road Network options": sources/europe/se/trafikverket-nvdb-options.geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"Trafikverket Road Network options" -> [sources/europe/se/trafikverket-nvdb-options.geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/europe/se/trafikverket-nvdb-options.geojson)

Reason:
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception
Error: Exception: WMS service exception

CC contributors to this imagery:
@grischard, @NKAmapper, @Marc-marc-marc, @Klumbumbus, @rbuffat, @magol

PR #1896, opened 2023-06-05
Update Trafikverket WMS

NKAmapper:
Update with new URL + adding privacy policy.
Syncs with JOSM.

#1892
#1893
#1894


joar:
Works with ID editor as custom layer as well, but it's annoying to switch between custom layer and aerial imagery.

PR #1898, opened 2023-06-06
Standardize LINZ Basemaps URLs

### LINKED ISSUE #1890: WebP support
A few years ago we (LINZ) added some imagery from our basemaps service to repo, at that time we discussed the use of WebP as the format for serving the tiles https://github.com/osmlab/editor-layer-index/pull/913

Since then Safari now supports WebP, Would you accept a pull request to convert these from jpg to webp.

Reasoning:
- Webp is much higher use in our system so it has a much higher cache hit ratio 
- Webp generally is much smaller file size and higher quality 
- Webp supports transparency now that some layers are individual datasets this would be shown as a black outline.

blacha:
LINZ is trying to move it's users to more standard URLs as we have a collection of similar URLs for the same imagery.

I have split this into three commits 

Move from EPSG code to TileMatrix ID, [Standardise LINZ urls](https://github.com/osmlab/editor-layer-index/commit/aaa5acab04e6b57cd8a39a49d89b8083ceb5fc89) We have multiple tile matrices for the same EPSG code so using just a EPSG code can be possibly give users different tile matrix. so we are recommending people to use the TileMatrixId `WebMercatorQuad` over `EPSG:3857` 

Remove "aerial:" [Remove all "aerial:" prefixed imagery as it is a deprecated URL](https://github.com/osmlab/editor-layer-index/commit/c1e70c5e6b2272f17c9d14cd8face2985c6ab3db) Accessing imagery from "aerial:..." is deprecated and is going away "soon" all imagery can be access directly with via it's name. Names can be found in the drop down list on https://basemaps.linz.govt.nz or in https://github.com/linz/basemaps-config

[Convert LINZ imagery to webp](https://github.com/osmlab/editor-layer-index/commit/7e91a8134702c3968d258eef67c366dcff0f476f) This one is optional, webp has much more usage in our system which provides a better experience for most users. Fixes: #1890

### DIFF
--- a/sources/oceania/nz/LINZ-Hawkes-Bay-Cyclone-Gabrielle.geojson
+++ b/sources/oceania/nz/LINZ-Hawkes-Bay-Cyclone-Gabrielle.geojson
@@ -11,7 +11,7 @@
         "privacy_policy_url": "https://www.linz.govt.nz/privacy",
         "category": "photo",
         "best": true,
-        "url": "https://basemaps.linz.govt.nz/v1/tiles/hawkes-bay-cyclone-gabrielle-2023-0.1m/WebMercatorQuad/{zoom}/{x}/{y}.jpg?api=d01egend5f8dv4zcbfj6z2t7rs3",
+        "url": "https://basemaps.linz.govt.nz/v1/tiles/hawkes-bay-cyclone-gabrielle-2023-0.1m/WebMercatorQuad/{zoom}/{x}/{y}.webp?api=d01egend5f8dv4zcbfj6z2t7rs3",
         "max_zoom": 21,
         "country_code": "NZ",
         "type": "tms",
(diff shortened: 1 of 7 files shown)


dwsilk:
cc @k-yle @andrewharvey 

would appreciate it if we could get this merged and released as we would like to clean up our `aerial:` configurations and OSM is the primary user of these

grischard:
@dwsilk when we merge this, it will have to wait until the next https://github.com/openstreetmap/iD release to go live on openstreetmap.org, which might only be in a few weeks.

grischard:
@dwsilk also, congratulations on breaking our CI checks with the very first webp layer in our index :)

dwsilk:
Thanks @grischard. Hah, that change was optional as noted above and in the linked issue, but it is definitely our preference!

PR #1912, opened 2023-06-13
Add Florida DEP Ortho 2021

arch0345:
~~Geometry is copied from 2020, should be modified since the footprint for 2021 includes some new areas and doesn't cover some areas covered in 2020:~~
image

Edit: Geometry has been modified


cicku:
The last time I checked with DEP their data seems to be a mix of good and garbage, will check

arch0345:
> The last time I checked with DEP their data seems to be a mix of good and garbage, will check

When updating the geometry the quality of all the areas covered seems to be fine, seems to all be around 3-6 inch resolution.

PR #1927, opened 2023-06-15
Add NYSDOP CIR Orthoimagery

arch0345:
Also updated year/name for NYSDOP Natural Color Orthoimagery

### DIFF
--- a/sources/north-america/us/ny/NYSDOP_Latest_CIR.geojson
+++ b/sources/north-america/us/ny/NYSDOP_Latest_CIR.geojson
@@ -1359,7 +1359,7 @@
         ]
     },
     "properties": {
-        "id": "NYSDOP_Latest",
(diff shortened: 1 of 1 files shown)


cicku:
@ZLima12 Hi, is that CIR layer useful or not? I checked it few years ago and it appears to be worse on the quality than the Ortho itself.

ZLima12:
@cicku personally, I use the CIR layer a lot (in JOSM). I find it useful for mapping trails through wooded areas, as I feel that there is better contrast between the trail and surrounding areas. It can probably also be used in other situations where you need that extra contrast.

@arch0345 good catch with the outdated date range on the standard layer. I didn't notice those fields when I updated it in the past.

ZLima12:
As far as quality goes, I haven't noticed too much of a difference between the two layers, at least where I usually edit.

PR #1936, opened 2023-06-17
Add KyFromAbove Phase 2

arch0345:


cicku:
Are they overlapping?

arch0345:
No, Phase 1 only includes certain counties while Phase 2 includes the whole state

Edit: If you're talking about the removal of 2019/2020 sources, then yes phase 2 includes these as well as imagery captured in 2018/2021

PR #1863, opened 2023-05-16
Update IndianaMap

arch0345:
Changes source from tms to wms, closes #1850

Issue #1660, opened 2022-11-03
[Watchdog] Imagery "IndianaMap Orthoimagery 2018": sources/north-america/us/in/IndianaMap2018.geojson broken


rbuffat:
Watchdog failed for 5 consecutive days for:
"IndianaMap Orthoimagery 2018" -> [sources/north-america/us/in/IndianaMap2018.geojson](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/north-america/us/in/IndianaMap2018.geojson)

Reason:
Error: WMS None: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetCapabilities
Error: WMS 1.3.0: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.3.0
Error: WMS 1.1.1: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.1
Error: WMS 1.1.0: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.1.0
Error: WMS 1.0.0: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.0.0
Error: WMS None: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?service=WMS&request=GetCapabilities
Error: WMS 1.3.0: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?service=WMS&request=GetCapabilities&version=1.3.0
Error: WMS 1.1.1: Timeout for: https://maps.indiana.edu/arcgis/services/Imagery/Orthos_2018/MapServer/WmsServer?service=WMS&request=GetCapabilities&version=1.1.1
[...]

PR #1941, opened 2023-06-17
Update IndianaMap

arch0345:
Re: #1863

### DIFF
--- a/sources/north-america/us/in/IndianaMap_Latest.geojson
+++ b/sources/north-america/us/in/IndianaMap_Latest.geojson
@@ -9,15 +9,15 @@
         },
         "name": "IndianaMap Orthoimagery - Latest Available",
         "icon": "https://www.arcgis.com/sharing/rest/content/items/c470d38a730b4a5d839602174ed97d5c/resources/ingov-logo-white.png",
-        "url": "https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailableTiles/ImageServer/exportImage?f=image&format=jpg&bbox={bbox}&imageSR={wkid}&bboxSR={wkid}&size={width},{height}&foo={proj}",
+        "url": "https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailable/ImageServer/exportImage?f=image&format=jpg&bbox={bbox}&imageSR={wkid}&bboxSR={wkid}&size={width},{height}&foo={proj}",
         "max_zoom": 19,
-        "license_url": "https://www.in.gov/gis/files/Sanborn-Ortho-Presentation-1_8_2021_.pdf",
+        "license_url": "https://www.in.gov/idoa/proc/bids/RFP-20-051/051_Att%20H%20-%20Summary%20Scope%20of%20Work.docx",
         "country_code": "US",
         "type": "wms",
         "available_projections": [
             "EPSG:3857"
         ],
-        "start_date": "2016",
+        "start_date": "2017",
         "end_date": "2021",
         "description": "The latest available orthoimagery for the entire State of Indiana",
         "category": "photo",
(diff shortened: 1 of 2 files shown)


sources/north-america/us/in/IndianaMap_Latest.geojson
```diff
@@ -11,7 +11,7 @@
         "icon": "https://www.arcgis.com/sharing/rest/content/items/c470d38a730b4a5d839602174ed97d5c/resources/ingov-logo-white.png",
         "url": "https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailableTiles/ImageServer/exportImage?f=image&format=jpg&bbox={bbox}&imageSR={wkid}&bboxSR={wkid}&size={width},{height}&foo={proj}",
```
1ec5: [This ImageServer](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailableTiles/ImageServer) says it covers 2013 to 2017, whereas the `start_date` and `end_date` below say 2016 to 2021. [A different ImageServer](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailable/ImageServer/) says it covers 2016 to 2019. If any ImageServer claims to cover 2016 to 2021, I assume that means it covers both the completed 2016–19 project and the in-progress 2020–23 project.

/ref https://github.com/osmlab/editor-layer-index/issues/1660#issuecomment-1301743485

sources/north-america/us/in/IndianaMap_Latest.geojson
```diff
@@ -11,7 +11,7 @@
...
         "max_zoom": 19,
-        "license_url": "https://gis.iu.edu/dataset/statewide/in_2016.html",
+        "license_url": "https://www.in.gov/gis/files/Sanborn-Ortho-Presentation-1_8_2021_.pdf",
```
1ec5: This slide deck just says the imagery is “[free of charge](https://www.in.gov/gis/files/Sanborn-Ortho-Presentation-1_8_2021_.pdf#page=9)”.

The original statement releasing the 2016–19 Woolpert imagery into the public domain has been [archived by the Wayback Machine](http://web.archive.org/web/20220125235228/https://gis.iu.edu/dataset/statewide/in_2016.html) and should still be valid.

The 2020–23 project is also in the public domain, based on [this request for proposals](https://www.in.gov/idoa/proc/bids/RFP-20-051/051_Att%20H%20-%20Summary%20Scope%20of%20Work.docx) that the state awarded to Sanborn. Page 2, section 1.4.1:

> All orthoimagery and ancillary data products delivered through this contract shall be public domain data without restrictions on distribution or use.  Ownership of the final data products and any ancillary products will reside with the State of Indiana.

sources/north-america/us/in/IndianaMap_Latest.geojson
arch0345: Yeah you're right, it looks like this ImageServer doesn't include 2021:
image
[BestAvailibleTiles](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailableTiles/ImageServer)

image
[2021](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/2021/ImageServer)

As for the other ImageServer mentioned, the description might be outdated. Their [AGOL open data portal](https://www.indianamap.org/datasets/INMap::best-available-imagery-of-indiana/about) suggests that this covers 2017-2021 and sure enough, it seems to cover 2021:

image
[BestAvailible](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/BestAvailable/ImageServer)

image
[2021](https://imagery.gis.in.gov/arcgis/rest/services/Imagery/2021/ImageServer)

sources/north-america/us/in/IndianaMap_Latest.geojson
arch0345: I just copy-pasted the [url used for the 2021 service](https://github.com/osmlab/editor-layer-index/commit/4014bd5969693a6e196be208be6087c5dca5f370). Since this will include both 2017-2019 and 2021, not sure which one to add.

1ec5: I guess the 2020–23 license is more relevant, since the Sanborn imagery will eventually replace the older Woolpert imagery in this layer.

1ec5 (approved):
Thanks for updating these layers! The new imagery layer is quite nice to trace from.

PR #1958, opened 2023-07-01
Add FirstMap 2022

arch0345:


cicku:
What is that parks layer doing? I don't believe it is needed.

arch0345:
> What is that parks layer doing? I don't believe it is needed.

I thought it might be useful since it has a higher resolution (3 inch vs 6 inch)

PR #1996, opened 2023-07-18
added DTM/DSM for Slovakia from ÚGKK SR

zdila:


sources/europe/sk/DTM5-shading.json
```diff
@@ -0,0 +1,209 @@
...
+        "id": "dtm5-sk-shading",
+        "attribution": {
+            "text": "ÚGKK SR",
+            "required": true
+        },
+        "name": "DTM 5.0 shading",
+        "url": "https://dmr5-shading.tiles.freemap.sk/{zoom}/{x}/{y}.jpg",
+        "max_zoom": 18,
+        "country_code": "SK",
+        "best": true,
```
andrewharvey: I could be wrong but I don't think best is per category, rather it hints to editing software what should be the default imagery for this location, and DTM shading shouldn't be the default.

sources/europe/sk/DSM1-shading-copy.json
andrewharvey, line 1: 1. https://www.freemap.sk/#tip=privacyPolicy could be used as the `privacy_policy_url`.
2. Could we remove the `-copy` from the filename?

zdila:
@andrewharvey  I've resolved the issues you had targeted.

PR #2007, opened 2023-07-27
Name change from SDFE to SDFI

JitGub:
Name change from SDFE to SDFI.


cicku:
Any references?

JitGub:
Link(in Danish):
https://sdfi.dk/om-os/nyhedsrum/nyheder/nyhedsarkiv/2022/jun/nyt-navn-nye-opgaver-og-nye-medarbejdere

If the old webpage is entered (https://sdfe.dk/), the new webpage is used (https://sdfi.dk/).

cicku:
Can you also update the license URL? Looks like they all need to be updated as well.

PR #2022, opened 2023-08-09
New NRW and Kreis Viersen layers (Germany)

pathmapper:
This adds:

- iDOP layer for NRW, Germany
- vDOP layer for NRW, Germany
- ALKIS layer for Kreis Viersen, NRW, Germany

The license for all these is [Data licence Germany - Zero - Version 2.0](https://www.govdata.de/dl-de/zero-2-0), which is essentially a translation of CC0. This repository already includes data using this license.

iDOP:
20cm ground resolution orthophotos updated yearly for half the area of NRW

vDOP:
Preprocessing product of DOP in 10cm ground resoluton, updated constantly  

ALKIS:
The data includes PNG renderings of government cadastre data (ALKIS) of buildings, roads, etc. It is usually of very high quality, as the data ultimately comes from official government data and land surveying and has an accuracy of at most a few centimeters.

The data for Kreis Viersen is mostly the same as for NRW (and eventually contributes to the Kreis Viersen part of the NRW data) but the Kreis Viersen data is updated daily (very useful to add new buildings) whereas the NRW data is only updated twice a year, and won't be updated in 2023 due to a data migration.


pathmapper:
Would be much appreciated if someone could merge this. @andrewharvey maybe?

andrewharvey:
Since iDOP and vDOP cover the same area, we should select one to be the "best" / default.

pathmapper:
Thanks for looking into this!

> Since iDOP and vDOP cover the same area, we should select one to be the "best" / default.

Good point, addressed per https://github.com/osmlab/editor-layer-index/pull/2022/commits/54da28db8e0b222b805fdd442cfd1d12e4f4e7f3.

pathmapper:
To make it more clear:

vDOP is temporary available data. It's the preprocessing product of DOP and when the DOP are ready, the vDOP vanish.

So if available, vDOP is the best / most recent ortho but it's always only available for a certain time for a given area.

pathmapper:
Test is failing because of the not reachable license URL which is addressed per https://github.com/osmlab/editor-layer-index/pull/2023.

pathmapper:
Should be good to merge now.

KAMiKAZOW:
So this PR demoted NRW_Ortho. Took me long to find it. This change is bad. I don't know which area the new "best" sources cover but it's clearly not the whole of NRW.

image

image


Clearly only NRW_Ortho delivers good results over the whole of NRW:

image

pathmapper:
> Clearly only NRW_Ortho delivers good results over the whole of NRW

But NRW_Ortho is not necessarily the most up-to-date.

I do agree, the situation is complicated because we are having a lot of choice (which is nice) and what is the "best" depends on ones point of view.

We've created this tool where you are able to compare the current coverage and time stamps (yellow overlay, just zoom in):

https://kreis-viersen.github.io/quattromap/?settings=N4IgtgxiBcDMA0IIAsYgEYBsCGEDWIimAjGgHIBKA6gAQAyArgGYAu6AlpgCY0Bi2AJ3QBTQiAD2paBMwB9SrUasO3GgFlhLbF2wthAOzHiADlOJEATOWo0AbgBEA8gAU%2BgkUavTxchXaeuGlo6eoaIJl4ADAB0AKxEsNa07AFuQqLhid6%2BNiku6praugZGxlkx8SCYACxJNPYAKgDSRrXSpW0VAL5AA#6.79/51.492/7.696

grafik

Here is also an overview map and further infos:
https://www.bezreg-koeln.nrw.de/geobasis-nrw/produkte-und-dienste/luftbild-und-satellitenbildinformationen/aktuelle-luftbild-und-0

grafik

Infos regarding iDOP:

https://www.bezreg-koeln.nrw.de/geobasis-nrw/produkte-und-dienste/luftbild-und-satellitenbildinformationen/aktuelle-luftbild-und-4

Hope this helps!

KAMiKAZOW:
> But NRW_Ortho is not necessarily the most up-to-date.

So? Slightly out of date but existing is a far better experience than the current situation where we have to change the layer every single time.


> I do agree, the situation is complicated because we are having a lot of choice (which is nice) and what is the "best" depends on ones point of view.

No, a layer not covering all of NRW should not be the best layer for all of NRW. It's really not complicated at all.


> Hope this helps!

No, as long as long as the wrong layers are default for regions not covered, this doesn't help at all.

PR #2024, opened 2023-08-09
Add "dl-zero-de/2.0"  to list of compatible licenses

pathmapper:
There are already many sources with this license included in the `editor-layer-index`.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -11,7 +11,7 @@ permission.  Attribution of an OGL source is required to appear in the .geojson 
 
 - :white_check_mark: [Data licence Germany - Zero - Version 2.0 (dl-zero-de/2.0)](https://www.govdata.de/dl-de/zero-2-0)
 imagery sources are directly compatible with tracing on OpenStreetMap and do not
-require any additional permission.  Attribution of a Public Domain source is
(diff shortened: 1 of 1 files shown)


cicku:
I wouldn't say it is very appropriate to mention it here, because the license is not in English. Can you confirm if this is what you mean?

https://spdx.org/licenses/DL-DE-BY-2.0.html

pathmapper:
> I wouldn't say it is very appropriate to mention it here, because the license is not in English.

It is, the English version is at the bottom of the page :-)

>  Can you confirm if this is what you mean?
> 
> https://spdx.org/licenses/DL-DE-BY-2.0.html

No, the one you linked is with attribution and not the right one, seems like it is currently not listed on https://spdx.org/licenses/.

Here is the right one which is also linked in the PR:
https://www.govdata.de/dl-de/zero-2-0



grafik

pathmapper:
@cicku thanks for making me aware, that `dl-zero-de/2.0` isn't listed on https://spdx.org/licenses/.

I've submitted a request to add it -> https://github.com/spdx/license-list-XML/issues/2072

pathmapper:
Thanks @cicku

Issue #1490, opened 2022-03-26
New Layer Request: USGS 3D Elevation Program


OptikalCrow:
These two layers appear to be the most useful of the USGS 3D Elevation Program WMS services. One is a general hillshade which is useful for determining stream flowpaths, track roads through forested areas, and in some areas foot trails. The other is a dynamic contour lines layer useful as an overlay to assess land gradients at finer detail than scanned USGS topographic maps.

[JOSM includes the hillshade layer as an extra option](https://josm.openstreetmap.de/wiki/Maps/USA%20Country#USGS3DElevationProgram) not found in this layer index, so there is some precedent for adding it. It also includes a contour layer from the USFS, although the 3DEP should supercede this layer as the USFS availabilty is only found in USFS-administered areas. 

## URL of the imagery
https://elevation.nationalmap.gov/arcgis/rest/services/3DEPElevation/ImageServer
layers: 
3DEPElevation:Hillshade Multidirectional
3DEPElevation:Contour 25

## Two letter country code
US, CA, MX

## Desired layer name
USGS 3D Elevation Program - Hillshade
USGS 3D Elevation Program - Contour Lines

## License
[Public Domain as listed by [U.S. Code, Title 43 – Public Lands, Chapter 46: GEOSPATIAL DATA](https://uscode.house.gov/view.xhtml?hl=false&edition=2019&path=%2Fprelim%40title43%2Fchapter46&req=granuleid%3AUSC-2019-title43-chapter46&num=0&saved=L3ByZWxpbUB0aXRsZTQzL2NoYXB0ZXI0Ng%3D%3D%7CZ3JhbnVsZWlkOlVTQy1wcmVsaW0tdGl0bGU0My1jaGFwdGVyNDY%3D%7C%7C%7C0%7Cfalse%7Cprelim)](https://uscode.house.gov/view.xhtml?hl=false&edition=2019&path=%2Fprelim%40title43%2Fchapter46&req=granuleid%3AUSC-2019-title43-chapter46&num=0&saved=L3ByZWxpbUB0aXRsZTQzL2NoYXB0ZXI0Ng%3D%3D%7CZ3JhbnVsZWlkOlVTQy1wcmVsaW0tdGl0bGU0My1jaGFwdGVyNDY%3D%7C%7C%7C0%7Cfalse%7Cprelim)

## Is the license compatible with OSM?
Yes

## Attribution (text, URL, required?)
U.S. Geological Survey

## Privacy policy (URL)
[...]

PR #2025, opened 2023-08-12
Add Oregon DOGAMI Lidar

arch0345:
As featured in [Mapping USA 2021](https://www.youtube.com/watch?v=jsrmE8ueYE0). DOGAMI Lidar [is in the public domain](https://www.oregon.gov/dogami/lidar/Documents/DOGAMI_Lidar_Data_Distribution_Policy_20170411.pdf) and includes newer data than the USGS 3DEP source added in #1490. Ideally the geometry should be refined before merging:
image


cicku:
Why is the data not merged into 3DEP?

andrewharvey:
If it's still the case that this new data is not present in the existing USGS 3DEP source, then I would say we go ahead and merge it. Can always remove it later if it does become incorporated into the existing source.

arch0345:
Yeah, this is still the case. In the Portland Metro area, DOGAMI is using data from 2019 while 3DEP is using data from 2007. The 3DEP LidarExplorer website lists the 2019 data as 'pending publication'. Not sure when it'll get merged

PR #2028, opened 2023-08-13
Create Hamburg-ALKIS.geojson

G4rdener:
added Hamburg ALKIS layer adter reading "DE:Permissions/LGV Hamburg 2021"  and short discussion in the OSM-HH-Telegram


cicku:
What `short discussion in the OSM-HH-Telegram`?

G4rdener:
We talked about whether we are allowed to use the ALKIS layer or we have to ask that again. But we came to the conclusion that we are allowed to use it, because we had mentioned it explicitly in our request and the answer from LGV was that we are allowed to use all maps/layers.

PR #2033, opened 2023-08-17
Add fresh Berlin 2023 images

tordans:
Same license and all as before.


tordans:
@polarbearing FYI

cicku:
Can you drop any file prior to 2018? Also you do not need to specify best:false, which means nothing.

tordans:
> Can you drop any file prior to 2018? 

Actually, we still use those. They all have different shadow and lighting and – depending on the mapping, they are helpful. Why do you want them dropped? And why part of this PR?

> Also you do not need to specify best:false, which means nothing.

I find it helpful when all files follow the same structure and data is specified explicitly.

tordans:
@cicku it would be great if you could merge this. We invested a lot of effort to get the fresh images from official sources and prepare them on a community maintained server so people can use it.  IMO there  is nothing holding this PR back.

svenpilz:
I would also like to keep the images from 2016 until now. As @tordans said, sometimes the older images show have less tree cover or shadows. I use the 2016/2017/2028 often for sidewalks.

andrewharvey:
I think `best: false` is fine, it's being explicit to avoid doubt.

> Actually, we still use those. They all have different shadow and lighting and – depending on the mapping, they are helpful.

Agreed.

tordans:
@andrewharvey thanks a lot for merging those PRs!

PR #2034, opened 2023-08-17
Berlin "other" layer

tordans:
Those are three data layer that we can use for Berlin. They are city data that the open data community gives access to via a map proxy (https://github.com/codeforberlin/mapproxy-config).

The layer are in the "other" category, since that seems to fit most.

I did not add them before, because I was not sure if users will understand that those layers can be used but still need to be verified with local knowledge. However, I now think we should trust our users with this power :-).


tordans:
Update: I just learned that the URLs of those sources will change in the month to come. More at https://github.com/codeforberlin/mapproxy-config/issues/17

I would still vote to merge this now and update it as soon as the new sources are ready.

tordans:
@andrewharvey may I ask you to review and merge this. We have those layers in use in JOSM and via id editor custom background layers for years. This just makes the status quo easier to access. Licenses and such are valid and did not change at any time.

PR #2036, opened 2023-08-21
[Lyon] Fix some metadata in layers

cnotin:
Some metadata, and more importantly, WMS, URLs don't work anymore so I've fixed them with new URLs that were provided by Métropole de Lyon.
I've also added the new 2022 layer, also provided by Métropole de Lyon. I also discovered that JOSM has this layer, and the fix for the others, described on https://josm.openstreetmap.de/wiki/Maps/France#LyonOrthophoto2022


cicku:
Can you drop the 2012 layer?

cnotin:
Oops indeed there was a mistake in the URL. It's fixed with the latest commit. Is that what you meant?

andrewharvey:
@cnotin Sorry I missed this one and just merged #2061 which incorporated some of your changes. Normally I would work with the earlier contribution first, but I missed it.

Could you check if any further changes are needed and either resolve the conflicts or close? Thanks.

cnotin:
Oh damn! I've fixed the conflicts by keeping only the relevant changes

PR #2043, opened 2023-09-01
Add Sarasota Ortho 2022/23

arch0345:


andrewharvey:
It doesn't appear that this endpoint is publicly accessible, can you confirm @arch0345 ?

arch0345:
It looks like they got rid of the WMTS endpoint for 2022, however the one for 2023 still seems to work

PR #2048, opened 2023-09-03
Update URL for CAPCOG 2020

### LINKED ISSUE #2046: US / StratMap CapArea & McLennan Imagery (Natural Color 2020 no longer working
The "US / StratMap CapArea & McLennan Imagery (Natural Color 2020)" imagery no longer works. Tiles return HTTP 400 error. The 2021 version still works, but has some areas which do not actually load properly at higher zooms. Should the 2020 option be removed?

arch0345:
Closes #2046


cicku:

PR #2049, opened 2023-09-03
Drop best flag for Texas imagery sources

arch0345:
Bing in these areas are more recent than these sources. I also changed the category of these sources to 'historicphoto'

### DIFF
--- a/sources/north-america/us/tx/Brazos_County_Orthophoto_2019_WMS.geojson
+++ b/sources/north-america/us/tx/Brazos_County_Orthophoto_2019_WMS.geojson
@@ -9,7 +9,6 @@
         "start_date": "2019",
         "end_date": "2019",
         "country_code": "US",
-        "best": true,
         "attribution": {
             "url": "https://data.tnris.org/collection/aa2cd74e-9c2d-4f00-bae5-609b5e898093",
             "text": "Strategic Mapping Program (StratMap). Brazos County Imagery, 2019-01-29",
@@ -20,7 +19,7 @@
             "EPSG:3857",
             "EPSG:4326"
         ],
-        "category": "photo",
+        "category": "historicphoto",
         "privacy_policy_url": "https://tnris.org/site-policies/#privacy-and-security-policy",
         "min_zoom": 3
     },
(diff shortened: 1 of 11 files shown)


cicku:

bhousel:
> We do not drop best just because Bing provides newer imagery

actually yes, that is a good reason to drop the `best` flag.

andrewharvey:
> We do not drop best just because Bing provides newer imagery, based on that standard most of the best flag will be dropped.

I agree that we should try to balance imagery clarity and currency so it's not always easy to say which is better, but in general if Bing is has reasonable resolution to trace building outlines and is more recent than other sources, then it should be used as the default, and to achieve that we should drop best from the other sources.

arch0345:
Bing in these areas is of similar resolution to these other sources. While the 2021 CAPCOG imagery is newer west of Austin (captured 2019/2021), Bing is newer in all of the other areas covered (captured 2022/2023). I think it's more important to have newer imagery as default in these more urban areas since there's likely more development going on.

Edit: fixed areas/dates after further investigation

andrewharvey:
If this is supported by the local mapping community and not disputed, then no issues.

Though the license URLs appear to be 404ing now, causing the check to fail.

arch0345:
I've gone ahead and created [a thread on the OSMUS Slack](https://osmus.slack.com/archives/C0140BW761Z/p1697693497429759). As for the license urls, they seem to be incorrectly returning 404s. Visiting the url still shows the page (ie https://betadata.tnris.org/collection?c=3e1530dc-d5be-4683-811b-9d4b162f0aa9).

arch0345:
So far there hasn't been any consensus on whether or not this layer should be default. To quote other mappers on that thread: while there are issues with blank tiles appearing at high zooms (since it's a mosaic of three different sources w/ varying resolutions), this layer is better orthorectified compared to Bing.

[I'm trying to get OSMF to host more up to date imagery from CAPCOG](https://osmus.slack.com/archives/C2SNNUNUV/p1693850301971879?thread_ts=1693850301.971879&cid=C2SNNUNUV), which should resolve all three of these issues

grischard:
So once we get capcog imagery running, we're ok to drop this?

grischard:
Test CAPCOG cloud-optimised geotiff at https://grant.dev.openstreetmap.org/capcog-2022-nc-cir-12in/tiles/mosaic.json sent to a few editor developers

PR #2050, opened 2023-09-03
Fix ACC 2018

arch0345:
* Fixes discrepancy between file name and id
* Changed name to be more specific
* Change category to historicphoto as this source is 5-6 yrs old
* Removes best flag as Bing is newer in this area
* Fixes start/end dates
* Updates license/privacy policy urls

### DIFF
--- a/sources/north-america/us/ga/ACC_2018.geojson
+++ b/sources/north-america/us/ga/ACC_2018.geojson
@@ -1,13 +1,12 @@
 {
   "type": "Feature",
   "properties": {
-    "id": "2018_Ortho_Imagery",
-    "name": "2018 Ortho Imagery",
-    "category": "photo",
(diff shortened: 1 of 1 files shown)


cicku:

arch0345:
What are the qualifications for a source having the `best` flag then? As iD uses sources with the best flag as default, inexperienced users mapping in Athens-Clarke County could be mapping with imagery from 2017-2018 instead of similar quality imagery from February 2023 (this currently isn't the case the file name/id having different values causes this source not to show up in iD)

grischard:
'best' is a relic from ages ago when there was a handful of layers. If the 2017-2018 imagery has no redeeming qualities other than being local, we should drop the 'best' qualifier from it.

PR #1649, opened 2022-10-21
OpenAerialMap Mosaic by Kontur.io

Zkir:
OpenAerialMap Mosaic by Kontur.io has been added.
This is newly developed mosaic layer, which contains all the avalible OAM imagery in one layer. 
New images are displayed over older ones, min/max zooms are tracked.

PR #2052, opened 2023-09-04
OpenAerialMap-Mosaic isn't an overlay

simonpoole:
This is not an overlay.


andrewharvey:
Agreed this isn't an overlay. @Zkir any rational behind setting it as an overlay in your original contribution at https://github.com/osmlab/editor-layer-index/pull/1649?

andrewharvey (approved):

PR #2060, opened 2023-09-16
Rotate OSM US Mapbox credentials

iandees:
Our credentials are being used by someone that shouldn't use it, so we're rotating the access token.


simonpoole:
I'm fairly sure that all larger consumers of ELI support "private" API keys now days, so you could potentially remove them from here at the expense of having to hand them out to editor devs. At least that makes things a bit more expensive that simply checking a github repo.

iandees:
Thanks for the pointer. I know that this happens/happened for the Maxar sources, so I'll take a look at how those work.

PR #2134, opened 2023-11-13
MassGIS LiDAR Slope imagery

petercooperjr:
This imagery from MassGIS uses LiDAR data to show the variation in elevation (steepness) at each point. White represents something that is flat, with progressively darker values meaning steeper slope, up through black which represents a completely vertical slope. The source LiDAR data is from 2021 for the central and eastern parts of Massachusetts, 2018 for Nantucket, and between 2013 and 2015 for the western parts of the state. They have many details of the dataset at https://www.mass.gov/info-details/massgis-data-lidar-terrain-data. All MassGIS data is in the public domain.


cicku:
Are they not submitted to USGS 3DEP?

petercooperjr:
I'm not sure I understand the question, I'm a bit out of my depth trying to submit this. I know MassGIS works with USGS somehow, so if the slope visualization is available from USGS then that would be fine. I haven't seen an option in iD that gives me anything like what having `https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/Slope_Grayscale_2013to2021/MapServer/tile/{zoom}/{y}/{x}` as a custom background gives me, though, to be able to visualize terrain features under treecover.

It's a lot like the existing Shaded Relief layer (which just got updated to the current data in a separate pull request), but this visualization is focused more on the changes in elevation at a given point.

petercooperjr:
So I've done a little more research. Yes, the MassGIS data was jointly funded with USGS, and is shared with their 3DEP program.  I don't think the USGS offers the Slope visualization from the data, though. (And the recently-updated "MassGIS LIDAR Shaded Relief" layer I think looks sharper and more detailed than the "USGS 3D Elevation Program" layer, though I don't pretend to understand the technical differences between how they were each created from the underlying data.)

Does that answer your question?

grischard:
@petercooperjr it does, thank you!

PR #2141, opened 2023-11-20
Remove Sumter 2021

arch0345:
Unfortunately, it looks like Sumter County removed public access to their 2021 imagery. Since their 2020 imagery can already be viewed via Esri World Imagery and FDEP 2020, I decided not to add it.


grischard:
Thank you!

PR #2152, opened 2023-11-28
New layers from CUZK (Czech Republic cadastral office) 

Vlada-C:
Some very useful layers from CUZK (State Administration of Land Surveying and Cadastre) have become legally available for OSM mapping. The OSM CZ have managed to get an explicit approval on top of a CC-BY-4.0 license for these layers. This pull request is to add an excellent orthophoto of the whole Czech Republic. I believe that it is the best option for the region in terms of alignment accuracy, resolution and update frequency, so I have marked it accordingly. The earlier added IPR aerial imagery is even better but that one is limited to the area of Prague only. The second layer to be added is a digital model of the surface of the terrain. The description is inside the file.


Vlada-C:
The missing image (causing the automated check to fail) is actually included, but it has not been available in the final URL because the pull request has not been merged yet. Should I change the URL to a relative link?

zbycz (approved):
Looks ok, thanks! :) Though i am not a project maintainer, so my approve probably doesnt mean anything.

matheusgomesms:
Hi @Vlada-C

Thanks again for your work! Everything looks good to me, and don't worry about the icon error, it is normal, since it's not uploaded yet.

My only slight concern is the license. I know CZ community is very active and you have ČÚZK data in use since a long time ago, but for these images do you have the PDF/e-mails where they allowed the use? I might have overlooked, but I didn't find this document in the provided wiki link (https://wiki.openstreetmap.org/wiki/Cs:Česko/freemap#ČÚZK).

If you can also add this direct document link to the Contributors page, it would be even better.

I can gladly merge this PR once you provide that, thanks!

Vlada-C:
Thanks for the review, @matheusgomesms. I have updated the [Cs:Česko/freemap](https://wiki.openstreetmap.org/wiki/Cs:%C4%8Cesko/freemap#%C4%8C%C3%9AZK) wiki page with a link to a relevant thread on the talk-cz mailing list that, hopefully, provides enough of additional information regarding the license.

matheusgomesms:
Looks great, it's even on the [OSM Copyright](https://www.openstreetmap.org/copyright) page. LGTM, thanks!

PR #2162, opened 2023-12-04
Update Stamen Terrain Background

### LINKED ISSUE #2156: Update Stamen layers
Per the [announcement from Stamen](https://maps.stamen.com/stadia-partnership/), the old Fastly CDN URLs have been shut off for a bit. We have a [migration guide](https://docs.stadiamaps.com/guides/migrating-from-stamen-map-tiles/) that covers the necessary steps to get this back up and running.

TL;DR steps:

1. Change the URL pattern to something like https://tiles.stadiamaps.com/tiles/stamen_terrain_background/{z}/{x}/{y}{r}.png (note that these tiles have ben updated and support retina/hidpi now!)
2. Update attribution / license information as noted (Stamen, Stadia Maps, and OpenMapTiles need attribution + links now).

I actually came across this repo a while ago and meant to open a PR last month, but somehow missed it so dropping an issue now to track it since it's the root cause of https://github.com/facebook/Rapid/issues/1241.

ianthetechie:
Closes #2156.

I think I've made this PR correctly, but I couldn't find any instructions on exactly what the local build process was, so I reverse engineered a bit from the GitHub Actions workflows. It looks like some auto-generated files need to be checked into git. I also don't have the `tx` binary available.

### DIFF
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -808,7 +808,8 @@ en:
       name: 'Powiat stalowowolski: Buildings'
     stamen-terrain-background:
       attribution:
-        text: Map tiles by Stamen Design, under CC BY 3.0. Data by OpenStreetMap, under ODbL
+        text: © Stadia Maps © Stamen Design © OpenMapTiles © OpenStreetMap contributors
+      description: A background layer focused on terrain types (via OSM and ESA data) and elevation (hillshading of open elevation datasets via Joerd).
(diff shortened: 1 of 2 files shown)


grischard:
Thank you! For any further commits: no need to compile the imagery files manually.

cicku:
We are not going to merge this PR unless you remove the junk files because they conflict with the master.

PR #2170, opened 2023-12-15
Update date of Stuttgart-latest

Rilele:
The imagery Stuttgart-latest seems to be from 2023, although iD says it's from 2021. Compare to https://maps.stuttgart.de/stadtplan/ . The imagery is definitive the one from 2023 not from 2021. Look at https://www.openstreetmap.org/edit#map=20/48.73033/9.14713 for comparision.


cicku:
If this imagery serves as the "latest available", then the year should be removed from its name. Can you confirm?

Rilele:
I removed the year form the name.
What about "start_date" and "end_date"?

cicku:
You can keep the information there, they are useful metadata.

Issue #1943, opened 2023-06-18
"private" api key now required for Mapbox sat and locator layers


simonpoole:
This https://github.com/osmlab/editor-layer-index/commit/cc3f64158fac656c9c09fc9e894f1362e29c2ea1 seems to have gone nearly unnoticed. @tyrasd was there any indication from Mapbox if they would be prepared to provider additional keys? Or should we just remove the layers from ELI?

See also https://github.com/MarcusWolschon/osmeditor4android/issues/2323

tyrasd:
Hey. Ohhh, I'm very sorry: I totally forgot about Vespucci when making the round of contacting existing "big" users of the old key (it didn't help that Vespucci didn't show up in the list of referrer domains in the dashboard, but I should definitely have remembered that you'd be using it as well). :bow: 

Mapbox were actually quite nice to give us (i.e. _me_, for now) access to a special account which can create virtually unlimited keys for all kinds of OSM tools. I'll send you a key for Vespucci in a minute.

PR #2173, opened 2023-12-18
Update list of editors using ELI

arch0345:
Vespucci seems to still be using ELI (#1943)


cicku:
It switches between ELI and JOSM index. Since it has a mention of us I will merge the PR.

https://github.com/MarcusWolschon/osmeditor4android/blob/master/src/main/java/de/blau/android/contract/Urls.java#L44

PR #2180, opened 2023-12-21
Update NC OneMap Latest

arch0345:
Unfortunately the cached service doesn't include the new imagery from 2023, so I've moved the url back to the WMS server


cicku:
I do not approve this change, will wait for the state to update.

arch0345:
Looks like they finally updated the cached layer, this PR now just updates the start/end dates

PR #2194, opened 2024-01-01
Add NOAA coastal imagery 2023

arch0345:


arch0345:
I've found a WMTS endpoint that has all ten of these combined into a single mosaic. Converting to a draft for now

andrewharvey (approved):
looks good

PR #2196, opened 2024-01-02
Add San Francisco Ortho 2022

arch0345:


cicku:
What is the relation between the gov website and 3dg.is?

arch0345:
3dg.is points to an IP address owned by the San Francisco Department of Telecommunications and Information Services: (https://www.nslookup.io/domains/3dg.is/webservers/)

PR #2206, opened 2024-01-02
Add Cuyahoga Ortho 2023

arch0345:
Cuyahoga County's 2023 imagery was captured under OSIP III (as listed on [OGRIP's download portal](https://gis1.oit.ohio.gov/geodatadownload/)), which is in the public domain.


skorasaurus:
Thank you for adding. Can confirm that the imagery is from this time and has not offset (As expected) ; and very useful =)

PR #2211, opened 2024-01-04
Add Fairfield Ortho 2023

arch0345:
Fairfield's 2023 imagery was captured under OSIP III (Butler County's partial 3IN 2023 imagery as listed on [OGRIP's download portal](https://gis1.oit.ohio.gov/geodatadownload/)), which is in the public domain. This is different than the countywide mosaic being added in #2207, which is only 6IN


cicku:
2020 URL has changed, can you fix the issue?

PR #2217, opened 2024-01-14
Add NAIP HI & PR/USVI

arch0345:


arch0345:
@cicku when you have the time, could you please review this PR and #2218? Thanks

PR #2231, opened 2024-01-30
Buenos Aires City 2021 othophotos

Zalitoar:


matheusgomesms:
Could you please add the following info for both layers?

"end_date": "2017",
"start_date": "2017",

(and the same for 2021 layer)

Gracias!

Zalitoar:
@matheusgomesms end and start dates added in #e1ff0e5

PR #2256, opened 2024-03-05
Add TNMap Ortho

arch0345:
I confirmed with the TNMap team that their imagery is public domain


cicku:
TDOT imagery is always public domain but the update freq is pretty low.

Back to the submission, any reason to have double slashes?

arch0345:
> Back to the submission, any reason to have double slashes?

Looks like an oversight on my end, fixed it in https://github.com/osmlab/editor-layer-index/commit/652f94d7d65f05ab96670dc4f2c45be85f7bf078

PR #2265, opened 2024-03-24
Update VCGI imagery dates to 2021-2023

zekefarwell:
Just a small change updating the dates on these VCGI imagery layers that have been updated with newer imagery as shown here:
https://maps.vcgi.vermont.gov/apps/OrthoStatus/VTORTHO_Acquisition_Vintage5_Final.pdf


cicku:
Thanks!

PR #2269, opened 2024-03-27
Update SouthAfricaCDNGIAerial.geojson URL

Firefishy:
Update URL to match current preferred version.

### DIFF
--- a/sources/africa/za/SouthAfricaCDNGIAerial.geojson
+++ b/sources/africa/za/SouthAfricaCDNGIAerial.geojson
@@ -3,7 +3,7 @@
     "properties": {
         "id": "South_Africa-CD_NGI-Aerial",
         "name": "South Africa CD:NGI Aerial",
-        "url": "http://{switch:a,b,c}.aerial.openstreetmap.org.za/ngi-aerial/{zoom}/{x}/{y}.jpg",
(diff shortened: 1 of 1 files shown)


Manu1400:
## i.e.
* a title: https://aerial.openstreetmap.org.za/layer/ngi-aerial/7/69/74.jpg
* map: https://aerial.openstreetmap.org.za/

Manu1400:
@Firefishy please keep `{switch:a,b,c}` in the URL

grischard:
@Manu1400 http 2 makes this an anti-pattern. We recommend against it on tile.openstreetmap.org too.

Manu1400:
> @Manu1400 http 2 makes this an anti-pattern. We recommend against it on tile.openstreetmap.org too.

## on tile.openstreetmap.org
"Now that tile.openstreetmap.org supports HTTP/2 + HTTP/3 the old (a|b|c).tile.openstreetmap.org is no longer"
https://github.com/Leaflet/Leaflet/pull/8418

Firefishy:
> @Firefishy please keep `{switch:a,b,c}` in the URL

Since this targets primarily clients which use HTTP/2, my preference as the person who runs the service `aerial.openstreetmap.org.za` service is to no longer support the old `{switch:a,b,c}.` HTTP/1.1 hack which was used increase max browser connections. HTTP/2 multiplexing makes the hack redundant.

PR #2288, opened 2024-04-26
de/NRW: Orthophoto is best, vDOP is not

### LINKED ISSUE #2262: Reinstate NRW_Ortho as best for NRW
I don't know which area the new "best" sources cover but it's clearly not the whole of NRW:

image

image

Clearly only NRW_Ortho delivers good results over the whole of NRW:

image

### LINKED ISSUE #2273: NRW vDOP is blank while listed as best imagery for at least some areas
https://www.openstreetmap.org/edit?editor=id#map=19/51.18137/7.17874

screen02

crackwitz:
resolves #2262

might not address #2273 because the outcome of that issue *could* be to remove the vDOP entry entirely

### DIFF
--- a/sources/europe/de/NRW_ortho_wms.geojson
+++ b/sources/europe/de/NRW_ortho_wms.geojson
@@ -9,7 +9,7 @@
         "id": "nrw_ortho_wms",
         "description": "RGB-/Colorbild - Normale farbliche Darstellung.",
         "country_code": "DE",
-        "best": false,
+        "best": true,
         "category": "photo",
         "available_projections": [
             "EPSG:3034",
--- a/sources/europe/de/NRW_vdop_wms.geojson
+++ b/sources/europe/de/NRW_vdop_wms.geojson
@@ -9,7 +9,7 @@
         "id": "nrw_vdop_wms",
         "description": "10cm Bodenauflösung, Zwischenergebnisse aus dem Herstellungsprozess der Digitalen Orthophotos (DOP)",
         "country_code": "DE",
-        "best": true,
+        "best": false,
         "category": "photo",
         "available_projections": [
             "EPSG:3034",


tordans:
@grischard you reviewed an merged https://github.com/osmlab/editor-layer-index/pull/2277 so quickly that I hope you might have time to look at this as well.

- [Example URL with broken "best" images](https://ideditor-release.netlify.app/#background=nrw_vdop_wms&disable_features=boundaries&map=18.14/50.77696/6.08326)
- [Example URL with new "best" images](https://ideditor-release.netlify.app/#background=nrw_ortho_wms&disable_features=boundaries&map=18.14/50.77696/6.08326)

sources/europe/de/NRW_ortho_wms.geojson
```diff
@@ -9,7 +9,7 @@
         "id": "nrw_ortho_wms",
         "description": "RGB-/Colorbild - Normale farbliche Darstellung.",
         "country_code": "DE",
-        "best": false,
+        "best": true,
```
pathmapper, line 12:
```suggestion
        "best": false,
```

How do you define "best"?

For half of NRW currently the `nrw_idop_wms` imagery (green area on the following screenshot) is more recent (~ 1 year) than `nrw_ortho_wms` (blue area).

So maybe don't assign "best" to any source here.

See also 
https://github.com/osmlab/editor-layer-index/pull/2022#issuecomment-2006064942
https://github.com/osmlab/editor-layer-index/issues/2273#issuecomment-2080384564

image

tordans: Something that is missing cannot be "best". If multiple are present, then "best" is the most useful in terms of quality and currentness.

pathmapper: > Something that is missing cannot be "best".

Others might argue:

- Something that is not the most recent cannot be "best".
- Something that has not the best resolution cannot be "best".

Until there is a clear definition/rule for "best" based on consensus, this is a never ending game.

As mentioned in https://github.com/osmlab/editor-layer-index/issues/2273#issuecomment-2080384564:

> Given the lack of a clear definition for what is "best", I currently won't assign "best" to any of the three sources.

crackwitz: I don't see any need to open a policy debate over the attribute. Nor do I see any reason to decide anything for eternity. Surely one can simply ***update the data***, when imagery changes again in the future.

This PR merely compares `dop` and `vdop`, where clearly `vdop` is mostly empty. I would think `vdop` is clearly less valuable than `dop` at this moment. When it did have imagery, it was blurrier than `dop` too, so it was "better" in terms of age but not in terms of (optical) resolution.

This PR does not concern itself with `idop`, so please discuss that elsewhere. Do not use it to delay this PR. If you want `idop` to be featured/"best", you'd have to keep its bounds fresh on an almost daily basis... and that is not considering how long it might take for any commit here to become effective in _osm.org/edit_

pathmapper: I don't like the tone in whole parts of this discussion.

> This PR does not concern itself with idop, so please discuss that elsewhere. 

IMHO you can't discuss this matter entirely without taking `idop` into account.

> Do not use it to delay this PR.

It's not my intention to delay this PR. I was asked by @tordans in https://github.com/osmlab/editor-layer-index/issues/2273#issuecomment-2080059482 for my opinion.

As mentioned in https://github.com/osmlab/editor-layer-index/issues/2273#issuecomment-2080430233:
"I don't have a strong opinion here, go with `dop` as best if you want."

PR #2301, opened 2024-05-05
Adding 3 Serbia layers

stalker314314:
There are 3 layers here that we think should be used:
* Transparent layer that we use to import addresses and street on [DINA](https://dina.openstreetmap.rs/) (as per [wiki page](https://wiki.openstreetmap.org/wiki/Serbia/Projekti/Adresni_registar#Preparing_the_environment)). This data is open and part of https://openstreetmap.org/copyright already. You can try this overlay [here](https://vector-rgz.openstreetmap.rs/styles/rgz-big/#20/46.09371/19.65337).
* Raster layer based on Carto of official Serbia map. This is what we offer as community, per https://openstreetmap.rs/o-mapi/. Maps comes in two scripts (they are both official in Serbia) - cyrillic and latin. You can try both of them [here](https://tiles.openstreetmap.rs/). That practically means:
  * including Kosovo territory as part of Serbia,
  * making `name:sr` (or `name:sr-Latn`) as default label, and
  * removing Russian version of cyrillic italic font (Serbian and Russian italic differ)

I am only not sure what to put for "available_projections" (I omitted it) - one layer is based on maptiles vector server and other is mapnik/carto raster style.


grischard:
@stalker314314 please see the CI log for fixes.

Since this index is for imagery for OpenStreetMap editors, and https://tiles.openstreetmap.rs/ shows OpenStreetMap data, I'm not sure that the layers are needed here. Definitely yes on the address layer.

> Serbian and Russian italic differ

I didn't know that, interesting!

stalker314314:
Fixed CI errors, I hope. Regarding Serbian OSM rendering, I will let @pedjas reply. I sometimes use OSM as a layer in JOSM (it is convenient to locate yourself or find what you need visually), and this Serbian OSM rendering would serve same purpose, especially if you want to contribute in Kosovo (cultural heritage etc.), but Albanian is not your primary language and you can use this render to help yourself. Again - better to let @pedjas comment

pedjas:
> Since this index is for imagery for OpenStreetMap editors, and https://tiles.openstreetmap.rs/ shows OpenStreetMap data, I'm not sure that the layers are needed here. Definitely yes on the address layer.


https://tiles.openstreetmap.rs/ shows customized OSM data. It is not the same as OSM. It is bit adjusted to allow showing map tiles uniformly in Serbian language in both Serbian Cyrillic and Serbian Latin Scripts and some other content specific for mapping Serbia. For example: it gives priority to name:sr and name:sr-Latn tags instead of name tag. 

It is intentionally rendered matching standard OSM tile style, but that is just a look, content differs. This is unique map we need as background for easier editing in Serbian language.

stalker314314:
@grischard Friendly ping:) I guess @pedjas provided explanation - this can be used as a QA tool to spot check where Serbian is missing (my interpretation). If you are still not in favor of it, please let us know!

stalker314314:
Another friendly ping!:)

stalker314314:
Another ping, @grischard . I see @arch0345  is also active, maybe they can chime in?

arch0345 (approved):
Looks good to me, unfortunately I don't have write access so I can't merge this

PR #2310, opened 2024-05-05
Add Eureka Ortho 2023

arch0345:


cicku:
Isn't max_zoom 21 enough?

arch0345:
I'm not really sure how to find the min/max zooms for sources that aren't tms so I'll take your word for it

PR #2312, opened 2024-05-07
Add GeoScribble WMS overlay

Zverik:
See https://wiki.openstreetmap.org/wiki/GeoScribble — for editors that don't support the geometric notes yet.

Things to discuss:

* There is no explicit license mention, but the wiki page states it's PDDL. Should I put this license note somewhere else?
* There are two layers: limited to 3 months and 3 weeks of notes. Does 3 weeks work?
* `default=true`?
* `i18n=true`?

### DIFF
--- a/sources/world/GeoScribbles.geojson
+++ b/sources/world/GeoScribbles.geojson
@@ -7,7 +7,7 @@
             "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
         },
         "description": "Geometric survey notes from Every Door editor, aged three weeks at most",
-        "i18n": false,
+        "i18n": true,
         "id": "geoscribble-latest",
         "available_projections": [
             "EPSG:3857",
--- a/sources/world/GeoScribbles.geojson
+++ b/sources/world/GeoScribbles.geojson
@@ -6,7 +6,7 @@
             "text": "GeoScribble contributors",
             "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
         },
-        "description": "Post-processed Sentinel Satellite imagery.",
+        "description": "Geometric survey notes, aged three weeks at most",
         "i18n": false,
         "id": "geoscribble-latest",
         "available_projections": [
--- a/sources/world/GeoScribbles.geojson
+++ b/sources/world/GeoScribbles.geojson
@@ -1,0 +1,29 @@
+{
+    "geometry": null,
+    "properties": {
+        "attribution": {
+            "required": false,
+            "text": "GeoScribble contributors",
+            "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
+        },
+        "description": "Post-processed Sentinel Satellite imagery.",
+        "i18n": false,
+        "id": "geoscribble-latest",
+        "available_projections": [
+            "EPSG:3857",
+            "EPSG:4326"
+        ],
+        "max_zoom": 22,
+        "name": "GeoScribble latest notes",
+        "type": "wms",
+        "default": false,
+        "overlay": true,
+        "transparent": true,
+        "category": "other",
+        "valid-georeference": true,
+        "url": "https://geoscribble.osmz.ru/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=latest&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
+        "license_url": "https://wiki.openstreetmap.org/wiki/GeoScribble",
+        "privacy_policy_url": false
+    },
+    "type": "Feature"
+}
--- a/sources/world/GeoScribbles.geojson
+++ b/sources/world/GeoScribbles.geojson
@@ -6,7 +6,7 @@
             "text": "GeoScribble contributors",
             "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
         },
-        "description": "Geometric survey notes, aged three weeks at most",
+        "description": "Geometric survey notes from Every Door editor, aged three weeks at most",
         "i18n": false,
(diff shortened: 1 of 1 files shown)


tordans:
> * There is no explicit license mention, but the wiki page states it's PDDL. Should I put this license note somewhere else?

I suggest making it a bit more explicit by adding a section to the wiki and changing the link to https://wiki.openstreetmap.org/wiki/GeoScribble#licence 

> * There are two layers: limited to 3 months and 3 weeks of notes. Does 3 weeks work?

This is the age of the data that will show up? I think 3 month should be better since there will not be a lot of data to start with and moving from mobile to a different editor can take a bit of time…

> * `default=true`?
> * `i18n=true`?

What are those?

sources/world/GeoScribbles.geojson
```diff
@@ -0,0 +1,29 @@
+{
+    "geometry": null,
+    "properties": {
+        "attribution": {
+            "required": false,
+            "text": "GeoScribble contributors",
+            "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
+        },
+        "description": "Geometric survey notes, aged three weeks at most",
```
tordans:
```suggestion
        "description": "Geometric survey notes from Every Door editor, aged three weeks at most",
```

sources/world/GeoScribbles.geojson
```diff
@@ -0,0 +1,29 @@
...
+        "i18n": false,
+        "id": "geoscribble-latest",
+        "available_projections": [
+            "EPSG:3857",
+            "EPSG:4326"
+        ],
+        "max_zoom": 22,
+        "name": "GeoScribble latest notes",
```
tordans, line 17:
```suggestion
        "name": "GeoScribble latest notes (Every Door)",
```

sources/world/GeoScribbles.geojson
```diff
@@ -0,0 +1,29 @@
...
+        "type": "wms",
+        "default": false,
+        "overlay": true,
+        "transparent": true,
+        "category": "other",
+        "valid-georeference": true,
+        "url": "https://geoscribble.osmz.ru/?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&LAYERS=latest&STYLES=&SRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
```
tordans, line 24: This is what this URL would look like in iD https://ideditor.netlify.app/#background=custom:https://geoscribble.osmz.ru/%3FFORMAT%3Dimage/png%26TRANSPARENT%3Dtrue%26VERSION%3D1.1.1%26SERVICE%3DWMS%26REQUEST%3DGetMap%26LAYERS%3Dlatest%26STYLES%3D%26SRS%3D{proj}%26WIDTH%3D{width}%26HEIGHT%3D{height}%26BBOX%3D{bbox}&disable_features=boundaries&map=15.95/52.4708/13.4178

I think this needs a background to work.

sources/world/GeoScribbles.geojson
Zverik: I would not tie this to a single editor. Pretty sure we'll see this layer being populated from other apps before the year ends.

sources/world/GeoScribbles.geojson
Zverik: Exactly: https://github.com/openstreetmap/iD/issues/9280

It does need background to work. Just like OSM GPS layer. That's why there's `"overlay": true` key included.

Zverik:
> I suggest making it a bit more explicit by adding a section to the wiki and changing the link to https://wiki.openstreetmap.org/wiki/GeoScribble#licence

Good idea, just did that.

> This is the age of the data that will show up? I think 3 month should be better since there will not be a lot of data to start with and moving from mobile to a different editor can take a bit of time…

In my experience, I either map surveyed things in a week, or forget about them altogether. There are three options:

1. Keep it 3 weeks.
2. Change to 3 months.
3. Add two GeoScribble layers, one for weeks and one for months (that's the matter of `layer=` parameter).

> > * `default=true`?
> > * `i18n=true`?
> 
> What are those?

Questions on whether:

* Include the layer in the defaults, like OSM GPS layer and Bing & Mapbox imagery.
* Allow for translating the layer name and description.

sources/world/GeoScribbles.geojson
tordans: Agreed, that might very well be. But it will be easy to update it here, once that happens. Right now, I don't see a way that a user would be able to understand where those GeoScribbles came from just by looking at the title and description in the iD Editor Sidebar. — Adding a bit more context helps a lot.

sources/world/GeoScribbles.geojson
tordans: Ah, I get it now, thanks. I never really looked at the overlay-prop and feature in iD but looking at the code now this should show up in the overlay section. Nice.

tordans (approved):

matheusgomesms:
I believe _i18n=true_ could be used, since the description would then be available to be translated ("what is GeoScribble?" some users might ask at first), and also given that the proposed name is "GeoScribble latest notes", this _latest notes_ part can and should be translated as well to non-English speakers.

sources/world/GeoScribbles.geojson
```diff
@@ -0,0 +1,29 @@
+{
+    "geometry": null,
+    "properties": {
+        "attribution": {
+            "required": false,
+            "text": "GeoScribble contributors",
+            "url": "https://wiki.openstreetmap.org/wiki/GeoScribble"
+        },
+        "description": "Geometric survey notes from Every Door editor, aged three weeks at most",
+        "i18n": false,
```
matheusgomesms:
```suggestion
        "i18n": true,
```
Per my last comment.

PR #2314, opened 2024-05-08
Add Massachusetts MassGIS 2023 aerial imagery

petercooperjr:
(And removing 2019, since the standard seems to be to keep the most recent two sets of imagery.)


matheusgomesms:
Thanks!

PR #2317, opened 2024-05-10
SaxonyAnhalt_ATKIS_BU: Category "other"

tordans:
https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md?plain=1#L80-L87 defines "map" as "A map which is not mostly based on OSM data." which is not the case for ALKIS Data. Other Alkis layer where mapped as "other".


matheusgomesms:
Checks are annoying because of lack of privacy policy. Could you please add this, if correct?

"privacy_policy_url": "https://geodatenportal.sachsen-anhalt.de/gfds/de/datenschutz-service.html"

tordans:
@matheusgomesms I added the suggest privacy policy. This is a bit out of scope for this PR, but let's do this right here to make the test happy…
I checked and was not able to find a better URL that covers the domain that we use, so it looks like a good fit.

Quick research notes: The wiki has nothing specific; the [general domain](https://www.sachsen.de/datenschutz.html) has nothing more specific; the [JOSM index](https://josm.openstreetmap.de/wiki/Maps/Germany#Saxonylatestaerialimagery) does not include a link either.

matheusgomesms:
Thanks, @tordans!

PR #2329, opened 2024-05-22
Added Albanian Adress System as background Imagary for Albania

LegendFabix:
Added the official Albanian Adress System from ASIG as Backgound Imagary.


LegendFabix:
Could somebody just pull this into the base branch? Everything is fine, and i don't know why are you waiting guys...

PR #2330, opened 2024-05-23
Update NRW_ortho_wms.geojson

mrhr95:
changed description to explain origin and accuracy; changed name to accommodate for additional layers (colorIR and nIR)


matheusgomesms:
Danke!

PR #2331, opened 2024-05-23
Create NRW_alkis_buildings_GST.geojson

mrhr95:
Additional overlay for "NRW_alkis_buildings" to check accuracy of points in "NRW_alkis_buildings"


iandees:
Based on your description, is this a duplicate of an existing layer? Should we just change existing layer to be an overlay?

mrhr95:
Hey iandees,

No, this ist not a duplicate, but additional information to an existing layer. The existing layer gives lines of buildings, my added Layer gives only colored points, no lines. Colors depend on accuracy of points.

iandees:
Gotcha.

PR #2336, opened 2024-05-26
Update urls for San Francisco imagery

arch0345:


andrewharvey:
Thanks, I was unable to test this as the tile url is not responding for me, perhaps it's applying geo restrictions or down at the moment.

arch0345:
Yeah, I think it might be georestricted to just the US, it's working fine for me:
image

andrewharvey:
Maybe we need a new flag in the schema to inform users access may be geoblocked.

arch0345:
I was about to mention https://ags-proxy.openstreetmap.us to get around geoblocking but it seems like this only works for ArcGIS MapServer/ImageServers

PR #2341, opened 2024-05-28
Update ACTmapi sources

### LINKED ISSUE #2340: au/act urls will need to be updated due to server shutdown 31 May 2024.
The ACT Government is shutting down their server on 31 May 2024 and will be moving to an ArcGIS hosted server.

I have updated the JOSM map list to match this: https://josm.openstreetmap.de/wiki/Maps/Australia

https://josm.openstreetmap.de/wiki/Maps/Australia#ACTmapiImagery2024-02 is available in EPSG:3857, so should work in iD. The other three layers are only available in EPSG:7855, so only usable in JOSM I think.

Thanks.

arch0345:
Closes #2340


sources/oceania/au/act/ACTmapi-Imagery202402.geojson
```diff
@@ -0,0 +1,195 @@
...
+        "url": "https://tiles.arcgis.com/tiles/E5n4f1VY84i0xSjy/arcgis/rest/services/ACT_Aerial_Imagery_202402_AGOL_Compatible/MapServer/tile/{zoom}/{y}/{x}",
+        "best": true,
+        "start_date": "2024-02",
+        "end_date": "2024-02",
+        "max_zoom": 21,
+        "country_code": "AU",
+        "type": "tms",
+        "icon": "https://actmapi.act.gov.au/img/apple-touch-icon.png",
+        "category": "photo",
+        "min_zoom": 5
```
andrewharvey: It looks like the data only starts appearing from zoom 10, so should we set this to 10 instead?

andrewharvey (changes requested):
Thanks for the PR. I've added one comment.

I think leaving the WMTS only endpoints is okay, I assume that means only JOSM would support them?

arch0345:
> I think leaving the WMTS only endpoints is okay, I assume that means only JOSM would support them?

Yeah that's correct

andrewharvey:
Thanks looks good. I'll try to keep on eye on the older imagery as it would be nice to retain that in the index, and also for OHM's ELI.

PR #2345, opened 2024-06-08
Add OrtoSat 30cm 2023 imagery (Portugal)

mikedld:
Open questions:
* `id`, `name`, and filename — used my best judgement
* `best` — didn't set to `true` but  maybe I should?
* `max_zoom`, `min_zoom` — not sure if those are correct and where to get them from

Feel free to amend yourself, or leave a comment and I'll amend it.


matheusgomesms:
I took the liberty of:

- adding best=true (this is the most recent available imagery)
- changed a bit ID and name (to reflect it is available only in the central part of the country)

Looks great to me, thanks a lot! Let's wait a bit for the Portuguese community agree or not with these changes and then I'll merge this PR. Obrigado!

PR #2347, opened 2024-06-13
add several files for Japan region

nyampire:
Just added several files for Japan region.

* Inuyama city
* Kouchi city
* Nagahama city

Those files do not have the "attribution": {"url": } parameter. However, that is correct.
We (OSMF Japan) obtained the original files through an explicit request to the municipalities, as permitted under the Japanese Survey Act.

I rewind those files to "right-handed" with [mapster lint](https://mapstertech.github.io/mapster-right-hand-rule-fixer/), I hope it can pass the test.


nyampire:
The server does not have `privacy policy` URL.

So I could not describe it.

PR #2368, opened 2024-06-24
Add Mesa County GIS 2024 imagery

acarasimon96:
This high-resolution aerial imagery covers only cities, towns, and rural communities in [Mesa County, Colorado, USA](https://en.wikipedia.org/wiki/Mesa_County,_Colorado).


acarasimon96:
I noticed that this PR has been stuck at "1 workflow awaiting approval" since I submitted it. Any update on getting this looked at?

sources/north-america/us/co/MesaCountyValleywideImagery2024.geojson
```diff
@@ -0,0 +1,3229 @@
...
+            "url": "https://gis.mesacounty.us/",
+            "text": "Mesa County GIS",
+            "required": true
+        },
+        "icon": "https://gis.mesacounty.us/globalassets/images/county-logos/mesa-county-logo.png",
+        "category": "photo",
+        "min_zoom": 5
+    },
+    "geometry": {
+        "type": "Polygon",
```
arch0345:
```suggestion
        "type": "MultiPolygon",
```
It looks like this geojson is made up of multiple outer rings, so the type `MultiPolygon` should be used. Here's the formatting for geojson multipolygons (basically an array of polygons): https://docs.mapbox.com/android/java/api/libjava-geojson/5.8.0/com/mapbox/geojson/MultiPolygon.html

arch0345 (changes requested):

sources/north-america/us/co/MesaCountyValleywideImagery2024.geojson
acarasimon96: Just got this done. I also reordered the coordinates to follow the right-hand rule to satisfy the warning about it on geojson.io. I am a novice in dealing with GeoJSON, so further review is appreciated.

arch0345 (approved):

Issue #2332, opened 2024-05-24
Update OS OpenMap Local (Great Britain)


nc011:
OS OpenMap Local (April 2023) is already a layer. The data source for this layer is updated every 6 months (October and April), meaning there is now an April 2024 layer available.

## URL of the imagery
https://www.ordnancesurvey.co.uk/products/os-open-map-local

## Two letter country code
GB

## Desired layer name
OS OpenMap Local (April 2024)

## License
Same as current (2023) layer
[OGL 3.0](https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/)

## Is the license compatible with OSM?
Same as current (2023) layer
Yes

## Attribution (text, URL, required?)
Same as current (2023) layer. Presumably: "Contains public sector information licensed under the Open Government Licence v3.0."

## Privacy policy (URL)
Same as current (2023) layer

## Icon
Same as current (2023) layer

## Bounding polygon
Same as current (2023) layer

PR #2372, opened 2024-06-25
Update OS OpenMap Local (Great Britain)

Jez-C:
Layer already added by https://github.com/openstreetmap/chef/commit/57d7cd3ecda47266cc08d7e97b13e606405eb2a8 so this simply updates the index, resolving https://github.com/osmlab/editor-layer-index/issues/2332

### DIFF
--- a/sources/europe/gb/OSOpenMapLocal.geojson
+++ b/sources/europe/gb/OSOpenMapLocal.geojson
@@ -1,12 +1,12 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "OS-OpenMap_Local_2023_04",
-        "name": "OS OpenMap Local (April 2023)",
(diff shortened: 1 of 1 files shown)


Intrinsica1:
Is the 2023 layer still operational? if so, it might be useful to keep it for reference.

Jez-C:
@Intrinsica1 I agree that it is occasionally useful to refer back to layers from previous years. That's why all (except one) of the OS OpenMap Local releases going back to October 2016 have been kept at https://os.openstreetmap.org/ 
Because they're all archived there, I don't think it's necessary to have more than just the most up-to-date OS OpenMap Local layer in the editor index.

UKChris-osm:
I came here to report that the current layer is also incorrect, as it refers to the April 2023 layer, but is using the October 2022 URL, but I see this PR will update the OS GB layer to the most recent map available :)

Issue #2337, opened 2024-05-27
Default imagery doesn't load in JOSM


elevenchars:
When I set the JOSM preference described in the README and refresh my default imagery sources, nothing shows up
```
imagery.layers.sites	[https://osmlab.github.io/editor-layer-index/imagery.xml]
```
I wasn't able to see any errors. The default JOSM imagery.layers.sites value works.

ZeLonewolf:
Please re-open.  The README should not include steps which do not work.

cicku:
@ZeLonewolf How did you know that?

ZeLonewolf:
I attempted these steps and also found that they do not work in JOSM. 

That means that the README either needs to change to have the correct steps, or else be updated to reflect that ELI doesn't work in JOSM.

cicku:
I don't think so?

![image](https://github.com/osmlab/editor-layer-index/assets/686438/8307e3a8-fdf8-44be-a258-0801caedb3c1)
![image](https://github.com/osmlab/editor-layer-index/assets/686438/18522824-6435-4972-9229-9b59434f9afe)

ZeLonewolf:
I tested on JOSM build 19128 on Ubuntu, added the ELI URL to that key in preferences, re-loaded in the imagery list, and got an empty list as a result:

![image](https://github.com/osmlab/editor-layer-index/assets/3254090/f7485cda-d666-4c75-8369-f528b5cbb0c7)
![image](https://github.com/osmlab/editor-layer-index/assets/3254090/8fd45a4f-3b6a-482d-b517-206591310f58)
![image](https://github.com/osmlab/editor-layer-index/assets/3254090/6f752455-18ef-4b66-9896-9c8708955adf)

ZeLonewolf:
I also see in the logs:
```
2024-07-06 22:43:57.755 INFO: GET https://osmlab.github.io/editor-layer-index/imagery.xml -> HTTP/1.1 200 (421 ms; 1.16 MB)
```
So it seems that JOSM pulls down the file successfully and then maybe has a bug later on. Perhaps a bug in the latest version of JOSM.

cicku:
Thanks, but you really need to check with JOSM. I won't troubleshoot further when I cannot reproduce.

This issue sounds like a JOSM bug since day 1 and that's why I marked it as invalid, and I will insist on the decision to close it.

iandees:
[...]

PR #2373, opened 2024-06-27
Fix PEMA_Orthoimagery_2022

Intrinsica1:
Synchronizes PEMA_Orthoimagery_2022 with JOSM Imagery sources to fix its URL and bounds.


Intrinsica1:
Note that I used `scripts/convert_individual.py`'s output as reference to modify the original GeoJSON.

sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson
```diff
@@ -3,14 +3,14 @@
...
...YgLTEuMzI0LDYuNTA2IGggLTEuNTY4IGwgMS43ODcsLTguNzU5IGggMS45NTEgbCAxLjE4MSw0LjI0IDIuODc5LC00LjI0IGggMi4xNjkgeiIvPjxwYXRoIGQ9Im0gNDEuODA2LDM4LjEyOSAtMC4yNTMsLTIuMDM3IGggLTMuMyBsIC0xLjA2LDIuMDM3IGggLTEuNyBsIDQuNjQyLC04Ljc1OSBoIDIuMTY3IGwgMS4xNDUsOC43NTkgeiBtIC0wLjg1NiwtNy4yMTYgLTIuMDM3LDMuODU1IGggMi40ODQgeiIvPjwvZz48ZyBmaWxsPSIjZmZmZmZmIj48cG9seWdvbiBwb2ludHM9IjM4LjM1MywyMC44NzEgMzYuMzg3LDIzLjAzMSAzOC4yOCwyNC43NTggMzUuMTA3LDI0LjQxIDMzLjA5MiwyNi42MjQgMzMuMTE1LDI0LjE5MSAyOS45MDIsMjMuODM3IDMzLjE0NiwyMi43MTggMzMuMjE0LDIwLjMwNiAzNS4xNDUsMjEuOTg1Ii8+PHBvbHlnb24gcG9pbnRzPSIyNy4yMzEsMjEuNzExIDI3LjEzMywyMy41NjkgMjkuNTM5LDIzLjgzMiAyNy4xMTksMjQuNzM4IDI2Ljk3NSwyNi41MjIgMjUuNTQzLDI1LjI1MiAyMy4xMjMsMjYuMDAxIDI0LjYwNCwyNC4zNzIgMjMuMjQ2LDIzLjA2MyAyNS42NDEsMjMuMzQ2Ii8+PHBvbHlnb24gcG9pbnRzPSIyMi44MDMsMjYuMTMxIDIxLjQ2NSwyNy40MDEgMjIuMzQsMjguNTI3IDIwLjU2LDI4LjE1IDE5LjE2MywyOS4zNzQgMTkuNDAxLDI3LjkyOCAxNy42MTksMjcuNTI0IDE5LjUxOSwyNy4wMzEgMTkuODI5LDI1LjU0OSAyMC44MDUsMjYuNjMxIi8+PHBhdGggZD0ibSAzOC44NDksMjQuOTA2IGEgOC45MjQsOC45MjQgMCAwIDEgNC45NjgsMy4zNjggYyAxLjQzNCwyLjIyMiAtMC4xLDUuMDcyIC0wLjEsNS4wNzIgYSA2LjU3OCw2LjU3OCAwIDAgMCAtMS4zMTksLTUuNjI5IDkuNzMyLDkuNzMyIDAgMCAwIC0zLjU0OSwtMi44MTEgeiIvPjwvZz48L3N2Zz4=",
```
1ec5: Normally we check images directly into the repository as ordinary files. It looks like scripts/unembed-icons.py will take care of this for you.

sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson
```diff
@@ -3,14 +3,14 @@
...
-            "text": "Pennsylvania Emergency Management Agency",
+            "url": "https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=5177",
+            "text": "PASDA",
             "required": false
         },
-        "name": "PEMA Orthoimagery - Western (2022)",
-        "icon": "https://www.pema.pa.gov/PublishingImages/logo.svg",
-        "url": "https://imagery.pasda.psu.edu/arcgis/services/pasda/PEMAImagery2022/MapServer/WmsServer?request=getmap&version=1.3.0&service=wms&layers=2&styles=&format=image/jpeg&crs={proj}&width={width}&height={height}&bbox={bbox}",
-        "max_zoom": 18,
+        "name": "PEMA Orthoimagery [2022]",
```
1ec5: I think the previous name would be more accurate. Consistent with this layer’s official title, “PEMA Orthoimagery Western PA”, it covers only a slice of western Pennsylvania (including Pittsburgh). Perhaps the JOSM index could use that name instead?

For reference, a similarly named layer covers the whole state:

https://github.com/osmlab/editor-layer-index/blob/4388011bbba74926d3622a6c6355a05e73ad8e00/sources/north-america/us/pa/PEMA_Orthoimagery.geojson#L10

sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson
```diff
@@ -3,14 +3,14 @@
     "properties": {
         "id": "PEMA_Orthoimagery_2022",
         "attribution": {
-            "url": "https://www.pema.pa.gov/911-Program/Partners/NG911-GIS/Aerial-Imagery/Pages/default.aspx",
-            "text": "Pennsylvania Emergency Management Agency",
+            "url": "https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=5177",
+            "text": "PASDA",
```
1ec5: PEMA is the author and PASDA is the host, right? The previous text would be more accurate, although it doesn’t matter that much since attribution isn’t required.

sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson
Intrinsica1: the script unfortunately didn't seem to work/do anything

Intrinsica1:
really appreciate the reviews, thanks :)

Intrinsica1:
Just to inform that I believe it is ready to merge. Also to reiterate that this layer is currently in a nonfunctional state in the master branch.

1ec5:
The build is failing because the icon hasn’t been deployed yet, but obviously that’s not your fault:

> ERROR:root:sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson: icon url https://osmlab.github.io/editor-layer-index/sources/north-america/us/pa/PEMA.svg is not reachable: HTTP code: 404

I see that most other sources just link to the URL directly. We could link to https://www.pema.pa.gov/PublishingImages/logo.svg , but I think what you have should be fine once we merge.

Intrinsica1:
The reason I didn't link the original SVG is that I modified it to render properly in JOSM (both still visually the same). Though this is the fault of JOSM, and not the SVG.

1ec5:
Makes sense. I also noticed that, for a couple layers, someone had also converted the SVG to PNG and checked in the PNG, maybe for the same reason.

cicku:
Any reason you need to upload the icon to the repo, also @1ec5 did you notify me for any PEMA changes?

1ec5:
> Any reason you need to upload the icon to the repo

@Intrinsica1 said above that the SVG from the site didn’t render well when ELI is loaded in JOSM, so they checked in a simplified version. The JOSM rendering issue is not unexpected, since every SVG renderer has its own quirks; another ELI consumer could well have the same issue.

If the checked-in icon is an issue, we could just omit the icon; it’s nothing more than a nicety.

cicku:
Although I personally wish JOSM team the best on their bugfix journey, this issue along with another one #2337 requires me to emphasize again that **ELI is not the correct place to fix/discuss whatever problem that JOSM possesses.**

There will always be bugs/quirks in JOSM since it is technically a software, and no software is perfect. JOSM can have all kinds of problems but it is not our objective to find a workaround, we are only a layer index providing the information as-is, and our consumers do not just include JOSM.

1ec5:
I get your point and agree in general, but SVG is a very fragile format. If JOSM is having problems with this particular SVG, it’s entirely possible that iD would when viewed in some less popular browser. But it’s hard to say without more information. If @Intrinsica1 could provide a link to the bug report in JOSM’s Trac instance, we could try to work toward a resolution there. Or we could shrug and delete the icon from this repository. It’s quite a trivial matter – I don’t even think this icon would be all that legible in the corner of iD against the average PEMA imagery.

simonpoole:
May I point out that this is supposed to an editor neutral index of imagery. 

While PNG and JPEG are nearly universally supported on computing devices, the same can not be said of SVG, for example Android has no native SVG support, which just effects the small number of 4 billion devices. There are 3rd party libraries to provide such support, but, seriously, I don't think it is a good policy to expect SVG support to be added just for a couple of manufacturer icons.

tl;dr keep the icons in the repo to PNG and JPEG.

cicku:
> May I point out that this is supposed to an editor neutral index of imagery.

Of course I agree.

> While PNG and JPEG are nearly universally supported on computing devices, the same can not be said of SVG, for example Android has no native SVG support, which just effects the small number of 4 billion devices. There are 3rd party libraries to provide such support, but, seriously, I don't think it is a good policy to expect SVG support to be added just for a couple of manufacturer icons.

Then the CONTRIBUTING.md needs to be updated.

PR #2377, opened 2024-06-27
Add CAPCOG 2022 imagery

arch0345:


cicku:
You can put this as the license: https://api.tnris.org/api/v1/collections/a15f67db-9535-464e-9058-f447325b6251

PR #2380, opened 2024-07-01
Cleanup all PEMA Orthoimagery

Intrinsica1:
Cleans up the PEMA Orthoimagery to have consistent (both with each other and the JOSM Imagery sources) URLs, names, descriptions, icons, zoom, and projections. It also corrects the bounds to correspond to the imagery bounds, not PA state bounds.


Intrinsica1:
NOTE: PEMA_Orthoimagery_2021's new bounds appear to be incorrect, but you'll find that if you zoom into the missing area, imagery there does not load.

Intrinsica1:
~TODO: remove trailing zeros from 2018_2020's geometry~

cicku:
I maintain MD and PA layers, I do not accept this changeset:

1. PASDA will update layers once the new flight and processing finish.
2. This project has nothing to do with JOSM sources.

Intrinsica1:
I should have said this in my original comment, but all the changes to these layers in the JOSM Imagery sources were made by myself with the intention of contributing them here as well. This isn't just a random attempt to make them consistent, it's just a bonus that I thought to mention.

I will admit that the description should include the official list of states that are part of the imagery, instead of just what is accessible. But I don't think 2021's bounds should include imagery that is currently inaccessible - at least, it shouldn't include imagery that will never be added (as it stands now, the bounds include the whole state). I can make those changes to the pull if you would prefer.

It's an aside, but this project doesn't have "nothing" to do with the JOSM Imagery sources, as this project was descended from it.

cicku:
I need to be crystal clear:

1. ` This isn't just a random attempt`: Anybody can edit JOSM index and it is not an official source of truth. There is literally no review in that project, and we have found many non-free sources were added to that project in the past. There is also no reason to sync two projects **just because you wish or you think so**. As long as layers are working, it does not matter if they should be the same or not.
2. `don't think 2021's bounds should include imagery that is currently inaccessible`: What you think may not be what is convenient to the public/maintenance. The initial plan by PASDA is to merge all these into a single cached layer, but due to some flight delays as well as data processing issues, this has not happened for year 2022+. I would expect to only have 1 layer but not multiple, but I have no power to control that which is why we separate out.
3. `as this project was descended from it`: Now you sound condescending. Just because project A has influence on project B does not mean project B needs to pander.
---

As long as I'm active in this project, I do not accept random change requests to PA or MD layers unless the URL stops working. Your another PR introduced unwanted data to this repo, and I'm kinda surprised that the reviewer even gave you the pass.

Just FYI: Your previous changes are subject to revert as well.

Issue #2101, opened 2023-10-08
Channel Islands


SK53:
Generic "France" polygons, e.g., BDOrthoIGN include the Channel islands, but there is no imagery for these areas, and the default imagery layer always needs to be changed. The Bailiwick of Guernsey (iso code gg) and Jersey (je) need to be excluded from these polygons, probably by adding another point off St Malo.

osmuser63783:
Duplicate of #1260 which has a bit more info!

Issue #1260, opened 2021-09-09
Request extent-change to exclude unintended nearby islands


Lee-Carre:
For the following (background) imagery

* [BDOrtho IGN](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/BDOrthoIGN.geojson)
* [Ortho HR](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/OrthoHR.geojson)

please alter the extent to **exclude** the Channel Islands (Jersey, Guernsey, etc.).

The imagery is clearly intended for France (and the islands haven't been French territory for several hundred years). While there is partial low-resolution imagery for the islands, it's not usable for any productive work, and is a distraction in the list when needing imagery of specific features (since it often renders blank when highly zoomed in).

Unless, of course, the source needs to be updated.

Try it; zoom in to (Jersey at) a level as if you were going to add some building outlines, then select either of those imagery sources.

Lee-Carre:
Hmm, I notice that the GeoJSON for Ortho HR doesn't include the CIs, yet when highly zoomed in (to said Islands) using Vespucci the imagery is still listed as applicable.

Is that then a bug for Vespucci, instead?

simonpoole:
> 
> 
> Hmm, I notice that the GeoJSON for Ortho HR doesn't include the CIs, yet when highly zoomed in (to said Islands) using Vespucci the imagery is still listed as applicable.
> 
> Is that then a bug for Vespucci, instead?

No, it is the intended behaviour. 

We don't retain the polygons (aka we throw them away) and just store the (multiple if a MP) bounding boxes instead. The reasons for this are on the one hand the amount of heap it would cost to keep them in memory and on the other hand the lack of geospatial indices in Androids sqlite implementation (making a DB based store clumsy and slow). If Androids sqlite is one day updated to include the support, then we might reconsider.

Lee-Carre:
Having done some further checking, the following (overlay) sources need correction, too:

* [BANO](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/BANO.geojson)
* [BD Carthage](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/BDCarthage.geojson)
* [BD Topo](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/BD-Topo.geojson)
[...]

PR #2383, opened 2024-07-01
Update BDOrthoIGN.geojson to exclude Channel Islands

osmuser63783:
This is my first time contributing here, I hope what I am doing is right.

The issue I am trying to fix is that the iD editor defaults to the BDOrtho IGN layer when editing in the Channel Islands, but this layer doesn't have any data in the Channel Islands, only in France. This has been reported before in #1260 and #2101.
Screenshot 2024-07-01 at 22 11 06

I opened https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/BDOrthoIGN.geojson in QGIS and subtracted the Channel Islands admin boundaries, which I got from OSM. The result is a polygon with 'holes' for the Channel Islands: https://github.com/osmuser63783/editor-layer-index/blob/db397a513e3fdf600ac92f47f4c8c88d6a4ab8e3/sources/europe/fr/BDOrthoIGN.geojson

Screenshot 2024-07-01 at 22 21 12

I haven't tested this on a test server with iD or anything like that. Please let me know if this should fix it!


Intrinsica1:
This adds a lot of unnecessarily precise geometry to roughly drawn bounds. I think the best thing here is to simply retrace the existing geometry. Or ideally obtain the actual bounds from IGN!

~I have already started retracing a bit.~

Decided to look into getting bounds directly instead.

I've found `ORTHOIMAGERY.ORTHOPHOTOS.GRAPHE-MOSAIQUAGE:graphe_bdortho` at `https://data.geopf.fr/wfs/ows?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetCapabilities`

The issue is that it combines good bounds, with overly-complex internal geometry:
qgis-bin_nLxRxYRjAI

Dissolving it has been problematic.

arch0345:
I've simplified the geometry, inner rings aren't really needed to exclude these islands
image

osmuser63783:
Thanks a lot!

PR #2389, opened 2024-07-08
Add category value to xml output and more JOSM compatibilities fixes

### LINKED ISSUE #2388: ELI does not load in latest JOSM
Follow-on from #2337 which is locked

This issue is [confirmed](https://josm.openstreetmap.de/ticket/23785#comment:1) by the JOSM team who state:

> The problem is that we require an imagery category in the XSD. I don't know the reasoning for that off the top of my head.
The reason why things changed recently is we do more validation of requirements in the XSD that cannot be done via XSD due to technical limitations. 
> See [source:trunk/resources/data/maps.xsd@19130:638-640,686-688#L638](https://josm.openstreetmap.de/browser/trunk/resources/data/maps.xsd?marks=638-640%2C686-688&rev=19130#L638)[​](https://josm.openstreetmap.de/export/19130/josm/trunk/resources/data/maps.xsd#L638) .
> Anyway, the "easy" solution would be to add an arbitrary category to each entry. It does look like the ELI has a category property that can be used (and the JOSM category might be based off of that, I don't know).

Note this issue occurs on the _latest_ version of JOSM, which I confirmed on build 19128.  The JOSM team confirms this is a recent change.

From this description it appears that ELI may no longer be compatible with JOSM.

simonpoole:
This needs some testing, but likely does the trick.

Fixes https://github.com/osmlab/editor-layer-index/issues/2388

### DIFF
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -4,7 +4,7 @@ import io
 import xml.etree.cElementTree as ET
 from shapely.geometry import shape, Polygon, MultiPolygon
 
-root = ET.Element("imagery")
+root = ET.Element("imagery", {"xmlns" :"http://josm.openstreetmap.de/maps-1.0"})
 
 sources = []
 for file in sys.argv[1:]:
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -20,8 +20,8 @@ def add_source(source):
     name = ET.SubElement(entry, "name")
     name.text = props["name"]
 
-    name = ET.SubElement(entry, "id")
-    name.text = props["id"]
+    id = ET.SubElement(entry, "id")
+    id.text = props["id"]
 
     type = ET.SubElement(entry, "type")
     type.text = props["type"]
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -84,6 +84,7 @@ def add_source(source):
     if "description" in props:
         description = ET.SubElement(entry, "description")
         description.text = props["description"]
+        description.set("lang", "en")
 
     if "min_zoom" in props:
         min_zoom = ET.SubElement(entry, "min-zoom")
--- a/scripts/convert_xml.py
+++ b/scripts/convert_xml.py
@@ -3,7 +3,7 @@ import sys
 import io
 import xml.etree.cElementTree as ET
 from shapely.geometry import shape, Polygon, MultiPolygon
-from shapely import get_num_geometries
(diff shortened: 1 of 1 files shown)


simonpoole:
So after running the output through xmlstarlet, seems as if there are five remaining issues:

- [x] add xlmns attribute
- [x] lang attribute missing in description elements 
- [x] ELI uses a country code of "ZZ" in one entry and "XN" in another that JOSM doesn't include as valid
- [x] JOSM restricts max-zoom to 24 and ELI has one instance of 31 being used (which is naturally nonsense)
- [x] there are a handful of complaints about a point element turning up unexpectedly (needs investigation)

simonpoole:
>     * [ ]  there are a handful of complaints about a point element turning up unexpectedly (needs investigation)

This is due to JOSM limiting the number of points in a "shape" element to 999, potential solution: simply use a bounding box for these entries (8 as of right now), as we already have the bounding box, this means simply dropping the polygon in these cases.

simonpoole:
For @grischard the following seem to have polygons with more than 999 vertices

Using BBOX for osmim-imagicode-usgs_250k_ant has 1222 vertices
Using BBOX for gsi.go.jp_airphoto_2020 has 1255 vertices
Using BBOX for bev-inspire-orthofoto has 1323 vertices
Using BBOX for South-Tyrol-DSM_2013 has 1013 vertices
Using BBOX for South-Tyrol-DTM_2013 has 1013 vertices
Using BBOX for USDA-NAIP has 1001 vertices
Using BBOX for san_antonio_river_2016_wms has 5509 vertices
Using BBOX for BDGEx_ctm_multi has 1001 vertices
Using BBOX for SP_Itu_-_Eixos has 2716 vertices

tsmock:
> [...] the following seem to have polygons with more than 999 vertices
> [...]
> Using bounding box for SPW2021

There should be 121 vertices for SPW2021 (the only one I spot-checked).

simonpoole:
> > [...] the following seem to have polygons with more than 999 vertices
> > [...]
> > Using bounding box for SPW2021
> 
> There should be 121 vertices for SPW2021 (the only one I spot-checked).

Sorry, that was a last minute rush before I went on vacation, I've corrected the list now.

tsmock:
I hope you had a good vacation. :)

Anyway, it looks like some of those high-vertices count polygons can/should be simplified. But probably not in this PR.
For example, `san_antonio_river_2016_wms` (`5509`) loses `3504` vertices with `0.5`m max error simplification and `5091` with `1`m max error.

PR #2396, opened 2024-07-16
Add Arapahoe County Aerials Latest

arch0345:
I got permission from Dominick Cisson from Arapahoe County GIS to use this imagery for OSM


iandees:
Can you re-upload the permission doc?

arch0345:
The email conversation should be here: https://github.com/osmlab/editor-layer-index/blob/df94a043455d889d2ffaad27244f67223c999c8a/sources/north-america/us/co/Arapahoe.pdf
Check is failing as the url won't work until this is deployed to github pages on merge

PR #2397, opened 2024-07-16
Add Philadelphia Ortho 2023

arch0345:


cicku:
@iandees @arch0345 this is another one without my review but pushed into the production.

This layer should not cover anywhere in NJ because it sets the best at the wrong place.

image

iandees:
Thanks for the report. Please file a PR or an issue to resolve the problem you're pointing out here.

PR #2408, opened 2024-07-25
Add 19th C. Ireland maps

amandasaurus:
Port the 19th C. Ireland OS 1st & 2nd Ed. 6inch:mile maps from the JOSM Imagery Wiki with the convert_individual.py script. And some manual file renaming.

Thanks to National Library of Scotland for hosting the maps.


andrewharvey:
Are these all out of copyright? Should then the attribution required flag be false?

andrewharvey:
The CI error

> ERROR:root:No zoom level reachable. (Tested: 6,7,8,9,10,11,12,13,14,15,16,17) Tiles might not be present at tested location: -6.100466527920521,54.83941

is probably just due to limitations of the check.

I think if we can confirm the copyright situation this should be fine to merge.

cicku:
Today is 2025/03/22, I don't find any legitimate reason to include these from 18xx.

amandasaurus:
> Are these all out of copyright?

yes.

> Should then the attribution required flag be false?

I think it's nice to include an attribution to the insitution which is making them available. (Nat. Lib. Scot.) 🙂

amandasaurus:
> Today is 2025/03/22, I don't find any legitimate reason to include these from 18xx.

I am an active and long time mapper in Ireland, and these maps are still very useful today.

danieldegroot2:
Related https://github.com/openstreetmap/iD/pull/12878 .

andrewharvey (approved):
I approve.

>  think it's nice to include an attribution to the institution which is making them available. (Nat. Lib. Scot.) 🙂

I'm not really sure why we actually need the `attribution: true/false` flag, surely as a project we want to provide attribution regardless of if that is legally required or not.

PR #2415, opened 2024-07-26
ie: replace the geometry from one imagery with another (fixes #2176)

### LINKED ISSUE #2176: Ireland British War Office 1:25k GSGS 3906 bounding polygon is too rough
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ie/IrelandBritishWarOffice-GSGS3906.geojson

it is very spikey and is cutting across parts of land causing the layer to disappear from the imagery list at such locations, which can be annoying when using multiple layers for verification and constantly having to drag the map back and forth to retrieve the layer.

I would like to suggest changing the geometry to be more similar to that of other layers where it also includes chunks off the coast.
( https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/ie/IrelandBritishWarOfficeOne-Inch1941-43GSGS4136.geojson )

amandasaurus:


matheusgomesms:
Thanks, Amanda!

Do you think we could add a license_url and a privacy_policy_url in this layer?

grischard:
@amandasaurus would the polygon from http://download.geofabrik.de/europe/ireland-and-northern-ireland.html work?

amandasaurus:
I've rebased this PR, and added the `privacy_policy_url`. However, I'm unsure for the `license_url`. The original maps are out of copyright (being from the 1940s). I don't know any more.

andrewharvey:
I think it would be good then if the `license_url` could point to any metadata or source page for the original maps, ideally which then shows some information we can use to verify/determine they are out of copyright.

danieldegroot2:
@andrewharvey Individual layers hosted digitally by National Library of Scotland can be found at
https://maps.nls.uk/os/25k-gb-1940-43/
which lists their publication dates (and theoretically they could fall under the same as #2408 )

It does not appear to be queryable digitally through [TCD Map Library](https://www.tcd.ie/library/map-library/) ('Search Collections' at top). (only other website which indexes these maps is British Library but only short metadata)

PR #2419, opened 2024-07-27
Add FNSB Aerial 2023

arch0345:
I confirmed that aerial imagery from FNSB is considered to be in the public domain


iandees:
The license URL check probably should be a warning. Or maybe it should check inside the PR for the associated document to exist, assuming the root of the URL is the root of this repo.

PR #2434, opened 2024-07-29
Add Lorain Ortho 2024

arch0345:
I confirmed with the Lorain County Auditor's Office that OSM contributors would be allowed to use their imagery


iandees:
I adjusted the max_zoom to 21, because I agree with the CI and couldn't get it to give me any tiles at zoom 22+, even though the layer metadata says they exist.

PR #2436, opened 2024-08-01
[Help needed] Add imagery for DE/Ludwigshafen

coolultra1:
More information on the permission:

https://ludwigshafen.de/buergerservice/dienstleistungen/geodatendienste-wms

I've got their kind-of stretchy wording there confirmed to fit to OSM via an Email conversation with the city administration, so everything should be all right here.


coolultra1:
Anybody know how I would fix the GeoJSON? I'm actually clueless.

arch0345:
> Anybody know how I would fix the GeoJSON? I'm actually clueless.

https://github.com/coolultra1/editor-layer-index-1/pull/1 should fix the geometry

coolultra1:
@arch0345 Thank you very much; What was the trick? How'd you do it so easily?

arch0345:
I just copy pasted the suggested geometry from the CI error message and reformatted it in VSCode

PR #2453, opened 2024-08-18
Adding two new hillshade layers for Slovakia

arch0345:
Supersedes #1175


arch0345:
Looks like this was added back in #1996

PR #2455, opened 2024-08-21
Update Fairfax county imagery URL

wcedmisten:
Fairfax can't seem to find a stable URL

https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/

### DIFF
--- a/sources/north-america/us/va/Fairfax_VA_2024.geojson
+++ b/sources/north-america/us/va/Fairfax_VA_2024.geojson
@@ -1,13 +1,13 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "Fairfax_VA_2023",
+        "id": "Fairfax_VA_2024",
         "attribution": {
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks, looks like I forgot to update the name/id in #2401

wcedmisten:
@arch0345 do you know how long it takes for these changes to appear on OSM.org? I'm still seeing the broken Fairfax layer from before https://github.com/osmlab/editor-layer-index/pull/2353

image

arch0345:
Since this index is bundled with iD, unfortunately this means that the imagery list only updates when a new iD version is released. A new iD release is long overdue, but from what I can gather from https://github.com/openstreetmap/iD/issues/10412#issuecomment-2295140688, it might take a few weeks before this happens. In the meantime, you can use the URL for this imagery listed on https://osmlab.github.io/editor-layer-index/ as a custom source

PR #2458, opened 2024-08-22
Moldova imagery updates

Oni-DOS:
Added new imagery and added up to date data for old ones, removed 2016 reserver server because this server will be shutted down this year. Also about the permitting of using these imageries: the AGCC (later known as ARFC), the owner and maintainer of these imagery changed the access and using constraints to "No limitations to public access" and "No conditions to access and use", these can be seen in the imageries files metadata at license_url:. So this means that these imagery can be used without any constraints in OSM.


Oni-DOS:
Is the available_projections parameter very important for merging these files in the main branch?

arch0345:
> Is the available_projections parameter very important for merging these files in the main branch?

Yes, this is needed for editors like iD to know what value(s) to use for the {proj} parameter in the url

sources/europe/md/Orthophoto_2007_38towns.geojson
```diff
@@ -0,0 +1,3528 @@
...
+		"license_url": "https://geoportalinds.gov.md/geonetwork/srv/eng/catalog.search#/metadata/a648ff10-499b-4126-b986-6b29ddd569fa",
+		"privacy_policy_url": "https://www.legis.md/cautare/getResults?doc_id=143076",
+        "attribution": {
+            "url": "https://geoportalinds.gov.md/geonetwork/srv/eng/catalog.search#/metadata/a648ff10-499b-4126-b986-6b29ddd569fa",
+            "text": "© Agenţia Geodezie, Cartografie şi Cadastru / Geoportal INDS",
+            "required": true
+        }
+    },
+    "geometry": {
+        "type": "Polygon",
```
arch0345: It looks like this geojson is made up of multiple outer rings, so the type MultiPolygon should be used. Here's the formatting for geojson multipolygons (basically an array of polygons): https://docs.mapbox.com/android/java/api/libjava-geojson/5.8.0/com/mapbox/geojson/MultiPolygon.html
```suggestion
        "type": "MultiPolygon",
```

arch0345 (changes requested):

arch0345:
There should be an extra pair of brackets between each outer ring, see the documentation url I linked

Oni-DOS:
So what is the problem now? :D

arch0345:
It looks like some of these layers have a `wms:` prefix in the url, which seems to come from the  JOSM index. This prefix should be omitted in ELI

I can't figure out where the geometry error is coming from

Oni-DOS:
I checked with JOSM these files for geometry errors and it didn't finded them

arch0345 (approved):

PR #2459, opened 2024-08-23
Update MinnesotaCompositeImageService.geojson - Remove unavailable projection EPSG:3857

danieldegroot2:
Remove unavailable projection from Minnesota Composite layer, see https://github.com/openstreetmap/iD/issues/10430

It seems only the originally listed projection for NAD 83 is available at the server for this and similar resources.
( in particular I looked at
https://resources.gisdata.mn.gov/pub/gdrs/data/pub/us_mn_state_metc/base_metro_orthos_2020/metadata/metadata.html
)

### DIFF
--- a/sources/north-america/us/mn/MinnesotaCompositeImageService.geojson
+++ b/sources/north-america/us/mn/MinnesotaCompositeImageService.geojson
@@ -805,7 +805,6 @@
             "url": "https://www.mngeo.state.mn.us/chouse/wms/composite_image.html"
         },
         "available_projections": [
-            "EPSG:3857",
             "EPSG:26915"
         ],
         "country_code": "US",


tordans:
But https://imageserver.gisdata.mn.gov/cgi-bin/mncomp?LAYERS=mncomp&STYLES=&CRS=EPSG:3857&BBOX=-10528589.148591155,5827559.035650633,-10528436.274534605,5827711.909707184&FORMAT=image/jpeg&WIDTH=256&HEIGHT=256&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap works fine, so `3857` is available.

And iD actually requires `3857` given the wording in image

tyrasd:
The WMS does not report EPGS:3587 in the `GetCapabilities` call (see https://imageserver.gisdata.mn.gov/cgi-bin/mncomp?LAYERS=mncomp&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities), but as @tordans correctly assessed, it does return data in that projection, see e.g.:

  https://imageserver.gisdata.mn.gov/cgi-bin/mncomp?LAYERS=mncomp&STYLES=&CRS=EPSG:3857&BBOX=-10377969.984376183,5617892.267093581,-10377931.765862044,5617930.485607719&FORMAT=image/jpeg&WIDTH=256&HEIGHT=256&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap

I'm also not 100% sure whether @knkski original impression from https://github.com/openstreetmap/iD/issues/10430#issue-2483489920 is true: it could also be the case that Bing/ESRI are not perfectly accurately georeferenced in this area. see for example the good alignment of the parking lot outline and tribune building in the Minnesota Composite Imagery compared to Bing and ESRI:

mncomp:

image

Bing:

image

ESRI.

image


I also checked the layer in QGIS, and the same small offset between the mapped trees and the mncomp layer is also present in the `EPSG:26915` projection.

danieldegroot2:
What about the Hennepin layers?

tyrasd:
Yeah, ok, that layer seems to match Bing/ESRI as well. Still, it's very hard to tell which layer is actually the accurate on (even when comparing with GPS traces), as they differ only by a small offset. Maybe @DcJoeS or @arch0345 know more?

But that said, it's totally normal to have small offsets in any imagery layer. In some regions these offsets can even be large (in the 10s of feet/meters). That's why editors like iD have a built-in _background offset_ compensation tool:

image

I think there's no reason to remove this layer just because of a small potential offset, let's close this.

danieldegroot2:
Closing this PR is okay with me. Others can still comment on this issue if they want.

knkski:
I don't know exactly what the scope of the project is here, but it seemed like it would be nice to fix the issue once for everyone instead of having everyone realize that the MCIS data is offset, and then manually adjusting to fix it, so I opened the issue.

There's general agreement between Bing, Hennepin County Orthoimagery, and other imagery datasets, MCIS is the odd one out. I think people have been tracing roads and other features with the unadjusted MCIS dataset, because lots of them are off when viewed against any other imagery dataset. Here's an example image of how that dataset is off, based on the above pictures. I've added a red line connecting the two:

image

I brought up the potential projection issue based on the MCIS imagery being offset, and the response from the contact for the dataset saying that might be the cause. I'm not really familiar with projections enough to know if that's the root cause.

The context of filing the issue is that I'm updating OSM using iD, and think it would be nice to fix the offset issue for everyone. Based on @tyrasd's comment above that

> The WMS does not report EPGS:3587 in the GetCapabilities call

Is this a bug with the MCIS dataset/servers that I can report to them to get fixed? If not, would it make sense to automatically apply an offset so that people don't have to realize that there's an issue and fix it each time themselves?

See here for an example of where someone likely added a feature to the map using MCIS and it was noticeably offset:

https://www.openstreetmap.org/edit#map=21/44.9864352/-93.1694164

image

The Minnesota State Fair is going on now, and I thought it would be nice to correct features around the fairgrounds, and I noticed how off everything was, hence why I filed the issue

danieldegroot2:
A chair lift usually means elevation difference and not the best object to use for accuracy. (though, this one seems to be a theme ride)
However, this time it aligns with Esri -which is current Maxar imagery- and Esri Clarity -which is very old Bing imagery-.
Usually newer Bing imagery would be better aligned than old Bing imagery, but it is not necessarily the case.
There have been instances where imagery updates were made with the wrong projection (usually Bing in the UK iirc)

https://livingatlas.arcgis.com/wayback/#active=10&mapCenter=-93.16906%2C44.98638%2C19
(Esri is latest, Esri Clarity is oldest imagery in the Wayback archive)

knkski:
Here's probably a slightly better example for the MN State Fair grounds (though yes, that's a fair ride with no elevation change for the base of the poles). The roads and sidewalks in this intersection line up nicely for the MCIS dataset, and are off for all the other ones:

https://www.openstreetmap.org/edit?way=177770013#map=20/44.9812679/-93.1690051

Again, I don't really know what the OSM project(s) consider in-scope vs out-of-scope, but it seems like it would be a good idea to make sure people don't unknowingly add mispositioned features. I only figured out that the MCIS imagery dataset was offset after a while of wondering why features were placed oddly. If this is considered out-of-scope and there's no bug to report back to the MCIS contact, that's fine.

tordans:
@knkski something that handles the offset of imagery is out of scope for this repo and iD. There is some more info on this at https://github.com/openstreetmap/iD/issues/4764 and https://github.com/osmlab/osm-planning/issues/18 and scattered around in discussions for many years.

Personally I find it less of an issue than one might think first, especially since we continue to get better image sources in areas that are mapped in detail.

It would be nice to have a shared offset database, though! But I don't see anyone working on it any time soon ATM.

danieldegroot2:
This area has not been touched recently.
Road was added a very long time ago. It is unlikely we had (access to) this dataset.
Road and buildings don't line up in some places because they were added with older imagery which may have been less accurate. Many roads were [imported from the TIGER dataset](https://wiki.openstreetmap.org/wiki/TIGER)
It may also line up less because of the trees.

PR #2472, opened 2024-08-27
Slippy map Ny-Ålesund tiles for iD

zabop:
Some Svalbard layers are available from WMTS endpoints, but they do not support slippy map tiling. Hence, they are unavailable in iD. This commit aims to address this issue for one a small area ([Ny-Ålesund](https://www.openstreetmap.org/node/274786503#map=19/78.925530/11.929166)) of [one such layer](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/no/NPISvalbardsatellite.geojson).

The tileset I am proposing to include is available using this endpoint: [https://zabop.github.io/svalbard-xyz/tiles/{z}/{x}/{y}.jpg](https://zabop.github.io/svalbard-xyz/tiles/%7Bz%7D/%7Bx%7D/%7By%7D.jpg). I created the tiles using QGIS and the QTiles plugin. The area I chose to tile is the area covered by webmercator tile [13/4367/1053](https://lp-tools.toolforge.org/misc/bbox.html?sw=78.92083162470739,11.953125000000016&ne=78.92927320597397,11.90917968749994). Deepest zoom level: 18.

The GeoJSON added in this PR is based on [this GeoJSON](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/no/NPISvalbardsatellite.geojson). Unfortunately, `license_url` and the `icon` URLs do not work anymore. I aim to find up to date URLs for these two fields. I hope that once I have a proper `license_url`, it'll make it clear that it is allowed to retile the imagery from the [original WMTS](https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_25833/MapServer/WMTS/) before using it for OSM edits.

Once (if?) this works, and Svalbard has a new imagery layer available in iD, I aim to expand the tiled area and come up with proper infrastructure to serve the tiles. Using [GitHub pages](https://github.com/zabop/svalbard-xyz/tree/main) to do so is not ideal, but since the tiled area is so small, I think it'll do as a first step towards something greater.

### DIFF
--- a/sources/europe/no/NPI_Svalbard_satellite_tiled.geojson
+++ b/sources/europe/no/NPI_Svalbard_satellite_tiled.geojson
@@ -1,0 +1,55 @@
+{
+    "type": "FeatureCollection",
+    "features": [
+      {
+        "type": "Feature",
+        "properties": {
+          "id": "npi-svalbard-landsat-tiled",
+          "name": "NPI Svalbard satellite (tiled)",
+          "type": "tms",
+          "url": "https://zabop.github.io/svalbard-xyz/tiles/{z}/{x}/{y}.jpg",
+          "country_code": "NO",
+          "available_projections": [
+            "EPSG:25833"
+          ],
+          "attribution": {
+            "url": "https://geodata.npolar.no/",
+            "text": "© Norwegian Polar Institute / USGS Landsat | loaded as tiles"
+          },
+          "icon": "https://www.npolar.no/npcms/export/sites/np/images/logos/NPI-logo-eng.png",
+          "min_zoom": 13,
+          "max_zoom": 18,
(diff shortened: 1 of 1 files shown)


zabop:
We don't need to retile the WMTS ourselves, there is a WMTS serving EPSG:3857 tiles:

https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_3857/MapServer/tile/{z}/{y}/{x}?blankTile=false

Preview [here](https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_3857/MapServer?f=jsapi). Found via [this list](https://geodata.npolar.no/).

PR #2473, opened 2024-08-28
Adjust geometry for Svalbard orthophoto layer

zabop:
I believe the geometry for the [NPI Svalbard Orthophoto](https://github.com/osmlab/editor-layer-index/blob/35a9bbf09bc37ecad9d9a0ab52a0537da3e3ba15/sources/europe/no/NPISvalbardOrthophoto.geojson) layer was drawn at a time when this layer covered a smaller area than it currently does. Using the [WMTS endpoint](https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_25833/MapServer/WMTS/), I loaded the imagery to QGIS, visualized the currently defined geometry (green line) and draw a new border (red line):
Screenshot 2024-08-28 at 23 10 02
I think updating the borders could help mappers mapping areas outside the old border but inside the new border. (For example, the [station at Isbjørnhamna](https://en.wikipedia.org/wiki/Polish_Polar_Station,_Hornsund) is such a place.)


matheusgomesms:
Thank you! Please check the check (heh!) and update the geometry to respect ring orientation.

Also, please update the logo to: 
`https://www.npolar.no/wp-content/uploads/2022/05/NP-logo-skjerm.png`

It seems NO mailing list was updated and I could not find the permission anymore in the mailing list. Perhaps it was moved to Wiki, and you could replace it here?

Tusen takk!

zabop:
Thank you! I believe I was able to address the errors (I found the [license link](https://web.archive.org/web/20240114010212/https://lists.nuug.no/pipermail/kart/2018-February/006371.html) thanks to [this thread](https://community.openstreetmap.org/t/hvor-finner-jeg-lisensen-for-npi-svalbard-orthophoto/118091)). The least confident I am in the [`privacy_policy_url`](https://www.npolar.no/en/privacy-policy/): perhaps there is a better link thank what I managed to find.

sources/europe/no/NPISvalbardOrthophoto.geojson
```diff
@@ -1,223 +1,516 @@
...
+  "type": "Feature",
+  "properties": {
+    "id": "NPI-Svalbard-ortofoto",
+    "name": "NPI Svalbard Orthophoto",
+    "type": "wmts",
+    "url": "https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_25833/MapServer/WMTS/",
+    "country_code": "NO",
+    "available_projections": [
+      "EPSG:25833"
+    ],
```
arch0345:
```suggestion
    "type": "tms",
    "url": "https://geodata.npolar.no/arcgis/rest/services/Basisdata/NP_Ortofoto_Svalbard_WMTS_3857/MapServer/tile/{zoom}/{y}/{x}",
    "country_code": "NO",
```

arch0345 (commented):
I recommend using the 3857 MapServer so that this imagery will be accessible in more editors such as iD

zabop:
Good idea! I thought about adding the EPSG:3857 as well, but just replacing this with that one is even simpler.

grischard:
Is all of Svalbard covered? A simple rectangle would work if so.

zabop:
Thanks @grischard, I think all of it is covered, yes! Now the geometry is a simple rectangle.

PR #2476, opened 2024-08-29
Moldova: New topomaps (historic and actual)

Oni-DOS:
Hi guys, I added a lot of historic and modern topomaps, they the same have "No limitations to public access" and "No conditions to access and use", which means they can be freely used in OSM, for details see license_url: to access their metadata on geoportal.


Oni-DOS:
Is there anyone alive here?

matheusgomesms:
Thank you for all your work!

Are all these layers necessary to be added, such as from the 60 years ago?

I would keep only the 2007 and forward orthos, and the new topo (from past 10 years), but I will let others chime in.

Oni-DOS:
Hi, that will look strange but the old topomaps are still in use by gov and local administrations. Also, on old topomaps, are present the names for forests, hills, valleys, some small rivers, lakes, small islands on Prut and Nistru river. Some of this data is present on topo map from 2013, but is very limited. I don't see a problem for this data to be present in index, especially because it is also used by OHM, and is marked as historical. And in the end - you can see the perspective how cities and infrastructure was developed and it can be more useful for mapping.

Oni-DOS:
Sooooo? Guys, whats the problem with this pr? It has one month from its opening. Someone will merge it? @grischard @arch0345 @rbuffat

grischard:
Please don't ping individual maintainers.

It's a big PR, that makes it a bit tedious to review.

Editors are free to filter out historic layers; I believe only iD does it in practice.

tyrasd:
FYI: This PR has unfortunately broken the [`deploy`](https://github.com/osmlab/editor-layer-index/actions/workflows/push.yml) workflow, because it added files with a colon (`:`) in the file names, which the makefile unfortunately cannot process. //edit: this should be fixed now with f5352bb

Oni-DOS:
> FYI: This PR has unfortunately broken the [`deploy`](https://github.com/osmlab/editor-layer-index/actions/workflows/push.yml) workflow, because it added files with a colon (`:`) in the file names, which the makefile unfortunately cannot process. //edit: this should be fixed now with [f5352bb](https://github.com/osmlab/editor-layer-index/commit/f5352bbaf0a8fc0137cdc9aa3ff0d9b257bfbfea)

I'm sorry for causing this problem, didn't know that adding colons in names can cause problems like this, but that can help in future, for example if someone will change the automatic pr check's code to find colons and any forbidden symbols in files names then problems like this will be found before the closing of pr.
Sorry, and thank for fixing 👍

tyrasd:
No worries, I didn't know that myself either. :sweat_smile:  And good ideas to add a check to the PR workflow. I've added one in 01c54ccf now.

PR #2477, opened 2024-08-30
Berlin2024: Fix missing `max_zoom`

tordans:
Right now the images disappear when zoomed in too far.
https://www.openstreetmap.org/edit?node=599873093#map=20/52.4714604/13.4496780

We have this setting for [2022](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/de/Berlinaerialphotograph2022.geojson?short_path=7b50115), [2023](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/de/Berlinaerialphotograph2023.geojson?short_path=61f40d5)

No idea why I did not add it to 2024…


matheusgomesms:
Vielen dank!

PR #2479, opened 2024-09-01
Create SouthTyrol-Cadastre.geojson

ilselvaggio:
Cadastre map from wms https://geoservices5.civis.bz.it/geoserver/p_bz-Cadastre/wms?request=getCapabilities


matheusgomesms:
Thank you!

Is this layer working? I didn't manage to make it work on iD, JOSM, or even in their GeoServer.

sources/europe/it/SouthTyrol-Cadastre.geojson
```diff
@@ -0,0 +1,934 @@
+{
+    "type": "Feature",
+    "properties": {
+        "attribution": {
+            "url": "https://geoservices.buergernetz.bz.it/geokatalog",
+            "text": "© Autonome Provinz Bozen/Provincia Autonoma di Bolzano CC-BY 4.0",
+            "required": false
```
andrewharvey, line 7: if the data is CC BY then attribution is required so should be `required: true`

andrewharvey (changes requested):
Thanks for the PR and apologies it's been missed.

The service is working for me, but it looks like it would work better as an overlay above imagery rather than a background map. Could you set `overlay: true`? Per the schema at https://github.com/osmlab/editor-layer-index/blob/d3f5b9255d49d23e7f9d107002515dfa7fcfbab0/schema.json#L162

sources/europe/it/SouthTyrol-Cadastre.geojson
```diff
@@ -0,0 +1,934 @@
...
+            "EPSG:25832",
+            "EPSG:32632"
+        ],
+        "type": "wms",
+        "category": "photo",
+        "id": "SouthTyrol-Cadastre",
+        "name": "South Tyrol Cadastre",
+        "start_date": "2024-09-02",
+        "end_date": "2024-09-02",
+        "best": true
```
andrewharvey, line 25: Could we remove `best: true` this should be for imagery layers only.

sources/europe/it/SouthTyrol-Cadastre.geojson
```diff
@@ -0,0 +1,934 @@
...
+        "url": "https://geoservices5.civis.bz.it/geoserver/p_bz-Cadastre/ows?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=ParcelsPolygons&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}",
+        "max_zoom": 20,
+        "country_code": "IT",
+        "available_projections": [
+            "EPSG:3857",
+            "EPSG:25832",
+            "EPSG:32632"
+        ],
+        "type": "wms",
+        "category": "photo",
```
andrewharvey, line 20: could you change the category to either `map` or `other` since it's not an orthophoto. https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#categories

sources/europe/it/SouthTyrol-Cadastre.geojson
```diff
@@ -0,0 +1,934 @@
...
+        "id": "SouthTyrol-Cadastre",
+        "name": "South Tyrol Cadastre",
+        "start_date": "2024-09-02",
+        "end_date": "2024-09-02",
```
andrewharvey, line 24: If the data is continually updated, best leave off the start/end dates. Unless you know the data is fixed at this time period?

andrewharvey:
Could you also please remind me where the CC BY waiver is?

PR #2484, opened 2024-09-04
Improve French imagery with description

Nekzuris:
I also corrected zoom levels and reordered some parameters for consistency.


matheusgomesms:
Merci!

PR #2491, opened 2024-09-15
Make Jan Mayen topo available in iD

zabop:
Jan Mayen now has a topography WMTS available in EPSG:3857. I found the service [here](https://geodata.npolar.no/). Previously, the EPSG:25829 version was added, which iD couldn't use. I hope this update will benefit iD users (while not effecting anyone adversely, as JOSM and other tools can use the new version as well).

This PR is modelled on #2473.


arch0345:
Thanks, would you be able to convert this to a tms source like in #2473? Some editors like iD aren't able to parse WMTS endpoints

zabop:
Thank you @arch0345! Could you check again? I believe it's good now.

Issue #2443, opened 2024-08-05
Potential imagery sources for German states


arch0345:
We currently have imagery layers for 8 out of the 16 states in Germany. The imagery for the remaining 8 states are listed below. These are either licensed under DL-DE->BY-2.0 or CC BY 4.0, so explicit permission needs to be granted from these agencies before they are added to ELI.

- [x] Baden-Württemberg: https://www.geoportal-bw.de/geonetwork/srv/ger/catalog.search#/metadata/4927a198-e9ce-da9a-a5ef-c11f21ff173f (#2737)
- [ ] Bremen: https://metaver.de/trefferanzeige?docuuid=3dd11c0f-9c13-4ada-8b3a-885aa979e329
- [x] Mecklenburg-Vorpommern: https://advmis.geodatenzentrum.de/trefferanzeige?docuuid=b6a74a2b-9d69-4f74-ba53-21111da7851f (#2816)
- [x] Niedersachsen: https://ni-lgln-opengeodata.hub.arcgis.com/apps/lgln-opengeodata::digitales-orthophoto-dop20/about (#2807)
- [x] Rheinland-Pfalz: https://www.geoportal.rlp.de/mapbender/php/mod_showMetadata.php?languageCode=en&resource=layer&layout=tabs&id=61676#tabs-5 (#2806)
- [x] Saarland: https://geoportal.saarland.de/mapbender/php/mod_showMetadata.php?languageCode=en&resource=layer&layout=tabs&id=46303#tabs-5 (#2507)
- [ ] Schleswig-Holstein: https://opendata.schleswig-holstein.de/dataset/digitale-orthofotos-1-5000dop20
- [x] Thüringen: https://inspire-geoportal.ec.europa.eu/srv/api/records/10cbd9e4-8a84-4771-8ed1-a3cf66a01766?language=all (#2495)

arch0345:
@coolultra1 would you be able to contact these government agencies? I would, but I figure it would be better for someone fluent in German to do so

coolultra1:
@arch0345 Of course, thanks for collecting these. It could be German OSM contributors already tried to contact some, so I'll check the wiki first.

Will do this within this week.

coolultra1:
Thuringia says yes - Will add soon.

rbuffat:
[...]

PR #2495, opened 2024-09-17
Add Thuringia DOP20

coolultra1:
See also #2443 

Via Email, the responsible authority, the Thuringia Landesamt für Bodenmanagement und Geoinformation has confirmed me deriving data for OSM does not fall under usage under dl-de/by 2.0 under their assumption, as long as, when the imagery is shown in the editor, it's attributed correctly. (which is the case)


coolultra1:
Failed check is to be ignored as file will be reachable once merged.

arch0345:
Thanks for confirming with them, would you be able to upload a PDF of the email and list that as the license URL?

coolultra1:
@arch0345 

I've put the mail correspondence into a wiki page and linked that as the license URL. Is it okay like this?

arch0345:
Yes, that's fine

PR #2496, opened 2024-09-17
Add icons for all German imagery

coolultra1:
This adds icons for all German imagery sources that don't already have one. For places, their [Coat of Arms](https://en.wikipedia.org/wiki/Coat_of_arms) is used, a type of official symbol which is generally very well recognized, unique and does not fall under any type of copyright law. For the Deutsche Bahn, their logo is used. All icons are 100x100px.


coolultra1:
Failed checks are to be ignored as files will be reachable once merged.

grischard:
Cool contribution, thank you!

PR #2499, opened 2024-09-19
Switch to PNGs for FI background maps

todeveni:
Infrastructure related to Kapsi background maps have been upgraded and JPG support is still lacking.


matheusgomesms:
Thank you!

PR #2500, opened 2024-09-19
Update OrtoSat2023 imagery geometry (Portugal)

mikedld:


grischard:
The detailed coastline needlessly takes up space, can you simplify it like shown in http://download.geofabrik.de/europe/portugal.html but presumably without the islands?

mikedld:
I wasn't tracing the coastline, rather the actual limits of the imagery. Do you maybe want me to simplify all of it, not only the coast part?

grischard:
In general I'd say anything that simplifies within 500 metres outside the imagery boundary is fine. Over the water you can just pull a very simple boundary with 3-4 points.

mikedld:
Steps for the land part are about 4-5km so I'm not sure I can meet the 500m rule. Reduced the water boundary to 4 points now.

matheusgomesms:
Looks great I believe, thank you (again) @mikedld!

For clarification: Portuguese government recently update their imagery to cover the whole country, not only the central part. So the layer is the same, with the update of the new geometry (covering whole continental part) and the name.

Issue #2379, opened 2024-06-30
strict_check.py checked-in icons unaccounted for


Intrinsica1:
`strict_check.py` doesn't account for icons that are checked into the repo, but not yet deployed to github-pages.

```
ERROR:root:sources/north-america/us/pa/PEMA_Orthoimagery_2022.geojson: icon url https://osmlab.github.io/editor-layer-index/sources/north-america/us/pa/PEMA.svg is not reachable: HTTP code: 404
```

This issue was encountered on pull #2373.

PR #2507, opened 2024-09-25
Add Saarland and Baden-Würrtemberg imagery

coolultra1:
I've got written permission from these two states. Both correspondances are on the Wiki and linked as license URLs. Also See #2443.


coolultra1:
Failed run can be ignored because of #2379, so should be mergable.

I don't even think I'm that dumb or incapable, but I really wish adding imagery was easier/more intuitive. It's such a thankless task anyway. Or at least have the JOSM index killed off, finally. Or at least I should wonder why the python validation scripts won't run in my env.

vizsim:
I noticed that Saarland is available, however Baden-Württemberg not:
https://osmlab.github.io/editor-layer-index/

~~Maybe the typo in Baden-Württemberg (not Baden-Würrtemberg) is a problem?~~

EDIT: ah, makes sense as it has been removed:
https://github.com/osmlab/editor-layer-index/pull/2513

are there any news?
https://wiki.openstreetmap.org/wiki/DE:Permissions/Luftbild_Baden-W%C3%BCrrtemberg
ps: some personal informations in the answer should probably be blanked out

Thank you @coolultra1 for all the work!

coolultra1:
@vizsim No news from Baden-Württemberg yet, last contact was "We'll Check and definitely contact you on the results."

I'll probably, as it's unused, deleted the Wiki page for now and then redo it properly once the time hast come. 

Thank you for your interest, glad to be of any help.

PR #2511, opened 2024-09-28
Add updated Croatian imagery: dgu.hr: Croatia 2021-2022 aerial imagery

mnalis:
and mark previous _"dgu.hr: Croatia 2019-2020 aerial imagery"_ as historic.


matheusgomesms:
Thank you!

PR #2438, opened 2024-08-02
Update layers for Bavaria

arch0345:
Replaces 80cm orthophoto with 20cm version (both use the same source)

Also adds the following layers:
- 20cm color infrared orthophoto
- Lidar/hillshade
- Bavaria topographic map
- Bavaria web map
- Parcel/building overlay
- Contour line overlay

These layers are licensed under CC BY 4.0, however [a representative from Bayerischen Vermessungsverwaltung stated](https://wiki.openstreetmap.org/w/images/b/bc/20230526_Nutzung_der_OpenData-Produkte_der_Bayer._Vermessungsverwaltung.pdf) that listing them on the Contributors page is sufficient for attribution so [this data is approved to be used for OSM](https://wiki.openstreetmap.org/wiki/Contributors#Bayerische_Vermessungsverwaltung_%E2%80%93_www.geodaten.bayern.de).

Closes #2437

PR #2513, opened 2024-09-29
Remove BW-DOP20 due to unclear license

coolultra1:
- So, um, this is embarrassing. -

It was pointed out to me by another mapper reading the Wiki page for this imagery, the respective authority never actually permitted the usage, just said it was unclear and they'd check, meaning we can't (yet) use this imagery. This is a clear error of myself, as I just skipped over their response.

In consequence, I've re-checked permission on all imagery I have added so far and am relieved to see this is the only occurrence of unclear licensing in my work. (and, for that, in all of German imagery)

So for the time being, I am sorry for the bother and hoping this can be re-added soon.


matheusgomesms:
Thanks for the quick fix!

Can you at least keep the icon? In the future we can use it again, if they clarify the permission.

grischard:
Thank you. At least this didn't make it into an iD release. 

We can add the icon back from git history if and when it's necessary.

cvzi:
This was released in ID 2.30.3

tyrasd:
Argh… iD 2.30.3 was released to resolve another imagery licensing issue in Germany (Bavaria), see https://github.com/osmlab/editor-layer-index/pull/2438#issuecomment-2304393789 :see_no_evil: 

For some reason the `deploy` action on ELI's github repo seems to be broken at the moment: https://github.com/osmlab/editor-layer-index/actions/workflows/push.yml, so this fix did not make it into the published `imagery.json` file. :scream: At first glance, it seems the files containing colons (`:`) introduced in [#2476](https://github.com/osmlab/editor-layer-index/pull/2476/files#diff-9accd3845f10207c9a96c0e46866feeeeb223b5a892109a4b8d135908cc09be9) might be the root cause of that.

I'll push another patch release of iD to fix this issue tomorrow to remove the BW-DOP20 layer again.

matheusgomesms:
Hi @tyrasd, thank you for your amazing work!

Will the new release update only the BW-DOP20 layer, or fetch all the latest version of ELI?

If not, can you please include #2524 in the iD patch as well? Portuguese community complained a lot about this change, so it would be great to come back to the previous state. Thank you!

PR #2520, opened 2024-10-04
Add WMS sources for Environment Agency LiDAR in England

russss:
These are licensed under the Open Government License which is compatible with OSM. The geometry is a simplified version of the extent polygons provided.


grischard:
Is the polygon taken from the Environment Agency itself? Could it please be simplified to just a few points over the water?

grischard:
Would the polygon at http://download.geofabrik.de/europe/united-kingdom/england.html minus the hole in Yorkshire work?

russss:
I've simplified the extents based on the Geofabrik polygon (the layer does also stray slightly into Wales and Scotland in places)

PR #2521, opened 2024-10-05
Make latest FÖMI orto the default in Hungary

gabortim:
The latest FÖMI layer is _always_ 10 years behind, but government-based and still the best source for mapping and alignment. This should help new mappers to draw anything accurately.

### DIFF
--- a/sources/europe/hu/Hungary-FOMI-2011-2014.geojson
+++ b/sources/europe/hu/Hungary-FOMI-2011-2014.geojson
@@ -11,6 +11,7 @@
         },
         "start_date": "2011-08-11",
         "end_date": "2014-06-10",
+        "best": true,
         "url": "https://orto1.tile.openstreetmap.hu/fomi2011-2014/{zoom}/{x}/{y}.webp",
         "min_zoom": 8,
         "max_zoom": 18,


grischard:
10 years behind sounds like a lot. Could this be discussed in the Hungarian community?

gabortim:
Hi grischard,

Yes, it is a lot, but the government does not make available it any sooner. The community has no influence on this and in a lack of reliable source, this is still a sensible option IMO.

There is a law which would enforce the publication of such sources, but no member state heard about it, I guess: https://eur-lex.europa.eu/legal-content/EN/ALL/?uri=CELEX%3A32023R0138

It's not a direct answer to your question, more like a side note.

zabop:
Interesting side note @gabortim. I'm not completely sure it's the same EU law, but Sweden is [introducing changes](https://www.lantmateriet.se/sv/geodata/vara-produkter/Produktnyheter/Allmanna-nyheter/vardefulla-datamangder---forandringar-pa-gang/) to its aerial imagery access policy based on an EU directive on "High Value Data". (I heard about this [here](https://community.openstreetmap.org/t/lm-ortofoto-nedladdning-cog-api-hackathon/120366).) If more and more states follow, I hope the Hungarian refusal will become increasingly untenable.

andrewharvey:
>  The community has no influence on this and in a lack of reliable source, this is still a sensible option IMO.

Has the Hungarian OSM community discussed the change to make FÖMI orto the default in Hungary.

I find it's usually a balance between more recently updated Bing/ESRI but lower resolution and higher resolution/better location accuracy government imagery but less recently updated.

andrewharvey:
Looks like there is newer imagery as well #2618, would that invalidate this PR?

gabortim:
> Looks like there is newer imagery as well #2618, would that invalidate this PR?

No, see the coverage [here](https://josm.openstreetmap.de/mapsview?entry=F%C3%96MI%20orthophoto%202015%20leaf-off%2020%20cm).

Also, as discussed there, the shadows are causing some extra guesswork, which is not ideal, especially for alignment --  the main reason to use old ortophotos.

But #2630 ([coverage](https://josm.openstreetmap.de/mapsview?entry=F%C3%96MI%20orthophoto%202015)) can be made default IMO. Let's continue there as that is the latest freely available FÖMI for now.

PR #2522, opened 2024-10-06
update New York State bounding polygons

jmapb:
enlarge bounding polygons on statewide imagery layers to include Staten Island, Ellis Island, Liberty Island


arch0345:

PR #2527, opened 2024-10-08
Update PAgis_Ortho_2021.geojson

cp880:
Updated max_zoom level from 19 to 22.  Using JOSM I found that this map layer will zoom up to level 22 providing much more detail than the currently configured level 19.


zabop:
Shouldn't the URL point to a TMS or WMTS source and not a WMS one? `max_zoom` makes sense when there are tiles, and `WMS` doesn't have tiles. (I might be missing something.)

cp880:
I think max_zoom=19 tells JOSM to not request imagery from that source when JOSM's zoom level is >19, regardless of the imagery protocol used.

If I add this imagery URL with max zoom of 22 to JOSM, then I'm able to see more detail at zoom levels 20-22 compared to the detail available with the WMS URL generated by this Edit Layer Index (which sets max_zoom=19).

URL:
wms[22]:https://www.pagis.org/arcgis/rest/services/MAPS/AerialPhotos2021/MapServer/export?f=image&format=jpg&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&foo={proj}

Here is the detail available with max zoom=19:
image

Compared to max zoom=22:
image

zabop:
Oh great, thanks! I didn't know this is the case.

arch0345:

PR #2528, opened 2024-10-09
Minor text fixes and removed end_date key

Oni-DOS:
Hi guys, i fixed some names and removed end_date keys because older than 2006 imagery and maps aren't showed in iD, This should fix it. As how i've seen in other imagery from other countries this key is very rare used and doesn't affect some user experiences because users don't see them in UI, all the creation dates are duplicated also in name key.


Oni-DOS:
There is no problem with geometry. It was verified in the pr when the files were added

tyrasd (changes requested):
> removed end_date keys because older than 2006 imagery and maps aren't showed in iD

That's not a good reason to remove the attribute. :sweat_smile: 

Do you think the [20+ years old](https://github.com/openstreetmap/iD/blob/v2.30.4/scripts/update_imagery.js) layers are really useful for "regular" mapping in iD?

Oni-DOS:
Absolutely these are really useful for mapping, there are a lor of data that can be found only on these layers, like hills and valleys names, names of small rivers, islands. Also they can be useful for seeing some data in perspective. Of course they can't be used independently because they are outdated but can be a very useful basis for a more detailed view of objects on the map. For example I used them a lot in iD when I added hill names. Also, regarding the fact that they are outdated, they are still used by local authorities, despite the fact that they have more actual data.

grischard:
Can you please separate the text changes in another PR?

The right place to discuss and potentially change the way iD includes layers is over in its repository.

Issue #574, opened 2018-10-25
providing a stable output for iD


andrewharvey:
Snipping a quote from https://github.com/openstreetmap/iD/issues/5412#issuecomment-431020647
> ...it can't happen currently because ELI doesn't actually issue stable releases like all the other software that iD depends on. I'll probably end up rewriting it someday. Anyway faster turnaround on imagery attribution text changes is not really a huge priority for most people.

Currently the imagery index in iD is only updated each iD release. It's almost been 2 months now since the last iD release which means their imagery index is almost 2 months out of date. I'd like to see that improved, but not sure on the best path forward.

@bhousel you said this is because ELI doesn't issue stable releases, from my understanding after each commit Travis CI will test and rebuild the index, so is `https://osmlab.github.io/editor-layer-index/imagery.geojson` not considered the stable up to date release which iD could pull each day or so for updates?

bhousel:
> @bhousel you said this is because ELI doesn't issue stable releases, from my understanding after each commit Travis CI will test and rebuild the index, so is `https://osmlab.github.io/editor-layer-index/imagery.geojson` not considered the stable up to date release which iD could pull each day or so for updates?

Sometimes the build is broken - this is just a fact of development.  It happens to this index sometimes.  When it happens, Potlatch doesn't work.

For iD, I'm much more comfortable depending on software that's been published by a human.  Someone should take a quick look at it and say "ok looks good" before incrementing the version number and typing `npm publish`.  That's not really asking a lot.  This index doesn't really change much.

Before I publish a new version of iD, I grab the latest copy of ELI, run my script on it, and do a `git diff` to see what actually got changed.

Instead - when ELI does change, [semantic versioning](https://semver.org/) can be used to signal to downstream projects how disruptive the change is.  

- A change to an imagery source in the index could be a patch release, and iD could pull these automatically.  
- A new property added to the schema could be a minor release.  It would be easy to handle but I'd probably want to prepare for it.  
- Renaming / removing a property / changing the schema would be a major release, and would break downstream projects. ELI should be allowed to do this for good reasons.

andrewharvey:
> Sometimes the build is broken - this is just a fact of development. It happens to this index sometimes. When it happens, Potlatch doesn't work.

[...]

PR #2530, opened 2024-10-13
add CI script to automatically update a git tag

k-yle:
Currently, anyone who wants to consume ELI's data has to fetch the data directly from the main git branch ([via github pages](https://osmlab.github.io/editor-layer-index/imagery.geojson) or [github](https://github.com/osmlab/editor-layer-index/raw/gh-pages/imagery.geojson)), since this repository does not do versioning. 

This can be risky, and cause confusion when the CI is failing on the main branch (like what happened last week).

To give data consumers more certainty, this PR introduces a weekly cronjob which moves [a git tag](https://github.com/k-yle/editor-layer-index/tags) to the latest commit on the main branch, but only if the CI is passing.

This gives people the flexibility to consume data directly from the main branch, or from this git tag if you want to be more safe.

Tested in a fork of this repository ([logs here](https://github.com/k-yle/editor-layer-index/actions/runs/11311561456/job/31457943990))

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: Release
+
+on:
+  schedule:
+    - cron: '0 7 * * 2'
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    permissions:
+      checks: read
+      contents: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Check that the CI is passing
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const checks = await github.rest.checks.listForRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: '${{github.event.repository.default_branch}}',
+            });
+
+            const isSuccessful = checks.data.check_runs
+              .filter((run) => run.name !== process.env.GITHUB_JOB)
+              .every((run) => run.conclusion === 'success');
+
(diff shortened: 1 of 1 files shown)


andrewharvey:
cc @arch0345 @grischard @rbuffat @cicku 

I had suggested an automated versioned release in #574 but was promptly shut down, but due to changes in the maintainers we should consider this again.

I think what you're suggesting is still an improvement over the status-quo and fine with me though I would make it run daily, given it's automated there's no real reason to delay it a week in my opinion.

I also think we should go further and implement automated version releases pushed to npm.

grischard:
I would rather make sure we only push to the main branch when CI is passing, instead of adding the complexity of tags.

k-yle:
> I would rather make sure we only push to the main branch when CI is passing

I'm not familiar with ELI's CI system, but just a general comment: This is an impossible problem to solve. 

Even if you enable ["Require branches to be up-to-date before merging"](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging), there's still a slight chance that the CI will fail on the mainline branch, like what happened during the week when I raised this PR. 

Neither Git nor GitHub has a technical solution that will ensure the mainline branch is always passing, hence the need for releases[^1]

[^1]: or an indicator of the last stable commit, as proposed here

gy-mate:
I agree—versioning (i.e. tagging) seems like a more stable and verifiable solution.

PR #2560, opened 2024-11-17
update URLs WienGIS

marderh:
see https://cdn.basemap.at/basemap.at_URL_Umstellung_2023.pdf


arch0345:

PR #2576, opened 2024-12-13
Update Layers of Saxony-Anhalt

SafetyIng:
The state surveying administration in Saxony-Anhalt has updated the data, moved it to another service and released certain additional data. These adjustments are hereby incorporated.


grischard:
Hi! The PR checks are currently failing on the lack of a privacy policy URL. What do you think about making the name of the layer more human-friendly, and moving the copyright to the attribution field?

SafetyIng:
Okay, the data protection regulations of the service that offers this, i.e. LVermGeo, which provides the WMS service? Okay, I can add that. 

Regarding the legibility of the name: Here I am something like this in the 
‘Tagging for the renderer’ :D The authority requires that the attribution takes place in this form. But I think we can adapt that a bit too ;)

vsandre:
Fixed #1653 and Fixed #1903

PR #2578, opened 2024-12-16
fix oceania/au/wa main roads source

andrewharvey:


andrewharvey:
I'm happy with the geometry even given the warning.

PR #2583, opened 2024-12-19
Add Cape Town 2024 aerial imagery

Firefishy:
Clone of the Cape Town 2023 imagery with year updated. Imagery is available under the same usage terms as previous imagery.


arch0345 (commented):
It looks like the 2024 imagery only covers parts of the city

Firefishy:
@arch0345 I updated the boundary. Could you check?

PR #2598, opened 2025-01-21
Update FAQ.md with CC BY compatibility

andrewharvey:
Use stronger language around CC BY requirements.

OSMF LWG require a waiver for CC BY and imagery is not compatible for tracing in OSM without one.

Further NC or SA licenses don't have a waiver and as such we can't accept their use.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -14,9 +14,9 @@ imagery sources are directly compatible with tracing on OpenStreetMap and do not
 require any additional permission.  Attribution of such a source is
 not required, but is encouraged, to appear in the .geojson source file.
 
-- :question: [Creative Commons with condition (CC BY, CC BY-NC, CC BY-SA)](https://creativecommons.org/share-your-work/licensing-types-examples/) imagery sources are most likely compatible with tracing on
-OpenStreetMap, but to ensure no objection, we ask you confirm with the owning
-organization by using the [creative commons waiver](#where-can-i-find-an-imagery-license-waiver).
+- :question: [Creative Commons Attribution (CC BY 2.0, CC BY 3.0, CC BY 4.0)](https://creativecommons.org/licenses/by/4.0/) imagery sources are not compatible with tracing on
+OpenStreetMap without obtaining a waiver or premmission for the publishing organization
+by using the [OSMF's Creative Commons waiver](#where-can-i-find-an-imagery-license-waiver).
 
 - :question: For any other imagery sources - open an issue and we will look into it!
 There may be language in the imagery metadata or elsewhere on the owning organizations's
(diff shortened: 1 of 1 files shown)


simonpoole:
Is this based on a new position of the LWG? 

Previously it was that manually tracing from imagery didn't create a derivative of the imagery. Obviously displaying -in- an editor is already covered either by an open licence or by explicit permission. That is why there are the specific request documents that essentially point that out to flush out any potential issues: https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers

In particular and contrary to the new FAQ text, ND and NC licensed sources were consider usable.

andrewharvey:
Sorry I mixed up the waiver template for CC BY and the template text for aerial imagery waivers ([from here](https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates)).

I can open a new PR to make it clear that CC BY imagery can be used with the CC BY waivers or imagery tracing waiver, CC BY/BY-NC/BY-SA can be used with the tracing waiver, and then any other licensed imagery with a general waiver.

> Previously it was that manually tracing from imagery didn't create a derivative of the imagery.

Is the LWG policy document somewhere? Either OSMF requires a waiver or does not require a waiver. Given they provide 3 waivers (CC BY, imagery general, CC imagery) it would imply that we can only use with a waiver.

andrewharvey:
https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates lists

1. [Cover letter and waiver template for CC BY 4.0](https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Cover_letter_and_waiver_template_for_CC_BY_4.0) ([ODT Format](https://drive.google.com/file/d/0B3PN5zfbzThqeTdWR1l3SzJVcTg/view?usp=sharing))
2. [Cover letter and waiver template for CC BY 2.0/3.0](https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Cover_letter_and_waiver_template_for_CC_BY_2.0-3.0) ([ODT Format](https://drive.google.com/file/d/0B3PN5zfbzThqX0NNdFBDejE2RFE/view?usp=sharing))
3. [Template text for aerial imagery waivers](https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers) ([ODT Format](https://drive.google.com/file/d/0B3PN5zfbzThqLXg1TUlxalAtVE0/view?usp=sharing))

with number 3 being further split into

3. a) For an OSM waiver generally
3. b) For a tracing waiver as to imagery already/to be released under CC BY/CC BY-NC/CC BY-SA

My understanding is,

- for CC BY aerial imagery we can accept waiver 1/2, 3a or 3b.
- for CC BY, CC BY-NC or CC BY-SA aerial imagery we can accept waiver 3a or 3b
- for any other licensed or copyrighted imagery we can accept waiver 3a

simonpoole:
> Is the LWG policy document somewhere? Either OSMF requires a waiver or does not require a waiver. Given they provide 3 waivers (CC BY, imagery general, CC imagery) it would imply that we can only use with a waiver.

It is just a case of being safe instead of sorry, conceivably there are imagery providers that are the opinion that they own rights in data traced from their imagery. It should be documented somewhere in the LWG minutes but March 2017 or so.

andrewharvey:
> It is just a case of being safe instead of sorry, conceivably there are imagery providers that are the opinion that they own rights in data traced from their imagery. It should be documented somewhere in the LWG minutes but March 2017 or so.

Understood, but from the DWG point of view I believe the waivers are required to use the data and therefore we must require them to include the imagery as a reference in editors and hence this index.

PR #2600, opened 2025-01-21
update FAQ in regards to waivers per comments in #2598

andrewharvey:
See #2598 for further context. cc @grischard @simonpoole


andrewharvey:
or if this is too complicated I could delete the whole section under "choose one of the waiver templates." and just leave it at that.

PR #2614, opened 2025-02-03
add au/vic/melbourne 2020 imagery and move layers to osm.org hosting

andrewharvey:
pending https://github.com/openstreetmap/chef/pull/734


Firefishy:
Live now https://au-vic-melbourne-imagery.openstreetmap.org/

PR #2617, opened 2025-02-03
Added target="_blank" to JOSM and iD links

### LINKED ISSUE #2611: Add target="_blank" to iD and JOSM urls on website
I frequently click one of the links to load imagery into JOSM, but it's not uncommon that I want to add multiple sources in one go. I often forget to open the link in a new tab (middle mouse button or ctrl+click), so it takes me to the page that says "ok" (https://127.0.0.1:8111). Then when I go back, the map is reset to the whole world.

I think that it should be easy enough to add 'target="_blank"' to the urls in the code to open them in a new tab by default. As far as I can tell, it would be in `site.js`, and would be:
`return 'http://127.0.0.1:8111/imagery?' + (new URLSearchParams(params)).toString() + 'target="_blank"';`
on line 33, and
`return 'https://www.openstreetmap.org/edit?' + position + (new URLSearchParams(params)).toString() + 'target="_blank"';`
on line 47.

Since I don't know if this is right or complete, or desired, I didn't want to do a pull request.

pkoby:
//edit @tyrasd: this closes #2611

### DIFF
--- a/site/site.js
+++ b/site/site.js
@@ -77,8 +77,8 @@ d3.json("imagery.geojson", function(error, imagery) {
             '<h3>Available layers at this location:</h3>'+
             matches.map(function(match) {
                 return match.feature.properties.name +
-                    ` [<a href="${idURL(match.feature, e)}" title="Add to iD">iD</a>] ` +
-                    ` [<a href="${josmURL(match.feature)}" title="Add to JOSM">JOSM</a>]`;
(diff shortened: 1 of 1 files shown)


andrewharvey (approved):
Thanks. I made a minor change per https://github.com/osmlab/editor-layer-index/issues/2611#issuecomment-2649333508

PR #2618, opened 2025-02-06
Add Hungarian FÖMI 2015 leaf-off 20cm orthophotos

Yogurt4:
Added Hungarian FÖMI 2015 leaf-off 20cm orthophotos


sources/europe/hu/Hungary-FOMI-leaf-off-2015.geojson
```diff
@@ -0,0 +1,1341 @@
+{
+    "type": "Feature",
+    "properties": {
+        "name": "FÖMI orthophoto 2015 leaf-off 20cm (webp)",
```
gy-mate, line 4:
```suggestion
        "name": "FÖMI orthophoto 2015 (leaf-off, 20 cm, WEBP)",
```

- Put all parameters between brackets
- Fix capitalization and spacing

gy-mate (changes requested):

andrewharvey:
would you say this is the best imagery for the area? If so we should set `best: true` and this would then invalidate #2521

Yogurt4:
Yes, it is the best (public) imagery for the areas - resolution-wise. However, it's 10 years old (as discussed in #2521). Furthermore, it's a leaf-off ortho, e.g., made in late autum-early winter. That said, the shadows are very long and sharp, often covering smaller buildings. Or, as someone noticed it recently, you see the shadows of the rails instead of the rails themselves, at a considerable offset.

Yogurt4:
See also [PR #2630](https://github.com/osmlab/editor-layer-index/pull/2630) for the country-wide 40 px layer.

PR #2620, opened 2025-02-10
fix wms url of PEMA Orthoimagery Cycle 2 layer

tyrasd:
see https://github.com/openstreetmap/iD/issues/10750#issuecomment-2647077382

//cc @cicku

### DIFF
--- a/sources/north-america/us/pa/PEMA_Orthoimagery_2023.geojson
+++ b/sources/north-america/us/pa/PEMA_Orthoimagery_2023.geojson
@@ -9,7 +9,7 @@
         },
         "name": "PEMA Orthoimagery Cycle 2 (2021-2023)",
         "icon": "https://osmlab.github.io/editor-layer-index/sources/north-america/us/pa/PEMA.svg",
-        "url": "https://apps.pasda.psu.edu/arcgis/rest/services/PEMAImagery2021_2023/MapServer/export?f=image&format=jpg&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&layers=show:3",
+        "url": "https://apps.pasda.psu.edu/arcgis/rest/services/PEMAImagery2021_2023/MapServer/export?f=image&format=jpg&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&layers=show:3&__dummy__proj={proj}",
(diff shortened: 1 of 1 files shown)


tyrasd:
the test complains about a missing `{proj}` property:

> ERROR:root:Error in sources/north-america/us/pa/PEMA_Orthoimagery_2023.geojson : Missing parameter in sources/north-america/us/pa/PEMA_Orthoimagery_2023.geojson: ['{proj}']

Here, the `{wkid}` should be sufficient in principle. Alternatively, we could also use the WMS URL posted by @hlfan: ` https://apps.pasda.psu.edu/arcgis/services/PEMAImagery2021_2023/MapServer/WMSServer?request=getmap&version=1.3.0&service=wms&layers=2&styles=&format=image/jpeg&crs={proj}&width={width}&height={height}&bbox={bbox}`, but I'm not sure which one is the "official"/preferred endpoint for this service.

sources/north-america/us/pa/PEMA_Orthoimagery_2023.geojson
```diff
@@ -9,7 +9,7 @@
         },
         "name": "PEMA Orthoimagery Cycle 2 (2021-2023)",
         "icon": "https://osmlab.github.io/editor-layer-index/sources/north-america/us/pa/PEMA.svg",
-        "url": "https://apps.pasda.psu.edu/arcgis/rest/services/PEMAImagery2021_2023/MapServer/export?f=image&format=jpg&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&foo={proj}",
+        "url": "https://apps.pasda.psu.edu/arcgis/rest/services/PEMAImagery2021_2023/MapServer/export?f=image&format=jpg&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&layers=show:3&__dummy__proj={proj}",
```
hlfan, line 12: Only the layers parameter was needed, renaming foo may make the CI issue too obvious ;)

tyrasd: #2621 should fix the CI warning and the need for the unnecessary confusing dummy parameters

cicku:
PASDA likely changed the URL a bit, thanks for the patch.

PR #2629, opened 2025-02-12
Remove foo={proj} workaround from sources

arch0345:
Closes #2628

Also use {wkid} parameter instead of hardcoding bboxSR and imageSR values

PR #2631, opened 2025-02-12
drop now unnecessary dummy parameters

### LINKED ISSUE #2628: Remove foo={proj} workaround from sources
Following up from #2621 we should now follow through and remove the "hacky workarounds like https://github.com/osmlab/editor-layer-index/commit/f39630af46356ad405502fcdfd8a3adc3ec793a6, […&foo={proj}](https://github.com/search?q=repo%3Aosmlab%2Feditor-layer-index+foo%3D%7Bproj%7D&type=code) in a lot of layers." cc @tyrasd

tyrasd:
Fixes #2628.

There were a couple of sources with hardcoded `…SR=3857`, but as those all have `"available_projections": [ "EPSG:3857" ]` set properly, I replaced these with `{wkid}` and removed the dummy `{proj}`.

Please double-check the `scripts/sources/oceania/au/nsw/NSW_SS_Floods.js` file, as that's the only non-source file that was changed, and I'm not familiar with how this script is used.


tyrasd:
oops, I didn't see #2629 :see_no_evil: :sweat_smile: sorry for the noise

PR #2642, opened 2025-02-21
Rename madrid-ortofotos to madrid-ortofotos.geojson

AlvaroRuizCuevas:
Adding the missing file extension.


arch0345 (dismissed):
~~It looks like the coordinates are in the wrong order, for geojson it should be lon, lat~~

Edit: nevermind, the geometry seems to be correct. Not sure why the CI is erroring out

AlvaroRuizCuevas:
It seems to me that the bounding box is not in the right place. I wonder if it's an SRC issue, but I can't find the solution.
image

Anyway, I manually tested the WMS (https://servpub.madrid.es/georaster/ORTOFOTOS/ORTOFOTO_MADRID/ows?service=WMS&version=1.3.0&request=GetMap&layers=ORTOFOTO_MADRID&bbox={bbox}&width={width}&height={height}&crs={proj}&format=image/png) and it works correctly.

PR #2652, opened 2025-03-10
Hungarian "Balaton" orthophoto is no longer the best one as it got outdated

Yogurt4:
The photos were taken in 2010, we have an all-country imagery from 2015.

Plus some grammar corrections for other Hungarian orthophoto layers.


andrewharvey:
Sorry I merged a newer PR before I saw yours, and it conflicts with your changes. Can you take a look and see if any of your changes here should still be applied?

Yogurt4:
No problem. I've talked with that guy and agreed that he sends a PR for those.

I'm starting to get fed up with this issue though. Due to the nature these photos are created (not covering the entire area) I cannot guarantee strict check will be successful.

PR #2656, opened 2025-03-19
Create ViennaMehrzweckkarte-situation.geojson

hlfan:
Add general purpose map detail/situation overlay from https://stp.wien.gv.at/GeodatenviewerWMS?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities.
A transparent WMS version of https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/at/ViennaBeschriftungen-annotations.geojson for higher zooms.


andrewharvey:
Can you point to the CC BY waiver please?

hlfan:
As stated in the contributors wiki page, it's the same as for basemap.at, https://wiki.openstreetmap.org/w/images/1/1d/Permission_to_Incorporate_basemap.at_Into_OpenStreetMap_.pdf

sources/europe/at/ViennaMehrzweckkarte-situation.geojson
```diff
@@ -0,0 +1,576 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "wien.gv.at-gp-situation",
+        "attribution": {
+            "url": "https://data.wien.gv.at",
+            "text": "Stadt Wien",
+            "required": false
```
andrewharvey: Thanks for confirming the waiver. If the source is CC BY why is attribution not required?

hlfan: The waiver says the attribution on the contributions page is enough

andrewharvey: This field is for attribution the editor must show to attribute the background imagery or overlays users see in the editor, which being CC BY is required.

The waiver specifies how subsequent derived data which is included in OSM data is attributed. Even then attribution is still required, just attribution via the contributors page is enough, but regardless this field isn't about that.

hlfan: Well, I just copied that from the WMTS version ¯\_(ツ)_/¯
https://github.com/osmlab/editor-layer-index/blob/04db2948815708f89f372a78626476c8eab42d26/sources/europe/at/ViennaBeschriftungen-annotations.geojson#L8

andrewharvey: I'm only providing my review of the PR at hand. I think the pre-existing one would also be wrong then. If you don't think my interpretation of these fields is incorrect, best to open a ticket for a wider discussion about the schema. :pray:

hlfan: Since these layers had that from the first version, there was a open data rebrand since then, so I updated the concerned layers

hlfan:
@andrewharvey so is anything else needed?

andrewharvey:
Sorry I missed this.

andrewharvey (approved):

PR #2663, opened 2025-04-01
Added more Saxony WMS, renaming and sorting

vsandre:
Saxonys Aerial Images were a little bit outdatet and not showing all possible WMS.

- sorted and renamed the coresponding files to stand by side
- use the same polygon for all GeoSN-Layers 
- added the latest and most updatet wms-layer "rohdop", more historical aerial images and other usefull layers


vsandre:
The remaining error is caused by renaming of the icon-file to match other "Saxony-*"-Files

andrewharvey:
I realise it's just a rename and the logo was prexisting, but I'm not sure we should be directly including logos that we likely don't have the rights to reproduce, could it just be a URL pointing to the logo from the publisher instead?

vsandre:
Are there any specs the logo must comply with?
Here are logos at GeoSN [middle](https://www.geosn.sachsen.de/img/000%20Cliparts_Logos_Teaser/Logo_GeoSN_Zweitlogo_rdax_465x226_87.png) and [big](https://www.landesvermessung.sachsen.de/img/allgemein/Logo_GeoSN_Zweitlogo_rdax_1240x602_87.png)

vsandre:
Likewise other imagery in germany I changed the layer icon/image to the coat of arms (an deleted the old GeoSN-Version)
@andrewharvey Saxony.png is new an, so the checks have to fail.

vsandre:
Not affected by #2685, please merge my commit

andrewharvey:
I'm not a fan of including the logo in the repo here, it makes the repo larger and it seems a copyright minefield, but it seems like this has been ongoing in other sources so if other maintainers feel we should then that's ok.

Is there an issue with pointing directly to the URL of the logo as hosted by the publisher or a version hosted on Wikimedia Commons?

vsandre:
The other german logos were added with #2496.  These png-files are 100px square and with transparent background. If we link to Wikimedia you get something rectangular 100px width ~~with a white background~~.

wikimedia  squared image

I am open to suggestions or new rules for layer images. But I would like to do this in another bug report/pull request.

andrewharvey:
Thanks for the info. I see per #2496 and https://commons.wikimedia.org/wiki/File:Coat_of_arms_of_Saxony.svg (under Permissions) that this is considered in the public domain. I guess we can include it. It does make the repo larger, and I still think it's probably overall better to have these logos somewhere else, but I don't want to hold this up.

Issue #2636, opened 2025-02-17
KyFromAbove Phase 3


1ec5:
KyFromAbove Phase 3 is 3-inch imagery collected throughout Kentucky between 2022 and 2024. All KyFromAbove imagery is [in the public domain](https://registry.opendata.aws/kyfromabove/). They’ve published imagery for eastern and southern parts of the state and will publish the rest early this year, presumably to the same location.

There’s [a dynamic ImageServer](https://kyraster.ky.gov/arcgis/rest/services/ImageServices/Ky_KYAPED_Phase3_3IN/ImageServer) and [a cached MapServer](https://kygisserver.ky.gov/arcgis/rest/services/WGS84WM_Services/Ky_Imagery_Phase3_3IN_WGS84WM/MapServer).

vsandre:
Created a pull request #2665 using the cached MapServer

vsandre:
Pull request is merged, issue should be closed.

1ec5:
Thank you! In the future, you can use a [closing keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) in your PR description to ensure that the issue gets closed automatically instead of lingering around.

PR #2665, opened 2025-04-03
Create KyFromAbove Phase 3 File

vsandre:
Solves #2636


arch0345 (changes requested):
It looks like this uses the same geometry as phase 2. Although it is true that phase 3 will eventually cover the whole state, as of now it looks like it only covers part of the state. The geometry should be changed to reflect this so that this layer isn't listed in an area where phase 3 imagery isn't available yet

vsandre:
Changed the polygone to reflect the coverage as of today.

arch0345 (approved):

Edit: it looks like there's an error with the layer value, I'll try troubleshooting when I have the time

vsandre:
The image layer is advertised with name "1". So we could simply change it to this value. "2" and "3" are not important for the mapping process.

`<Layer queryable="1">
<Name>1</Name>
<Title>
<![CDATA[ Image ]]>
</Title>`
https://kygisserver.ky.gov/arcgis/services/WGS84WM_Services/Ky_Imagery_Phase3_3IN_WGS84WM/MapServer/WMSServer?VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities

PR #2673, opened 2025-04-10
Add Brest Métropole 2023/2024 WMTS orthophoto

XioNoX:
First contribution, please let me know if there is anything I can improve.
Unfortunately there is only a WMTS feed. Hopefully a WMS feed that can be used in ID once https://github.com/osm-fr/infrastructure/issues/586 is done.

Thanks


sources/europe/fr/Brest-Metropole-Orthophoto2024.geojson
```diff
@@ -0,0 +1,442 @@
...
+        "attribution": {
+            "text": "Brest M\u00e9tropole - 2023-2024",
+            "url": "https://geo.brest-metropole.fr/portal/apps/sites/#/geopaysdebrest/pages/donnees-photographies-aeriennes",
+            "required": true
+        },
+        "max_zoom": 22,
+        "license_url": "https://www.etalab.gouv.fr/licence-ouverte-open-licence",
+        "description": "2023 for roughly the city of Brest, 2024 for the other cities of the area",
+        "best": true,
+        "logo-image": "https://osmlab.github.io/editor-layer-index/sources/europe/fr/Brest_Metropole_logo.png"
```
andrewharvey: it's just `icon` https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#icon

```suggestion
        "icon": "https://osmlab.github.io/editor-layer-index/sources/europe/fr/Brest_Metropole_logo.png"
```

andrewharvey:
The WMTS source only provides EPSG:2154 which limits to editors which can't reproject.

I suggest to use the URL

`https://geo.brest-metropole.fr/arcgis/rest/services/public/PhotoAerienne/MapServer/export?f=image&format=png&transparent=true&imageSR={wkid}&bboxSR={wkid}&bbox={bbox}&size={width},{height}`

Since that will request Web Mercator tiles which allows use in iD and other editors.

andrewharvey:
Sorry I'm not across the licensing, is there a page somewhere documenting this license as permissible to OSM?

XioNoX:
> I suggest to use the URL [...]
Thanks so much !! We spend quite a lot of time figuring out if that was doable !

> Sorry I'm not across the licensing, is there a page somewhere documenting this license as permissible to OSM?

It's the main license used by the french government and local authorities. You can see that most orthophoto in `/sources/europe/fr/` use it.

I was able to find this for example (in french) : https://wiki.openstreetmap.org/wiki/FR:OpenData#Licences_compatibles_avec_celle_d'OpenStreetMap_(ODbL)
That says that at long as you show the paternity (source?) you can use it.

andrewharvey:
> Thanks so much !! We spend quite a lot of time figuring out if that was doable !

The only reason I knew the URL is  I'd previously spent a lot of time working it out for other source.

You might prefer to use format=jpg, since you're not using it as an overlay you don't need the transparency from png.

XioNoX:
I also changed best to `false` as the tiles take a significant amount a time to generate (up to ~25s). Let me know if there is a way to optimize this.

andrewharvey:
Thanks. You can add

```
"available_projections": [
            "EPSG:3857"
],
```

which should fix the check error.

I don't think you can do much about the slowness. In JOSM it may be able to use the service without reprojection, so the JOSM index may want to do that instead of use this export URL.

andrewharvey:
I added the `privacy_policy_url` based on following the links from the URL.

andrewharvey:
I've just merged this one, but thinking about the included logo, I'm not sure it's okay to include it directly like you have.

The imagery index is licensed CC BY SA 3.0 and while we can include other licensed content, for example with `scripts` being under their own licenses, it's complicated to manage that for logo's given they are unlikely to be released under an open license.

How about we instead just use `https://brest.fr/sites/default/files/medias/brestfr/images/accueil/Groupe%2029.svg`?

XioNoX:
Thanks a lot ! Any idea when it will be usable on ID ?

About the logo, I don't mind, but the risk is that this URL becomes a 404 with time.

An alternative could be to use the (outdated) logo from Wikipedia : https://fr.wikipedia.org/wiki/Brest_M%C3%A9tropole#/media/Fichier:Brest_M%C3%A9tropole_logo.svg

andrewharvey:
> Thanks a lot ! Any idea when it will be usable on ID ?

The imagery index in iD is only updated when a new version of iD is released.

> About the logo, I don't mind, but the risk is that this URL becomes a 404 with time.

Yes that happens a lot, we just need to keep updating it.

> An alternative could be to use the (outdated) logo from Wikipedia : https://fr.wikipedia.org/wiki/Brest_M%C3%A9tropole#/media/Fichier:Brest_M%C3%A9tropole_logo.svg

That works.

XioNoX:
Thanks! I leave it to you if it's ok. I don't even know where the logo is being used. I'm fine with all options.

PR #2674, opened 2025-04-11
Remove NY Orthos layer 4, update attribution url

jmapb:
- Remove layer 4 from LAYERS parameter, which is once again causing "parameter layers contains unacceptable layer names" errors in WMS requests (see https://github.com/osmlab/editor-layer-index/pull/1868/files and https://github.com/osmlab/editor-layer-index/pull/2589/files)
- Update imagery attribution url (currently a dead link)

### DIFF
--- a/sources/north-america/us/ny/NYSDOP_Latest.geojson
+++ b/sources/north-america/us/ny/NYSDOP_Latest.geojson
@@ -1363,7 +1363,7 @@
         "attribution": {
             "required": false,
             "text": "New York State Statewide Digital Orthoimagery Program",
-            "url": "http://gis.ny.gov/gateway/mg/index.html"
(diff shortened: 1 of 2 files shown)


ZLima12:
This looks good, but the layers should not be changed for the color IR yet. These layers correspond to the years of imagery, with 0 being the oldest and 3 or 4 being the newest. In this case, they removed the 2020 imagery from the regular, non IR imagery, making 2021 go from layer 1 to 0, and 2024 go from 4 to 3. However, they have not yet changed the color IR, so removing layer 4 would remove the 2024 imagery.

jmapb:
Aha, true. I imagine the IR version will follow suit soon; I'll keep an eye on it.

PR #2657, opened 2025-03-19
Add Canada Ontario 2018-2022 orthoimagery

galen8183:
Add GEospatial Ontario's (GEO's) 2018-2022 orthophotography WMS service from
https://geohub.lio.gov.on.ca/maps/lio::geospatial-ontario-imagery-data-services/about

Licensed under the [OGL Ontario](https://www.ontario.ca/page/open-government-licence-ontario),
a license [approved by the OSMF LWG](https://osmfoundation.org/wiki/OGL_Canada_and_local_variants)

Obsoletes #2342 

Signed-off-by: Galen CC <galen8183@gmail.com>

Issue #790, opened 2020-02-07
documenting licensing


andrewharvey:
It's the job of ELI to:

- Only accept imagery which can be used for tracing into OSM
- Maintain some kind of documentation of (1) so if someone later claims why are you using this imagery for tracing in OSM we can back that up with documentation from here.

(up until today) we had a few fields for recording licensing:

- `license` license of the imagery using SPDX id or "COMMERCIAL"
- `permission_osm` enum of explicit, implicit, no
- `license_url` URL for the license OR permissions for the imagery

I'm proposing an alternative:

- `license` - The name of the license the imagery is generally made available under. If the license has an SPDX identifier, we prefer to use that. If it's a non-standard license you may omit this.
- `license_url` - The URL of the license the imagery is generally made available under. If it's an SPDX license you may omit this.
- `waiver_url` For sources that require an additional waiver or explicit permission to use in OSM beyond the license, the URL to the waiver should be present.

At a minimum you have to supply at least one of these fields.

simonpoole:
At the time I introduced `license ` it was to be able to determine if a source can be used freely outside of the narrow context of use for editing OSM. 

The obvious (hypothetical at this point) use case is creating a (global) mosaic of free aerial imagery sources, which while it wouldn't be perfect, these days it could have significant hi-res coverage. Even if the project is less ambitious than that it would still be useful to quickly determine if there is a freely usable source for a certain area. Given that it is something that a contributor to this repo already has had to find out, it would seem to be reasonable to maintain that information here, instead of forcing this to be redone time and time again and it is very unfortunate that the work already done was thrown away on a whim.

Anyway back to your suggestion, there are three cases that I can see

1. imagery that is licensed on commercial terms, `waiver_url` (perhaps `permission_url` would be more generic) in general the commercial terms are not something we are interested in so no need to link to those, the absence of the other fields would need to imply a commercial / non-open licence.
2. imagery that is licensed on open terms that however require a waiver for use in OSM, `license` needs to be set to the SPDX value, `waiver_url` needs to present (this could actually be validated).
3. imagery that is licensed on open terms that are compatible with OSM,  `license` needs to be set to the SPDX value, and imho `license_url`should be pointing to the location the license for the source is stated (not the license text, as we want to be able to verify that the license is actually the correct one for the source).

andrewharvey:
I agree with this @simonpoole 

[...]

PR #2676, opened 2025-04-17
Add Canada Ontario 2018-2022 orthoimagery

galen8183:
Followup of https://github.com/osmlab/editor-layer-index/pull/2657

Add GEospatial Ontario's (GEO's) 2018-2022 orthophotography WMS service from
https://geohub.lio.gov.on.ca/maps/lio::geospatial-ontario-imagery-data-services/about

Licensed under the [OGL Ontario](https://www.ontario.ca/page/open-government-licence-ontario),
a license [approved by the OSMF LWG](https://osmfoundation.org/wiki/OGL_Canada_and_local_variants)

Regarding the licence attribution concerns in my previous PR (https://github.com/osmlab/editor-layer-index/pull/2657#issuecomment-2764635486),
I'm really not sure what's wrong here... the source licence URL is/was already
linked in the attribution property, and *like many other layers* the license_url
links to the OSMF page on OGL variants. Please let me know if all of those
entries are incorrect, or if linking to a wiki page is in fact correct here.


andrewharvey:
The `license_url` doesn't have consistent use within existing sources and the current CONTRIBUTING guide and schema documentation aren't clear on what exactly this field should contain. In https://github.com/osmlab/editor-layer-index/issues/790#issuecomment-2608743987 I've proposed what I think we should use, but until it's documented I don't think we can expect too much of the field, and this PR uses it to point to documentation of the licensing permissions which is okay for the time being.

If we want the field to point to the publisher's statement on the data license, https://www.ontario.ca/page/copyright-information is probably the page to point to, then https://osmfoundation.org/wiki/OGL_Canada_and_local_variants would go under `permission_url`. But we'd want to get this schema change accepted and documented first (https://github.com/osmlab/editor-layer-index/issues/790#issuecomment-2608743987)

andrewharvey:
What about the 2023-2027 layer? Should that be added to as `best`?

https://ws.geoservices.lrc.gov.on.ca/arcgis5/rest/services/AerialImagery/GEO_Imagery_Data_Service_2023to2027/ImageServer

galen8183:
> What about the 2023-2027 layer? Should that be added to as best?

It's [documented as "coming soon"](https://geohub.lio.gov.on.ca/maps/lio::geospatial-ontario-imagery-data-services/about), I'd be happy to add it now but perhaps it shouldn't be marked as `best` until it's complete. WDYT?

andrewharvey:
> It's [documented as "coming soon"](https://geohub.lio.gov.on.ca/maps/lio::geospatial-ontario-imagery-data-services/about), I'd be happy to add it now but perhaps it shouldn't be marked as best until it's complete. WDYT?

That's fine, can leave out for now and add later when it's ready I think.

artemisquadatro:
Hello all,

I’d like to know if the permission_osm field can be changed from "implicit" to "explicit".

LWG has reviewed Open Government Licence – Ontario and concluded that it is compatible with the ODbL. Please read the assessment below: 

https://osmfoundation.org/wiki/Licensing_Working_Group/Minutes/2024-03-04#Assessment_of_Open_Government_License_-_Ontario

Given this ruling, I believe it is safe to change "implicit" "explicit" to allow it to be used in OSM editors.

andrewharvey:
> I’d like to know if the permission_osm field can be changed from "implicit" to "explicit".

I don't know the history behind that field, it's not mentioned on the contributing documentation at https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md so the only documentation is within [the schema](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json) where it is described as "explicit/implicit permission by the owner for use in OSM".

Is there any reason we need to record this? I understand CC0 would be implicit permission, since the license doesn't mention OSM so we implicitly assume it's suitable for OSM. Other sources that may be copyrighted but the owner has provided direct support for OSM would be explicit as we have explicit permission from the owner for use in OSM.

But what about CC BY + waiver? I suppose that is explicit? But then if we have fields for the license the data is under, and then another field for the justification/further documentation on why this licnese is suitable, seems the permission_osm is not needed.

I've proposed changing how we record the licensing at https://github.com/osmlab/editor-layer-index/pull/2605 but no feedback yet.

> Given this ruling, I believe it is safe to change "implicit" "explicit" to allow it to be used in OSM editors.

I assume that a condition of being accepted into this index is the source can be used to improve OSM, regardless of the implicit/explicit permission.

Am I mistaken?

artemisquadatro:
> I assume that a condition of being accepted into this index is the source can be used to improve OSM, regardless of the implicit/explicit permission.

Yes, any source included in the Editor Layer Index (ELI) is allowed to be used for mapping in OSM. However, layers marked as "implicit" don’t appear in the iD imagery selector.  I'm not too entirely sure of the reasoning behind this, but it seems that only layers tagged as "explicit" are included in the iD imagery lists.

andrewharvey:
> Yes, any source included in the Editor Layer Index (ELI) is allowed to be used for mapping in OSM. However, layers marked as "implicit" don’t appear in the iD imagery selector. I'm not too entirely sure of the reasoning behind this, but it seems that only layers tagged as "explicit" are included in the iD imagery lists.

I couldn't find anything within the iD or ELI code that would do that? Where's the code?

artemisquadatro:
> > Yes, any source included in the Editor Layer Index (ELI) is allowed to be used for mapping in OSM. However, layers marked as "implicit" don’t appear in the iD imagery selector. I'm not too entirely sure of the reasoning behind this, but it seems that only layers tagged as "explicit" are included in the iD imagery lists.
> 
> I couldn't find anything within the iD or ELI code that would do that? Where's the code?

My apologies for the extremely late reply. Couldn't find any code but this layer along with the new NRCan-HRDEM-DTM both marked as "implicit" do not show on the ID layer selector.

andrewharvey:
> My apologies for the extremely late reply. Couldn't find any code but this layer along with the new NRCan-HRDEM-DTM both marked as "implicit" do not show on the ID layer selector.

It's likely that iD hasn't taking in the updates yet, this was merged May 16th but iD released May 12 (so wouldn't have included this change), there was a patch release Jun 4th but it may not have taken in ELI updates.

Currently iD only takes ELI updates during the iD release which is ad-hoc and infrequent.

PR #2678, opened 2025-04-17
upgrade ci runner from 20.04 to 24.04 due to 20.04 being unsupported

andrewharvey:


grischard:
Straightforward enough. We could also just switch to "latest".

PR #2682, opened 2025-04-30
[nz] replace topo50 proxy server with the official basemaps service

k-yle:
replaced https://github.com/Cameron-D/linzcache with https://github.com/linz/basemaps 

I haven't increased the geojson extent yet, we can do that in the future depending on https://github.com/linz/basemaps/issues/3436


andrewharvey (approved):
:+1:

PR #2686, opened 2025-05-06
Create TirisDGM-ReliefShading.geojson

hungerburg:
Recently our local administration GIS started to offer another terrain model with multiple light sources. There are some advantages to it over the other, they complement. Licence as other, I suppose. Cloned from existing WMS end point and changed url.


hungerburg:
Metadata was changed to make it unique, different from the one cloned from. That produced an error in QA.

andrewharvey:
Thanks. Can you remind me where the OSM permission is?

hungerburg:
The OSM permission is at "licence url" https://www.tirol.gv.at/data/nutzungsbedingungen/

It is CC BY 4.0 - The originator wants to be called "Datenquelle: Land Tirol - data.tirol.gv.at", right now, attribution here reads "tiris.tirol.gv.at" which is their front-end for their data, webpage and WMS. I just copied that from the other aerials by same provider.

andrewharvey:
Thanks. CC BY 4.0 is not compatible with OSM, OSMF requests a waiver to be completed at https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

I understand we have existing sources, but we still need to perform due diligence to ensure only compatible sources are included in the index.

hungerburg:
From reading the blog post, I understood this to apply to data imports. Which is not the case for aerials, where in my understanding CC-BY attribution is fulfilled with showing the providers name in the image. But what do I know.

Looking further, I found this waiver, https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers - especially the second paragraph. Just in case tracing from aerial makes a derivative work. Is this available in German?

andrewharvey:
> But what do I know.

I'm in the same boat too. I'm just going by the guidance released by OSMF's LWG.

> From reading the blog post, I understood this to apply to data imports. Which is not the case for aerials, where in my understanding CC-BY attribution is fulfilled with showing the providers name in the image.
> Looking further, I found this waiver, https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers - especially the second paragraph. Just in case tracing from aerial makes a derivative work. Is this available in German?

I think based on what the LWG has provided here we do need one of these waivers, or an equivalent permission letter.

hungerburg:
As I am not that much of an expert in such matters, I informed the Local Chapter more than a month ago. Still waiting for a reply. On the local forum I got a recommendation to use a waiver that does not apply. I still do not see why imagery backgrounds available on same terms as existing ones cannot be added, while those that are already in can remain?

andrewharvey:
> On the local forum I got a recommendation to use a waiver that does not apply.

For what it's worth, there's a lot of CC BY sources in the index that relies on the CC BY waivers (despite them being written for data). In AU/NZ where I work we try to obtain a blanket CC BY wavier to cover all CC BY open data without consideration if that's vector data or raster data.

> I still do not see why imagery backgrounds available on same terms as existing ones cannot be added, while those that are already in can remain?

If other CC BY sources don't have a waiver or waiver equivalent permission statement from the publisher we should remove them, and raise the issue with DWG about any data entered into OSM relying on these sources.

We could also write to LWG for further guidance.

hungerburg:
The deed is linked from https://www.openstreetmap.org/copyright and still available https://www.tirol.gv.at/data/nutzungsbedingungen -- The orthophoto (corresponding to the terrain model in this issue) was included with commit https://github.com/osmlab/editor-layer-index/commit/530dd6391d256fed8ae0b92ded56be2b336bccc4 in 2017. Likely the deed was 4.0 then already, as 4.0 came out in 2013, but might have been 3.0 or 2.0 even, if that matters.

I guess the standards are higher now. The consequences you mention truly horrific. This affects more sources. This one here at least gives attribution proper.

hlfan:
Well the Tiris layers have been licensed under CC-BY **with extensions** that allow OSM usage:
https://github.com/openstreetmap/trac-tickets/issues/4401

hungerburg:
@hlfan Can you reproduce the contents of "the extensions"? For me there is only a dead link there. As far as I can tell from what I learn from what is still available, the extensions are about attribution, but attribution is not what stalls this pull request.

hungerburg:
Some time passed, I think the extensions not up to scrutiny. @andrewharvey in for a big killing spree?

andrewharvey:
> Tiris layers have been licensed under CC-BY with extensions that allow OSM usage

Where is this stated? I run into the same issue as @hungerburg at https://github.com/osmlab/editor-layer-index/pull/2686#issuecomment-5222431648 that http://www.tirol.gv.at/applikationen/e-government/data/nutzungsbedingungen/ leads to a 404 error.

When I visit https://www.tirol.gv.at/data/nutzungsbedingungen/ I get the following (translated into English)

> Terms of Use
> License
> Open Government Data of the State of Tyrol is licensed under a license of[Creative Commons Attribution 4.0](https://creativecommons.org/licenses/by/4.0/deed.de)International. Please note the license extensions! 
> 
> License extensions
> The State of Tyrol must be credited as the rights holder in the following manner:
> "Data source: State of Tyrol - data.tirol.gv.at"
> The public should be informed about new applications and services created using data from the Open Government Data Portal of the State of Tyrol. The State of Tyrol is therefore authorized to publish information about such applications and services, while adhering to data protection regulations, and to use this information for reporting purposes. Furthermore, the State of Tyrol is permitted to publish (link to) these applications and services on its website "data.tirol.gv.at".
> The licensee who uses data from the Open Government Portal of the State of Tyrol for their applications or services agrees to actively support the State of Tyrol under[opendata@tirol.gv.at](mailto:opendata@tirol.gv.at)to inform you where these applications or services can be found.
> The data from the Open Government Data Portal of the State of Tyrol may not be used for applications or publications that support or incite criminal, illegal, racist, discriminatory, defamatory or pornographic activities.

These license extensions state the specific attribution text they would like but the issue with the CC BY license (and why LWG asks for a waiver on top of CC BY) is because downstream users of OSM just attribute OSM in their usage, they don't attribute all the contributors to OSM such as the State of Tyrol, and LWG has determined that this may be problematic, hence in the CC BY waiver they ask data contributors who publish under CC BY that they agree that OSM attributes State of Tyrol (previously on /copyright but now on the OSM wiki Contributors page) and downstream users of OSM only attribute to OSM and not State of Tyrol.

To be honest these extensions make the situation worse for OSM not better, take point 3

> The licensee who uses data from the Open Government Portal of the State of Tyrol for their applications or services agrees to actively support the State of Tyrol under[opendata@tirol.gv.at](mailto:opendata@tirol.gv.at)to inform you where these applications or services can be found.

If you integrate State of Tyrol data into OSM, downstream users of OSM do so under the ODbL and are not expected to report back to State of Tyrol how they are using OSM data which includes State of Tyrol data...

Therefore I can't see how State of Tyrol data under that license can be permitted to be used for OSM without the LWG's CC BY waiver.

hungerburg:
@andrewharvey Coincidentally (or not) last Saturday late Afternoon (Aug 22 17:00) the "tiris" aerials where removed by somebody calling themselves "tiris" from the JOSM background layers listing. C.f. https://josm.openstreetmap.de/wiki/Maps/Austria?version=101 - given the explanation that they are CC-BY without waiver.

Did this index here receive requests for similar action on your side?

andrewharvey:
> Did this index here receive requests for similar action on your side?

Any request would be visible via an Issue or PR, I'm not aware of anything.

> given the explanation that they are CC-BY without waiver.

If they are CC BY and there's no waiver or waiver equivalent, then yes they should be removed per LWG's position.

PR #2696, opened 2025-05-26
Add High Resolution Digital Elevation Model Mosaic dataset from NRCan

### LINKED ISSUE #2695: Add NRCan HRDEM Mosaic
Natural Resources Canada has several Digital Elevation Model products derived from LIDAR data.

The layer that is probably the most of interest is the layer named `dsm-hillshade`

It is published under the OGL Canada license, which has been deemed compatible by the LWG

Open data portal entry: https://open.canada.ca/data/en/dataset/0fe65119-e96e-4a57-8bfe-9d9245fba06b

WMS link: https://datacube.services.geo.ca/ows/elevation?service=wms&request=GetCapabilities

wolfy1339:
Closes #2695


matheusgomesms:
Which one is the intended layer?

`dsm-hillshade` (https://datacube.services.geo.ca/wrapper/ogc/elevation-hrdem-mosaic?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=dsm-hillshade&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox})
Captura de Tela 2025-05-26 às 21 23 50

`dtm-hillshade` (https://datacube.services.geo.ca/wrapper/ogc/elevation-hrdem-mosaic?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=dtm-hillshade&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox})
Captura de Tela 2025-05-26 às 21 24 09

wolfy1339:
Both layers are useful, in their own right.

Judging from the USGS3DEP, it would be `dtm-hillshade` that would be default layer

matheusgomesms:
If both are important, both can be added :)

Feel free to create another PR with the other layer, please fix the URL (with the ones I provided), and if possible please improve the country geometry. The current bbox spans half of USA and entire Greenland, so users there will have that layer shown to them (although no data will be provided).

matheusgomesms:
(And let's wait for input from others maintainers here, they have more experience than me)

grischard:
Both can indeed be useful, dtm to map paths under the trees, dsm to map buildings...

In Luxembourg, I had access to the raw point cloud and created a "mapper's delight" version with the buildings but not the trees - but that's a bit of a hassle, and not really necessary if there's a good enough product already.

grischard:
Looks good to me, OK for me to merge when the checks pass

matheusgomesms:
Strict check didn't pass: ERROR:root:18.71% of geometry is outside of the layers bounding box. Geometry should be checked

However, I checked the layer and the provided geometry, and looks good to me. Shall we merge it?

(thank you for your PR, by the way!)

grischard:
The geometry looks right but can the bbox be updated? Good to merge otherwise.

wolfy1339:
The script doesn't seem to take into account the BBOX from the file, and instead gets it from the WMS server.

The BBOX defined on the WMS server is wrong, since there is data up north.

I have updated the Polygon to fit into the BBOX advertised by the server

matheusgomesms:
Wonderful, thank you for your patience and contribution!

PR #2697, opened 2025-05-31
add Singapore OneMap Orthophoto

k-yle:
We already have 2 layers from Singapore OneMap, and now https://onemap.gov.sg has an aerial imagery layer too (from the home page, click <kbd>☰</kbd> → <kbd>Select Map Styles</kbd> → <kbd>Orthophoto</kbd>)


pietervdvn:
I've corrected the category in https://github.com/osmlab/editor-layer-index/pull/2721

k-yle:
@pietervdvn thanks! you could probably also add `best` if you wanted, it's significantly higher quality than the default layers

PR #2698, opened 2025-05-31
Czechia: Update to WMS which provides up-to-date imagery (2025)

martin-kokos:
The previous service serves imagery dataset frozen in 2022.


andrewharvey:
> The previous service serves imagery dataset frozen in 2022.

Might be useful to retain the old one as another entry, but I'll merge this in the meantime.

PR #2701, opened 2025-06-01
Extend French orthophoto layer to New Caledonia

zabop:
Let's have the high resolution French orthophoto layer available in New Caledonia!


andrewharvey:
Thanks.

PR #2702, opened 2025-06-01
.json -> .geojson switch, so Slovak DSM shows up in iD

zabop:
As far as I understand the situation, only those sources which are specified in files with .geojson extension show up in iD.
iD uses imagery.geojson, and to produce that, [this line](https://github.com/osmlab/editor-layer-index/blob/6b97518d38be8f8732973d4423737b5c95d1532f/Makefile#L2) selects only the .geojson files, not the .json ones.

I hope this commit will make _DSM 1.0 shading_ available in iD in Slovakia.


andrewharvey:
Ah yes, good find!

PR #2706, opened 2025-06-04
Ortophoto Madrid metadata updated

Crashillo:
Since the url is pointing to the _Ortofoto actualizada_ (most up-to-date ortophoto), it becomes to the best choice. Besides the years restrictions no longer apply.


matheusgomesms:
Gracias!

PR #2715, opened 2025-06-10
Added Fortaleza Base Maps

matheusgomesms:
Added kinda new base maps from the Municipality of Fortaleza (it is a combination of many layers they have, very useful), and added the icon for all Fortaleza layers.

Same sources as before, same everything, just a new layer.


matheusgomesms:
Check failed because the icon is not in the repository (yet) :P

PR #2723, opened 2025-06-14
Add new 2025 aerial imagery for France

Nekzuris:
New IGN Ortho Express 2025 🚀
I did the geometry according to this plan, only a small portion is currently published, the rest will arrive in the coming months.
PVA Ortho 2025
You can follow the survey plane here [F-GALP](https://www.flightradar24.com/data/aircraft/f-galp)


andrewharvey:
Thanks looks good. Could you please remind me where the documentation is for https://www.etalab.gouv.fr/licence-ouverte-open-licence/ being compatible for use with OSM?

cquest:
The "Open License / Licence Ouverte" is like CC-BY, and compatible with OSM.

We have tons of opendata in France used daily by contributors under this license... like past years edition of this aerial imagery.

andrewharvey:
The job of the maintainer here includes vetting the sources to ensure they are compatible. While I've proposed changing the schema to include this at https://github.com/osmlab/editor-layer-index/pull/2605 so it can be assessed from the source json alone, currently sources only include is the license the data/imagery is under, but nothing about why that license is compatible so it needs to be supplied with the PR.

> The "Open License / Licence Ouverte" is like CC-BY, and compatible with OSM.

CC BY is not compatible with OSM, it needs a separate waiver https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/.

> We have tons of opendata in France used daily by contributors under this license... like past years edition of this aerial imagery.

Can you please help me out and point me to the documentation from LWG or others saying that it's compatible. I tried looking through the PRs from prior sources but couldn't find this.

Nekzuris:
I'm not exactly sure what you're looking for, but IGN clearly states that OSM can use their imagery in previous agreements:
https://www.openstreetmap.fr/bdortho/
https://www.openstreetmap.fr/une-nouvelle-convention-pour-un-partenariat-avec-lign/

This was later changed to be more open (not limited to only OSM):
https://fr.wikipedia.org/wiki/Licence_ouverte_(%C3%89tat_fran%C3%A7ais)

PR #2727, opened 2025-06-16
Remove MML-sources. Their response indicates 'offline due to limited HR'

pietervdvn:
They indicate: `Resource is no longer available. Please use MML's public service for the moment as we don't have enough human resources for maintaining this properly.  https://www.maanmittauslaitos.fi/kartat-ja-paikkatieto/aineistot-ja-rajapinnat/karttojen-rajapintapalvelut/karttakuvapalvelu-wms-0`

Source: https://kapsi-tiles.y2k.fi/tiles/taustakartta/webmercator/19/299070/137223.png


todeveni:
Instead of removing, switch to WMS, which seems to work. #2753

grischard:
Fixed in #2753 instead. Thank you both!

PR #2728, opened 2025-06-16
Remove USDA-NAIP

pietervdvn:
I tested [USDA-NAIP](https://gis.apfo.usda.gov/arcgis/rest/services/NAIP/USDA_CONUS_PRIME/ImageServer/tile/20/408248/245032) and [USDA-NAIP-PR](https://gis.apfo.usda.gov/arcgis/rest/services/NonCONUS/PR/ImageServer/tile/20/470600/331523), both give a *not found*.

I do think that these imagery sources have gone offline due to the US' administration taking down all scientific items related to climate change.

(In case you are wondering: I've created a small script testing all the ELI layers and am cleaning up the low hanging fruit)


andrewharvey:
> (In case you are wondering: I've created a small script testing all the ELI layers and am cleaning up the low hanging fruit)

Thanks, and thanks for researching the status of each further, because sometimes the service may have moved elsewhere, or might be down temporarily.

I'm not sure we we should keep these around somewhere in case they come back in the future (in a place that's easier that searching through the git history), or just don't worry about it.

arch0345:
These layers are still working for me, it just looks like they changed the highest zoom level to be 18/19 from 20

PR #2736, opened 2025-06-23
Hamburg: Added ALKIS

G4rdener:
We had a reread of the permission from 2021 of the HH LGV (https://wiki.openstreetmap.org/wiki/DE:Permissions/LGV_Hamburg_2021) and came to the conclusion that we can use the ALKIS too.


matheusgomesms:
Danke!

Issue #2625, opened 2025-02-12
Deploy an iD preview site as a PR workflow


andrewharvey:
https://github.com/openstreetmap/id-tagging-schema has a neat feature where each PR will built a preview site of iD incorporating the PR changes.

While iD is not the only consumer of ELI, it would still be nice to have this feature here, so that we can test how the imagery will look and work in iD before merging it.

cc @tyrasd who built this feature for id-tagging-chema

tyrasd:
> cc @tyrasd who built this feature for id-tagging-chema

Sorry, I missed your question. I did build the auto-deploy feature for iD and the tagging schema repo. The code is in the two `build` [workflows](https://github.com/openstreetmap/id-tagging-schema/tree/main/.github/workflows)[^1]. It would be possible to use that for this repository as well. Let me know if I should help with that! :bow:


[^1]: this is a bit more convoluted than the normal staging build workflow (see e.g. https://github.com/openstreetmap/iD/pull/10768) because in a PR context, one does not have access to github secrets which are necessary to deploy the build results somewhere like on netlify.

arch0345:
This would be super useful! I'd appreciate if you could help with setting this up @tyrasd

k-yle:
if https://github.com/openstreetmap/iD/pull/12152 goes ahead, it will be a bit easier to implement this, since the imagery URL can just be swapped out dynamically

PR #2739, opened 2025-06-25
Recommend testing URL in iD

orblivion:
https://rapideditor.github.io/imagery-index seems to be broken and it hasn't been updated since 2023.

I get a mapbox token error in the console. I would think that I could put in a different geojson and click "View it!" but nothing seems to happen when I do.

I figured we could just suggest people test it in iD, but I'm new to this corner of OSM so maybe that's misguided.


andrewharvey (approved):
Thanks. Long term #2625 would go a long way to help with testing, but until then I think manually testing via Custom in iD is the best option.

PR #2746, opened 2025-07-01
Spain: Add PNOA provisional

Crashillo:
Include the provisional orthoimages layer of the Spain region

[GetCapabilities](https://wms-pnoa.idee.es/pnoa-provisionales?request=GetCapabilities&service=WMS&version=1.3.0&)


Crashillo:
This failing check is an _HTTP too many request_ error given by openstreetmap.org. Other spanish layers use the same request. Could you just ignore or disable?

`ERROR:root:sources/europe/es/PNOARapida.geojson: license url https://wiki.openstreetmap.org/w/index.php?title=ES:Fuentes_de_datos_potenciales_de_Espa%C3%B1a/Servidores_WMS is not reachable: HTTP code: 429`

andrewharvey:
Which source is this from https://wiki.openstreetmap.org/w/index.php?title=ES:Fuentes_de_datos_potenciales_de_Espa%C3%B1a/Servidores_WMS ?

Crashillo:
The first one, PNOA. This is another endpoint of the same service, with the most updated imagery, but partial cover.

andrewharvey:
Ok so the permission is https://wiki.openstreetmap.org/wiki/File:OPENSTREETMAPCartaAbiertaIGN-espa%C3%B1ol.pdf

PR #2747, opened 2025-07-01
Spain (Andalusia): Add CDAU map

Crashillo:
Include the Callejero Digital Andaluz Unificado (~ Andalusian streets map) layer

[GetCapabilities](http://www.callejerodeandalucia.es/servicios/cdau/wms?service=wms&request=getcapabilities)


Crashillo:
This failing check is an _HTTP too many request_ error given by openstreetmap.org. Other spanish layers use the same request. Could you just ignore or disable?

`ERROR:root:sources/europe/es/PNOARapida.geojson: license url https://wiki.openstreetmap.org/w/index.php?title=ES:Fuentes_de_datos_potenciales_de_Espa%C3%B1a/Servidores_WMS is not reachable: HTTP code: 429`

andrewharvey:
Could this be added to https://wiki.openstreetmap.org/w/index.php?title=ES:Fuentes_de_datos_potenciales_de_Espa%C3%B1a/Servidores_WMS ?

Crashillo:
Well, it is already there. It has no section itself because the service belongs to IDEAndalucia, where you can find, at the comments column, a mention to CDAU wms.

andrewharvey:
Ok, so the permission is https://wiki.openstreetmap.org/wiki/ES:Callejero_Digital_de_Andaluc%C3%ADa_Unificado

PR #2748, opened 2025-07-01
Replace spain geojson

### LINKED ISSUE #2732: Spanish background layers are not available over Ceuta and Melilla
As per openstreetmap/iD#8510 cc.@jcgf-osm

Ceuta and Melilla missing from geojson files for PNOA and Catastro: https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/europe/es

Thanks,

Crashillo:
Fixes #2732

Uses a different geojson (from the geofabrik's .poly file) for those layers who spread all over Spain surface: PNOA, IGN Raster map, and Cadastre


grischard:
Thank you!

The CI checks were on a wiki server error, so I've merged this.

PR #2753, opened 2025-07-09
Switch from TMS to WMS for MML background maps

todeveni:
Kapsi tile map service has ceased to operate couple of months ago. WMS is still working, so switch finnish NLS background maps to use that instead.

Tested with iD editor using each URL as custom background.


grischard:
Merging this. It fails the CI checks on a lack of privacy policy, but there was none before either, and is "just" a URL change.

PR #2762, opened 2025-07-19
Add Köln TrueDOP 2024

ginnyTheCat:
This adds [TrueDOP 7.5cm resolution aerial imagery](https://open.nrw/dataset/6ba3c12b-8dca-4a0b-a614-f7a71662a846) from the city of Cologne, Germany taken in June 2024. It's licensed under the [Data licence Germany - Zero - Version 2.0](https://www.govdata.de/dl-de/zero-2-0) and should therefore be compatible with OSM. I'm struggled to find a good privacy policy URL so I'm unsure about that, and whether it should be labeled "best". "NRW 10cm" (which in Cologne is from 2023 I think) is already quite good so this is not the biggest improvement, but still.

Comparison between Köln 2024:
Intersections seen in "Köln 2024"

and NRW 10cm
same intersection in "NRW 10cm"

Again Köln 2024 (notice the new bike infrastructure):
Intersection seen in "Köln 2024" with new bike infrastructure visible

and NWR 10cm
same intersection in "NRW 10cm" without the bike infrastructure


grischard:
Ah, it doesn't like the ö in the id field. The software should probably be fine with it, but can we have Koeln instead?

PR #2765, opened 2025-07-27
Added public layers for LIDAR HD in France

StC-OSM:
These two layers are published under the same license terms as the other IGN layers already available in ELI; they provide two surface models computed from LIDAR HD data over Metropolitan France.


sources/europe/fr/IGN-MNS-LIDAR-HD.geojson
```diff
@@ -0,0 +1,2288 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "fr.ign.mnslidarhd",
+        "name": "IGN MNS LIDAR HD",
+        "description": "Modèle numérique de terrain issu de la campagne LIDAR HD de l'IGN.",
+        "type": "tms",
+        "url": "https://data.geopf.fr/tms/1.0.0/IGNF_LIDAR-HD_MNS_ELEVATION.ELEVATIONGRIDCOVERAGE.SHADOW/{zoom}/{x}/{y}.png",
+        "best": true,
+        "category": "photo",
```
andrewharvey:
```suggestion
        "category": "elevation",
```
elevation?

https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L61-L74

sources/europe/fr/IGN-MNS-LIDAR-HD.geojson
```diff
@@ -0,0 +1,2288 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "fr.ign.mnslidarhd",
+        "name": "IGN MNS LIDAR HD",
+        "description": "Modèle numérique de terrain issu de la campagne LIDAR HD de l'IGN.",
+        "type": "tms",
+        "url": "https://data.geopf.fr/tms/1.0.0/IGNF_LIDAR-HD_MNS_ELEVATION.ELEVATIONGRIDCOVERAGE.SHADOW/{zoom}/{x}/{y}.png",
+        "best": true,
```
andrewharvey: "best" should be reserved for aerial/satellite imagery, not an elevation model

sources/europe/fr/IGN-MNS-LIDAR-HD.geojson
StC-OSM: Fixed now.

sources/europe/fr/IGN-MNS-LIDAR-HD.geojson
StC-OSM: Fixed now

andrewharvey (approved):

sources/europe/fr/IGN-MNS-LIDAR-HD.geojson
Nekzuris: Why @andrewharvey  ? The precision of LiDAR data is better than any aerial or satellite imagery for drawing roads, paths, buildings etc...
And it's not like we have dozen of competing best layer, this should be on top imo.

andrewharvey: I hadn't really thought about LIDAR as being the default. If that's what the local mappers think would work best as the default imagery shown, it's fine with me.

`best` is always going to be subjective, for example low res but recent vs high res but old it's going to depend on if at the moment you're trying to trace features with accuracy, or you're just looking for recent changes.

I think as a general default, especially for new mappers who aren't GIS people and who open iD for the first time to add their local pub or add the address for their home or business, having imagery shown is probably better?

Nekzuris: Yeah I agree, photography should stay the default. I don't think `best` means default, I thought it was just to mark the layer with a star and move it to the top of the layer list. There are already places with multiple `best` layers (like over Strasbourg), and I don't think it's an issue, though I don’t know how the default is selected then.

andrewharvey: In iD,

> `best` - If set to `true`, ... considered "better than Bing" and may be chosen by default when iD starts...

https://github.com/openstreetmap/iD/blob/develop/API.md

With the default selection happening at https://github.com/openstreetmap/iD/blob/c888cba4984aff49ef4a055e930d52d382a4e126/modules/renderer/background.js#L467-L495

I thought the intent behind `best` was to indicate the best default.

grischard: Yeah, 'best' is a long outgrown attribute and subjective at, uh, best ;). When there's more than one best, behaviour is undefined. Some editors pick the smallest footprint, others the most recent date.

Ideally our QA tool would have flagged this.

Nekzuris: Maybe we should add some logic to not pick `"category": "elevation"` as default.
Or a specific key `"default": true`
I want to mark this layer as `best` because it's really the best ortho-corrected layer we have and the only one for tracing precise paths and rivers in forests.

grischard: That's beyond the scope of this repo, really. The lidar layers are very useful, but 'best' is for aerial imagery for now. If iD's maintainers wanted to change this, I'd be in favour of it.

andrewharvey: discussion around the "best" concept -> https://github.com/osmlab/editor-layer-index/issues/130

Issue #2760, opened 2025-07-16
Update OS OpenMap Local (Great Britain)


nc011:
OS OpenMap Local (April 2024) is already a layer. The data source for this layer is updated every 6 months (October and April), meaning there is now an April 2025 layer available.

## URL of the imagery
https://www.ordnancesurvey.co.uk/products/os-open-map-local
https://osdatahub.os.uk/downloads/open/OpenMapLocal

## Two letter country code
GB

## Desired layer name
OS OpenMap Local (April 2025)

## License
Same as current (2024) layer
[OGL 3.0](https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/)

## Is the license compatible with OSM?
Same as current (2024) layer
Yes

## Attribution (text, URL, required?)
Same as current (2024) layer. Presumably: "Contains public sector information licensed under the Open Government Licence v3.0."

## Privacy policy (URL)
Same as current (2024) layer

## Icon
Same as current (2024) layer

## Bounding polygon
Same as current (2024) layer

As an aside, is there any way to automate this update seeing as there is a new release every 6 months from the same location?

andrewharvey:
> As an aside, is there any way to automate this update seeing as there is a new release every 6 months from the same location?

I think it's still best to do this manually for the time being.

Here was the Oct 2024 PR https://github.com/osmlab/editor-layer-index/pull/2575

Looks like https://github.com/openstreetmap/chef/blob/master/cookbooks/imagery/recipes/gb_os_sv.rb isn't yet updated, and is a dependency for the update here to the index 

cc @Firefishy

Firefishy:
The most recent version on server is always available using the alias urls. eg: https://os.openstreetmap.org/om-local/5/15/10.png

andrewharvey:
> The most recent version on server is always available using the alias urls. eg: https://os.openstreetmap.org/om-local/5/15/10.png

Perfect, in that case we should probably swap to using that URL in the index. Then we can consider if we want additional entries for each of the historical versions or not. I'll leave it for UK mappers to decide and open a PR.

nc011:
Just to be certain, would that mean swapping

``"url": "https://{switch:a,b,c}.os.openstreetmap.org/layer/gb_os_om_local_2024_10/{zoom}/{x}/{y}.png"``

to 

``"url": "https://{switch:a,b,c}.os.openstreetmap.org/om-local/{zoom}/{x}/{y}.png"``?

[...]

PR #2767, opened 2025-08-01
Openmap local latest

nc011:
Change GB's OS OpenMap Local to always point to the latest version on the osm servers.

See https://github.com/osmlab/editor-layer-index/issues/2760


andrewharvey (approved):

sources/europe/gb/OSOpenMapLocal.geojson
```diff
@@ -1,12 +1,11 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "OS-OpenMap_Local_2024_10",
-        "name": "OS OpenMap Local (October 2024)",
-        "url": "https://{switch:a,b,c}.os.openstreetmap.org/layer/gb_os_om_local_2024_10/{zoom}/{x}/{y}.png",
+        "id": "OS-OpenMap_Local_latest",
+        "name": "OS OpenMap Local",
+        "url": "https://{switch:a,b,c}.os.openstreetmap.org/om-local/{zoom}/{x}/{y}.png",
```
andrewharvey:
```suggestion
        "url": "https://os.openstreetmap.org/om-local/{zoom}/{x}/{y}.png",
```
In the past sub domains offered enhanced performance but this is no longer true, so let's remove them here (ref https://operations.osmfoundation.org/policies/tiles/)

PR #2768, opened 2025-08-01
Improve quality of Henrico County, VA 2024 aerial layer

jacobwhall:
- Adjust URL to query an endpoint with better processing at high zooms
- Request up to zoom 21 to fully exhaust source image quality

This new URL has two layer GET parameters, which refer to the same layer from the source. The first is a caching layer which speeds up response times in my experience.

Below are before and after screenshots for these changes in iD. The change in colors is from the source, when requesting imagery equivalent to zoom 21. At lower zooms (z<=20), the colors are identical because it is the same source imagery.

Screenshot From 2025-07-31 22-56-50
Screenshot From 2025-07-31 22-56-40

### DIFF
--- a/sources/north-america/us/va/Henrico_VA_2024.geojson
+++ b/sources/north-america/us/va/Henrico_VA_2024.geojson
@@ -1,7 +1,7 @@
 {
   "type": "Feature",
   "properties": {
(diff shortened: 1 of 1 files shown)


jacobwhall:
Related to #2632

andrewharvey (approved):

PR #2774, opened 2025-08-04
Added older Ortofotomozaika SR maps and updated the latest one

zdila:
We allowed following sites to use the maps:

```nginx
if ($http_origin ~* "^https?://(.*\.freemap\.sk(:\d+)?|mapcomplete\.org|every-door\.app|www\.systemed\.net(:\d+)?|preview\.ideditor\.com(:\d+)?|ideditor\.openstreetmap\.org(:\d+)?)$") {
    add_header Access-Control-Allow-Origin "$http_origin";
}
```


andrewharvey:
All 3 are marked `best: true` should only the "best" be marked?

PR #2778, opened 2025-08-05
Adjust end dates

simonpoole:
Fixed start/end dates don't make really make sense for these products.

All three products are continuously updated, the terrain ones in a six year cycle the orthofoto faster. Making the start and end dates quite meaningless outside of a rough indication from when the data is.

Maybe it would be a better idea to allow specification of an "update_cycle" value, which would allow the user at least to determine the maximum age of the imagery they are using.


andrewharvey:
> Maybe it would be a better idea to allow specification of an "update_cycle" value, which would allow the user at least to determine the maximum age of the imagery they are using.

I think that makes sense. Though I've seen imagery on an update cycle getting stale too, URLs break, or the updates just don't appear on schedule, so perhaps just for imagery that seems to reliably keep to a cadence we could add that.

andrewharvey (approved):
The end date of 2030 seems a bit strange but if you think this is best, all good.

PR #2779, opened 2025-08-06
Rename entry to correct name Baden-Württemberg

joto:


joto:
Check fails because the image linked is not there, but it can not succeed until the PR is merged...

Issue #2738, opened 2025-06-25
Update remaining NH GRANIT URLs


orblivion:
NH GRANIT changed its URL scheme. This has already been addressed for one of the four NH GRANIT maps:

#2689

The other three should be updated as well:

https://github.com/osmlab/editor-layer-index/tree/master/sources/north-america/us/nh

Thanks!

orblivion:
cc @ThatAverageProcast in case you still wanted to do this

ThatAverageProcast:
Il do 2 of them, and let one for you t do so you can get some experience in, feel free to ask any questions if you get stuck at all or want to change the arragenment

orblivion:
Oh, okay! I mean it seems very easy, I could just as well do all of them. I just have to figure out how to test it.

orblivion:
Like, do I need to spin up a whole instance of iD?

ThatAverageProcast:
for testing I used go map and made a new user defined imagery and put in the new url

orblivion:
In CONTRIBUTING.md I found: https://rapideditor.github.io/imagery-index/index.html Would that suffice?

ThatAverageProcast:
That seems to be another good way of testing, if it works there then id proceed with making a pull request

ThatAverageProcast:
Hm, when doing the infrared imagery (CIR), it spits out the error objectForData(). Odd as the only thing that I’m aware of that changed is nhgeodata being changed to image. When putting the URL into Safari, it says invalid zoom levels. Will have to investigate more.

orblivion:
Also the rapideditor imagery-index seems to be broken even for existing sources. But I can add it as a custom background in iD (I'm gonna make a PR to suggest that instead). I can confirm that the one you already changed works, but the other three with the `images` in the URL don't work.

orblivion:
See #2782 for the CIR. For the hillside ones, someone (a very prolific editor) said that the "Multi" hillside works best right now:

`https://nhgeodata.unh.edu/image/services/ImageServices/LiDAR_Bare_Earth_multi_HS_NH_2022/ImageServer/WMSServer?FORMAT=image/png&TRANSPARENT=TRUE&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap&LAYERS=LiDAR_Bare_Earth_multi_HS_NH_2022:Multi Hillshade&STYLES=&CRS={proj}&WIDTH={width}&HEIGHT={height}&BBOX={bbox}`

[...]

PR #2782, opened 2025-08-10
Updating source URLs for north-america/us/nh

orblivion:
Updating the following to reflect a change in the service:

* NH_2021_2022_6in_CIR.geojson

`NH_2021_2022_6in_RGB.geojson` was updated in a previous commit. The other two still need to be updated.

NOTE:

Strangely, there is one difference between `NH_2021_2022_6in_CIR.geojson` and `NH_2021_2022_6in_RGB.geojson`, which is that the latter has a `:None` in the URL (Same place that the old URL scheme had a `:2122CIR` which turns out needed to be removed). If I remove the `:None` from the RGB, it seems to behave the same, so maybe we should just do that. But if I add `:None` to the CIR, it becomes RGB.

BTW credit goes to someone in the OSMUS Slack, I didn't find this URL myself.


orblivion:
I would be curious what @ThatAverageProcast would say about those weird nuances in the URL since we were discussing this on #2738.

PR #2796, opened 2025-09-11
[BG] Add MAF orthophoto

Dimitar5555:
This imagery covers the entire territory of Bulgaria and was provided by the Ministry of Agriculture and Foods. There are no restrictions on its use.

About one-quarter of the imagery is updated each year, so the full dataset isn’t from a single year. To make this clearer, I’ve added a `vintage_url` parameter that points to a GeoJSON file showing when each section was captured. This parameter will be used in iD, and it can be renamed if needed. I will add it to the schema, once we confirm it's final name.

The final vintage URL will be confirmed soon, but for now it is visible  here:
https://gist.github.com/Dimitar5555/9ff7ee499c93148b4b4b5690889eebe3#file-maf-vintage-geojson


sources/europe/bg/maf-orthophoto.geojson
```diff
@@ -0,0 +1,311 @@
...
+    "i18n": true,
+    "type": "tms",
+    "url": "https://bg-imagery.openstreetmap.org/layer/maf-orthophoto-latest/{z}/{x}/{y}.png",
+    "category": "photo",
+    "country_code": "BG",
+    "description": "Latest orthorectified aerial imagery provided by the Ministry of Agriculture and Food of Bulgaria.",
+    "permission_osm": "explicit",
+    "license_url": "https://inspireportal.egov.bg/geonetwork/srv/bul/catalog.search#/metadata/b13b693f-3702-49f4-897c-9021b84e83e3",
+    "privacy_policy_url": false,
+    "vintage_url": "??????????",
```
Dimitar5555: (TBD)

Dimitar5555:
It is ready for review. Let me know if you prefer another name for `vintage_url`.

ignisf (approved):

grischard:
The idea is good, but please discuss the parameter with the editor maintainers before we can merge.

grischard:
If you want, we can split this PR in two and merge just the new imagery quickly.

ignisf:
@grischard, I believe @Dimitar5555 has already discussed this with @tyrasd

Dimitar5555:
> The idea is good, but please discuss the parameter with the editor maintainers before we can merge.

I've discussed it Martin Raifer (tyrasd) already. I'm not sure if there are other editors which use ELI and also show the vintage of a given imagery. Unless there is such editor, I don't see a reason for splitting this PR.

grischard:
At least the seven editors in the readme should be consulted and not object. Schema compatibility should be kept with JOSM.

The schema of the vintage geojson should be documented, and the QA scripts should check that it's accessible and valid.

Dimitar5555:
I've dropped the `vintage_url` parameter for now. I will discuss it with the other editors before making a PR for it.

grischard:
I'm not sure if any of the editors display the imagery date parameters we currently have?

An alternative to your original PR that doesn't require schema changes would be to include multiple layers with the same endpoint and different dates.

Dimitar5555:
> An alternative to your original PR that doesn't require schema changes would be to include multiple layers with the same endpoint and different dates.

It won't work in iD and in other editors likely. You can start in one end of the country then move within the editor to the other end of the country and it will still show the same date and imagery. At least until you change the imagery or reload the page.

andrewharvey:
I think it's a great idea to store vintage information for mosaics, but let's keep that in another ticket.

ignisf:
@andrewharvey, is it OK to merge this now since @Dimitar5555 has removed the `vintage` element? Just so we get the imagery out there before people have started using the new misaligned ESRI imagery.

PR #2683, opened 2025-04-30
Add 4 open data orthoimagery in Auvergne-Rhone-Alpes from CRAIG's WMS

Campanu82:
Hello,

Not sure about the method (first contribution), but following my post on https://forum.openstreetmap.fr/t/orthophotos-grand-lyon-2023-et-vallee-du-veneon-2024/33615, I try to add 4 recent orthoimagery from CRAIG available in WMS : 
- grand_lyon_2023
- veneon_juin_2024
- veneon_octobre_2024
- vichy_2021.

These images are released under Licence Ouverte / Open Licence v2.0.

PR #2800, opened 2025-09-19
Add Lyon orthophoto 2023 and latest

cnotin:
Copied from https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/fr/Lyon-Orthophoto-2022.geojson

They now offer a "latest" URL so I think it's a good idea to use it so we always get the freshest layer, even though it means that the metadata will become stale if they update what's the latest one. WDYT?


cnotin:
Well actually the 2023 is a duplicate of https://github.com/osmlab/editor-layer-index/pull/2683/files#diff-40125fc24328a3ff2030c0f45348fd6c5fbcb672c295628a5e85e2638541346c which is the same dataset but hosted by a different organization

PR #2801, opened 2025-09-20
add lower saxony orthophoto

Tillbtn:
add the newly available orthophotos for Lower Saxony, Germany


vsandre:
Already integrated with duplicate #2807

PR #2803, opened 2025-09-23
Add filter functions to UI for selecting by country

joto:
This adds a downdown menu which allows filtering the layer shown by country. The list is automatically populated with all available country codes + "world".

filter-be


grischard:
Thanks Jochen

PR #2806, opened 2025-09-27
New source: Germany Rheinland-Pfalz

joto:


vsandre:
Checks fail because of newly added icon

Created a [pull request](https://github.com/joto/editor-layer-index/pull/1) to @joto to include historic orthophotos and ground shading

PR #2813, opened 2025-10-01
Add Italian cadastral mapping provided by the Italian Revenue Agency (AgE)

Cm-8:
The Italian Revenue Agency (AgE) offers cadastral mapping in compliance with the European Open Data Directive using the Web Feature Service (WFS) standard.
The service is available under the CC-BY 4.0 license.

More information here:
- https://www.agenziaentrate.gov.it/portale/cartografia-catastale-wfs
- https://wms.cartografia.agenziaentrate.gov.it/inspire/wms/ows01.php?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities


matheusgomesms:
Grazie!

However, I'm not sure we are allowed to use CC-BY without an explicit authorization, as better explained here: https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

Do you think you could manage to get an authorization to use their data?

PR #2816, opened 2025-10-03
Add new Source Mecklenburg Vorpommern

vsandre:
DOP, DTK, DGM, DTK, borders, historic DOP 2019-2024


vsandre:
#2443 one more German state

PR #2817, opened 2025-10-07
[website] allow linking to a specific imagery layer

k-yle:
Currently, [the ELI website](https://osmlab.github.io/editor-layer-index) does not allow you to share a link to a specific imagery layer.

This PR makes the [URL hash](https://developer.mozilla.org/en-US/docs/Web/API/URL/hash) synchronise with the layer that's selected on the website. 

Now you can send someone a link like this: 
https://osmlab.github.io/editor-layer-index/#Rio%20Mosaico%202022

to test locally: run `yes | npx serve` (or equivalent command to start a local webserver)


grischard:
Good idea, thanks!

PR #2819, opened 2025-10-12
Add Georgian Orthophoto plans

ekatyshev:


matheusgomesms:
Thanks for the contribution! Check is failing, do you know if there is a privacy URL available from that source?

ekatyshev:
> do you know if there is a privacy URL available from that source?

No, there isn't such an URL

matheusgomesms:
Can we use this one? https://www.napr.gov.ge/en/single-page/processing-of-personal-data (asking to the other maintainers, if this page is acceptable)

ekatyshev:
Can I do something to expedite the decision?

ekatyshev:
The matter has been rised again https://github.com/osmlab/editor-layer-index/issues/2993

nikyk:
Orthophoto is extremely useful for Georgia, please add this layer

urli-by:
This winter, I filled up an entire settlement's streets and building numbers using data from NAPR in another window. If this layer were directly accessible, the process would be a hundred times simpler.

PR #2820, opened 2025-10-13
AGIS 2025

simonpoole:
This is ortho imagery from spring 2025.

I've set the best flag as this is the most current imagery for the area
in question.


matheusgomesms:
Thanks, Simon!

PR #2826, opened 2025-10-25
Update Aachen Liegenschaftskataster URL

jomo:
The old URL is no longer working. Otherwise there don't seem to be any changes.


jomo:
I forced push an update to add some more projections. The `GetCapabilities` request does not include them, but the server supports them anyways.

grischard:
Thanks!

PR #2827, opened 2025-10-25
Mark Munich imagery as preferred

map-per:
Marked the _LandeshauptstadtMuenchen20cm_ imagery as "best" as it's the best imagery available in Munich. It has the highest resolution and it's an orthophoto.


matheusgomesms:
Danke! :D

map-per:
:wave:

PR #2845, opened 2025-11-12
Increase max_zoom from 14 to 15 in MTBmapno.geojson

fredrik-lindseth:
By request of the map author / owner @Reitstoen.

No tiles exist at z14

### DIFF
--- a/sources/europe/no/MTBmapno.geojson
+++ b/sources/europe/no/MTBmapno.geojson
@@ -119,7 +119,7 @@
         "description": "Norwegian mountain biking map from OSM (max zoom 14-16, varies per region)",
         "icon": "https://osmlab.github.io/editor-layer-index/sources/europe/no/MTBmapno.png",
         "id": "mtbmap-no",
-        "max_zoom": 14,
+        "max_zoom": 15,
         "min_zoom": 3,
         "name": "MTBmap.no",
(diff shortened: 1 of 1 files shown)


andrewharvey:
> INFO:root:Processing sources/europe/no/MTBmapno.geojson
> INFO:root:Type: tms
> INFO:root:Zoom levels reachable. (Tested: 3,4,5,6,7,8,9,10,11,12,13,14,15)
> ERROR:root:sources/europe/no/MTBmapno.geojson has no license_url set
> ERROR:root:sources/europe/no/MTBmapno.geojson has no privacy_policy_url. Adding privacy policies to sources is important to comply with legal requirements in certain countries.

The PR check is stricter now, if there is no privacy policy, you can set `privacy_policy_url: false`.

I'm not across the licensing/permissions for this source, but there's no `license_url` set.

The Website say "© [OpenStreetMap](https://www.openstreetmap.org/copyright) | © [Kartverket](http://www.kartverket.no/) | © [mtbmap.no](https://mtbmap.no/info)" so there's more copyright than just OSM. I can see hillshading and contours, where are these sourced and are these okay for tracing from?

I know you're just changing the maxzoom, but it's also a good time to ensure we can use this source for mapping.

PR #2852, opened 2025-12-04
Add Canada Ontario Ortho 2023-2027

galen8183:
Add Geospatial Ontario's 2023-2027 orthophoto WMS layers from https://geohub.lio.gov.on.ca/maps/lio::geospatial-ontario-imagery-data-services/about
These files are identical to the 2018-2022 layers apart from updated dates

This was recently requested in the OSM community forum: https://community.openstreetmap.org/t/choice-of-ontario-based-backgrounds-in-id-editor/137159

### DIFF
--- a/sources/north-america/ca/on/Ontario_Mosaic_2023-2027_CIR.geojson
+++ b/sources/north-america/ca/on/Ontario_Mosaic_2023-2027_CIR.geojson
@@ -1,0 +1,132 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "Ontario_Mosaic_2023-2027_CIR",
+        "attribution": {
+            "url": "https://www.ontario.ca/page/open-government-licence-ontario",
+            "text": "Contains information licensed under the Open Government Licence – Ontario",
+            "required": true
+        },
+        "name": "Ontario Orthophotography Mosaic 2023-2027 (CIR)",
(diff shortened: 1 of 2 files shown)


galen8183:
> 72.67% of geometry is outside of the layers bounding box. Geometry should be checked

This is because the layer is (obviously) incomplete, LMK if the geometry should match the currently available imagery and be updated later, or kept as the future boundary

andrewharvey:
> This is because the layer is (obviously) incomplete, LMK if the geometry should match the currently available imagery and be updated later, or kept as the future boundary

I think it's better if the geometry matches the currently available imagery, so editors you don't suggest it in areas where there's no coverage. Though I understand that increases maintenance, more likely the imagery coverage increases and we don't pick up on it straight away.

Since you haven't marked these as `best: true` they won't be the defaults, so in that case I think it's okay to leave the geometry matching the expected future state as the imagery get's populated.

andrewharvey (approved):

galen8183:
FWIW I will be keeping an eye on when this imagery updates since I edit across Ontario frequently and am happy to PR updates as needed. Judging by current release cycles the layer will likely only be complete in early-to mid-2028 (roughly one year after acquisition), so perhaps it is best to follow it's true geometry now.

PR #2870, opened 2026-01-02
Updated layers in Latvia

Davis-Klavins:
Updated layers in Latvia, including changes resulting from comparison with JOSM imagery sources (https://josm.openstreetmap.de/wiki/Maps/Latvia).


cicku:
Why did you change the license from https://wiki.openstreetmap.org/wiki/Attribution/vzd.gov.lv_explicit_permission to a page with no license mention?

Davis-Klavins:
Because it relates to the service provided by LVM GEO (they don't have a dedicated license, only description page that states that data are free to use). Data provided by that service are the same to which applies https://wiki.openstreetmap.org/wiki/Attribution/vzd.gov.lv_explicit_permission.

PR #2880, opened 2026-01-12
Adding a new source with Karlovac imagery from 2023, and changing the url for Rijeka imagery

Janjko:
Rijeka imagery wasn't working on some browsers, so I changed the url of the imagery, to the osm-hr server with the same imagery.

Karlovac is new, on the same server.


Janjko:
The strict check has two errors I don't know how to handle. First, the message `ERROR:root:Could not retrieve url: https://data.rijeka.hr/hr/dataset/gis-grad-rijeka:`

That URL works for me, so I'm guessing they are blocking Github Actions IPs?

And `ERROR:root:100.0% of geometry is outside of the layers bounding box. Geometry should be checked`

I checked both geojsons and the imageries, and they align. I'm not sure if that is what the message is implying.

andrewharvey:
> That URL works for me, so I'm guessing they are blocking Github Actions IPs?

More likely geoblocking, as it doesn't work for me. If that's the case we can ignore the error, but it has implications for editors that they are also only going subject to the same geoblocking. It's something we could improve in the future with additional metadata about which countries the services are accessible from.

UPDATE: looks like it's only the attribution URL that is the issue, not the imagery URL.

> And ERROR:root:100.0% of geometry is outside of the layers bounding box. Geometry should be checked

It might be an issue with the check, especially if it was unable to retrieve the layers bounding box, it might be buggy.

sources/europe/hr/osm-hr-karlovac-2023.geojson
```diff
@@ -0,0 +1,86 @@
+{
+    "type": "Feature",
+    "properties": {
```
andrewharvey, line 3: should this have an attribution property?

Janjko: Sorry about the late reply, I added the attribution for Karlovac, thanks.

Yes, there's geoblocking for Rijeka, I don't know why they do that.

Janjko:
Bump, I added attribution, tests that aren't passing seem like not a big deal.
Thanks.

Issue #2871, opened 2026-01-04
DVRPC Orthoimagery 2020 - Delaware County (1ft) not displaying correctly


Tre-brock:
In the ID editor the imagery has various artifacts, but when manually added they go away. It also has a smoothing affect on the ID version of the imagery that distorts shapes.
<hr>
The custom url is 

```
https://imagery.pasda.psu.edu/arcgis/services/pasda/DVRPC2020/MapServer/WMSServer?format=image\jpeg&version=1.3.0&service=wms&request=getmap&layers=4&styles=&crs={proj}&width={width}&height={height}&bbox={bbox}
```

ID

<img width="1512" height="866" alt="Image" src="https://github.com/user-attachments/assets/32341a2e-fa3e-42a3-9c1d-1fd0c6253aa1" />

Custom

<img width="1512" height="866" alt="Image" src="https://github.com/user-attachments/assets/0e24cfd8-c81b-458d-ae63-d62a8961dfe2" />

Tre-brock:
Note: This same thing seems to happen in all the imagery with this type of name

<img width="753" height="506" alt="Image" src="https://github.com/user-attachments/assets/86000c2a-dab8-4669-86f7-ec9835e6340b" />

Tre-brock:
Update: new imagery has been released by the same origination, its works by replacing 2020 in the URL with 2025, it would be good if this was added to iD. It's 1ft.

andrewharvey:
does #2882 fix this, or is this issue still present even with #2882?

Tre-brock:
I'm not sure what causes the issue, I did not fix it in my commit. All I did was add newer imagery from the same source.

Tre-brock:
Small update, I checked the [ideditor.netlify.app](https://ideditor.netlify.app) and the same problem is occurring on the new imagery. For some reason when it is manually imputed into iD it doesn't causes problems. The weirder thing is every time I clear my cache the artifacts are slightly different colors. Diffrent browsers cause the artifacts to be in different places.

BushmanK:
Visually, it looks like a corrupt JPEG artifact. In this context, it could be caused by some weird network errors rather than anything related to the iD or, moreover, the WMS layer connection template.

cicku:
Pretty sure it is something with iD and browser. Closing.

PR #2882, opened 2026-01-16
Updated DVRPC to 2025

Tre-brock:
This pr updates all of the "DVRPC" to the most recent 2025 imagery.

This is related to the issue: https://github.com/osmlab/editor-layer-index/issues/2871


cicku:
Unfortunately 2020 has to be kept because we generally carry at least 3 versions of the same imagery (for comparison/regression), please revise the PR and submit 2025 as new files instead.

Tre-brock:
I've fixed the issue

PR #2887, opened 2026-01-23
fix: source: MD Latest 3 Inch & 6 Inch Aerial Imagery

naterkane:
MD iMAP 3.0 ArcGIS Server Site has been Deprecated
* update paths to point to MD iMAP 4.0 ArcGIS Server
* update paths to icon assets, and other referenced pages 

see https://mdgeodata.md.gov/Deprecation/ 

Note: the only reasonable icon image I was able to find was on the deprecation page itself. I presume it won't exist there forever, but all other paths to icons (including on the MD website itself) seem to be broken.

I do believe the 3 Inch imagery geojson is in need of updating, since the imagery itself has changed and been added to since the previous entry. If the case, I suggest that update be handled separately to improve the quality of what's provided. This PR is essentially a hotfix as the MD 3 & 6 Inch imagery is currently completely unavailable.


naterkane:
Added updated getoJSON for this source based on this endpoint's response data https://mdgeodata.md.gov/imagery/rest/services/ThreeInch/MD_ThreeInchImagery/ImageServer/queryBoundary?outSR=4326&f=pjson

PR #2889, opened 2026-01-23
fix: source: MD Latest 3 Inch & 6 Inch Aerial Imagery

naterkane:
MD iMAP 3.0 ArcGIS Server Site has been Deprecated

update paths to point to MD iMAP 4.0 ArcGIS Server
update paths to icon assets, and other referenced pages
see https://mdgeodata.md.gov/Deprecation/

_Note: the only reasonable icon image I was able to find was on the deprecation page itself. I presume it won't exist there forever, but all other paths to icons (including on the MD website itself) seem to be broken._

I do believe the 3 Inch imagery geojson is in need of updating, since the imagery itself has changed and been added to since the previous entry. If the case, I suggest that update be handled separately to improve the quality of what's provided. This PR is essentially a hotfix as the MD 3 & 6 Inch imagery is currently completely unavailable.
> _Note: geoJSON updated in https://github.com/osmlab/editor-layer-index/commit/f46869a0475e1986f3a9b592d76ede42edbfdd62_

_This PR is a duplicate of https://github.com/osmlab/editor-layer-index/pull/2887, as not having `gh-pages` as the source branch may have been a problem. A "1 workflow awaiting approval" status gave me the sense something was up._


naterkane:
Looking through recently merged PRs, can @cicku or @andrewharvey lend some eyes? 
This PR is a hotfix, not a feature or enhancement. 

Additionally re: https://github.com/osmlab/editor-layer-index/pull/2882#issuecomment-3758089040 I would be happy to add additional/historic imagery sources. I would hope that could be a separate PR as earlier versions of this imagery had not been previously included.

cicku:
> Additionally re: [#2882 (comment)](https://github.com/osmlab/editor-layer-index/pull/2882#issuecomment-3758089040) I would be happy to add additional/historic imagery sources. I would hope that could be a separate PR as earlier versions of this imagery had not been previously included.

That's off topic. We only keep historical layers when the sources are eligible, MD doesn't apply.

naterkane:
> > Additionally re: [#2882 (comment)](https://github.com/osmlab/editor-layer-index/pull/2882#issuecomment-3758089040) I would be happy to add additional/historic imagery sources. I would hope that could be a separate PR as earlier versions of this imagery had not been previously included.
> 
> That's off topic. We only keep historical layers when the sources are eligible, MD doesn't apply.

Fair enough, the imagery is available, but I couldn't have known more context... I appreciate it. 

I'll look into the output from `Strict check` and update the PR, thanks!

PR #2890, opened 2026-01-27
fix: migrate NVE Maps to new URLs

Johannes-Andersen:
Migrated from the old APIs over to the new APIs as instructed in their migration dashboard https://kart.nve.no/enterprise/apps/dashboards/323c7238f3d4434389ad68040a4a5809

Old API endpoints will be turned off 15.03.2026, so in about 35 days

The old Nettanlegg2 service is deprecated, so it will not be migrated to the new kart.nve.no domain. It was thefore necessary to also migrate the Nettanlegg2 over to Nettanlegg4.

Also migrated the icon to use the official website, to avoid relying on fetching icon from third party news website.
 
Since the license_url was pointed to a now taken down email archive, I moved it to web archive until we can get a better permanent place for it.


Johannes-Andersen:
Should be good to review now :D

PR #2892, opened 2026-01-27
fix: move to new SSB domain

Johannes-Andersen:
Update to new ssb.no domain as the old domain is shutdown, and is therfore causing 500 errors for anyone trying to use it.
Migration docs https://register.geonorge.no/varsler/wms-alle-data-alle-%C3%A5rganger-fra-ssb/fdc28ce0-c702-45ae-b0bc-6d73381cfeab

Also updated the icon to not point to unrelated third party website `http://fosterhjemsforening.no/`, but to use the official SSB.no site

Since the license_url was pointed to a now taken down email archive, I moved it to web archive until we can get a better permanent place for it.


Johannes-Andersen:
Hummm, I see that most if not all the Norwegian sources link to a email list that is now 404ing
I’ll see if I can get ahold of an archive or a more recent permission confirmation

Johannes-Andersen:
Apolegies for the tag @andrewharvey, but would you be able to help me get this deployed to fix the failing calls in iD editor?
Possibly also look into https://github.com/osmlab/editor-layer-index/pull/2890 before it gets shut down in 30 days to avoid it also failing 🙏 

In advance, thank you!

PR #2897, opened 2026-01-31
Updated NYSDOP imagery to include latest 2025 layer

arlen-kleinsasser:
Updated NYSDOP natural color and ir imagery to include latest 2025 layer.


andrewharvey:
Thanks. I've merged this, but in my view it's best to split this into one source for each layer, so mappers can compare between the new/old images, helpful for spotting differences or just to understand if what someone has mapped is what was there a few years back and now demolished or just something new that isn't captured in the imagery yet.

jmapb:
@andrewharvey Don't think that would work well. The layers do represent a single year's imagery and we can request individual layers (LAYERS=0 or LAYERS=1 etc) but most NY counties only appear in a single layer -- we have to add them all up to get the whole state. And *which* counties each layer has changes too -- 0 is always the oldest year's imagery, then 1, etc.

As we see from this PR, the total number of layers isn't static either. Each spring some counties get fresh photography, and the new imagery is published per county when the processing is complete. Layer 4, with new 2025 imagery, started showing up late last year. At that point layer 0 was 2021. New York has been adding additional counties' 2025 imagery to layer 4 over the past few months. Eventually all of the 2021 imagery was obsolete, so the 2021 layer was removed from the WMS server a few days ago. Layers 0,1,2,3 now map to 2022,2023,2024,2025 and there's no more layer 4.

Unfortunately, requesting a nonexistent layer will generate an error and return no imagery, so there's a new PR (https://github.com/osmlab/editor-layer-index/pull/2920) to remove layer 4 from the urls. It looks like we'll have to keep up this ritual every year -- adding layer 4 when the first new county is published, then removing 4 when the layers shift down after the last 5-year-old imagery is replaced.

andrewharvey:
Thanks for the details @jmapb, makes sense.

PR #2905, opened 2026-02-15
Add latest Geneva orthophoto (2024) (4cm)

alkino:
This ortho has been published in 2026.
This is the best we got: 4cm !!!
Geneva is a city surrounded by France. So for this version (and the previous one) they worked with French services (IGN) so the imagery does not stop at the border!


sources/europe/ch/Canton_de_Geneve_orthophoto2024_4cm.geojson
```diff
@@ -0,0 +1,138 @@
+{
+    "type" : "Feature",
+    "properties" : {
+       "license_url" : "https://ge.ch/sitg/media/sitg/files/documents/conditions_generales_dutilisation_des_donnees_et_produits_du_sitg_en_libre_acces_-_revision_09.2020.pdf",
```
andrewharvey: Is that the correct link? For me it just redirects to https://sitg.ge.ch/

alkino: I'm unable to find a proper URL for license.
It is written "Accès libre" (free access), with the definition of what it means in the following "privacy policy url"

simonpoole:
This should likely get the best flag set to true. 

The swisstopo imagery was last updated in 23/24 so we should revisit that towards the end of the year/early 27 as then swisstopos three year cycle would indicate that they will have new coverage for Geneva.

alkino:
> This should likely get the best flag set to true.
> 
> The swisstopo imagery was last updated in 23/24 so we should revisit that towards the end of the year/early 27 as then swisstopos three year cycle would indicate that they will have new coverage for Geneva.

Done

andrewharvey:
Looks good, but I'm finding it hard to determine if the license is compatible, since as part of the review, we need to verify each added source can be used to derive features for OSM, I can see from https://sitg.ge.ch/actualites/nouvelle-orthophoto-du-canton-de-geneve-mars-2024 that it is under "Accès libre" per https://sitg.ge.ch/ressources/conditions-utilisation-donnees

It seems to be listed at https://wiki.openstreetmap.org/wiki/Switzerland/Datasources as compliant, but it doesn't really say why, only pointing out an agreement was obtained, but no link to the agreement, and the mailing list link being broken. http://lists.openstreetmap.ch/pipermail/talk-ch/2014-February/002580.html

@simonpoole If you have any thoughts on how we can improve the situation regarding validating license approvals as part of the PR reviews, I'm keen to hear.

alkino:
What can we do? All the previous version are already added and have the same license.

simonpoole:
https://wiki.openstreetmap.org/wiki/SITG_WMS would indicate that using the imagery is permitted, which has been my understanding for a long time. That seems to have been added by @rbuffat BTW. I've checked my personal mail archive and the one on lists.openstreetmap.ch and couldn't find a smoking gun, but IMHO the wiki entry would seem to be good enough.

As everything on openstreetmap.ch is run by a private individual it isn't particularly good as an archive of important documents, so a lot of the old stuff might vanish just like that. In any case @andrewharvey I've long suggested (long as starting with the pre-licence change work) that how we document permissions, contacts and terms for 3rd party sources is completely broken, see for example https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2017/Project_Ideas#Third_Party_Source_Management_System_2 for a better way.

andrewharvey:
> In any case @andrewharvey I've long suggested (long as starting with the pre-licence change work) that how we document permissions, contacts and terms for 3rd party sources is completely broken, see for example https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2017/Project_Ideas#Third_Party_Source_Management_System_2 for a better way.

Completely agree.

PR #2907, opened 2026-02-25
Standardize GRAFCAN files, and add topographic layer

Crashillo:
Rename previous files and and the topographic layer

Included a description for all spanish services


Crashillo:
Can anyone review this? The failing check is due to Error 429, mostly, or unrelated issues to already working layers

PR #2910, opened 2026-03-05
Add the Turku 2025 ortophoto layer

crimsondusk:
I also deleted the old 2018 layers, and updated the icon URLs


andrewharvey:
If the 2018 imagery is still available you could retain it as `historicphoto`.

PR #2917, opened 2026-03-10
move npm package name to the @openstreetmap namespace

tyrasd:
This unblocks https://github.com/openstreetmap/iD/pull/11966/, preventing us to avoid an annoying false positive "malware" detection when using ELI as a npm dependency. As a secondary effect, this would also allow us to publish ELI data as an actual npm package at some point in the future, if we would like to do so eventually.

Given that following https://github.com/openstreetmap/iD/issues/10772#issuecomment-3297426336, my attempt to fix this issue via npm's support were (still as of today) not successful, and that it seems like there is no interest/progress in fixing the underlying [upstream](https://github.com/npm/cli/issues/5110) issue, the only way to prevent the mentioned false positive in `npm audit` is to rename the package in `package.json`. I propose to use `@openstreetmap/editor-layer-index` in order to be in line with similarly named packages like `@openstreetmap/id-tagging-schema`.

For the time being, I have already parked a dummy [package](https://www.npmjs.com/package/@openstreetmap/editor-layer-index) on npm with the proposed name/namespace with only a readme with instructions about how to use ELI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "editor-layer-index",
+  "name": "@openstreetmap/editor-layer-index",
   "version": "0.0.0",
   "description": "available imagery for osm editors",
   "main": "index.js",


andrewharvey (approved):

tyrasd:
@andrewharvey shall we go ahead and merge this now? I was wondering if there are any projects that consume ELI as an npm package that would need to adjust to the new name, but apart from iD and its clones, I don't think there are any. But maybe I'm forgetting something.

andrewharvey:
> @andrewharvey shall we go ahead and merge this now?

Yes please do.

> I was wondering if there are any projects that consume ELI as an npm package that would need to adjust to the new name, but apart from iD and its clones, I don't think there are any. But maybe I'm forgetting something.

According to https://docs.npmjs.com/cli/v8/using-npm/package-spec#git-urls there are a few ways to specify a GitHub dependency.

A quick search on GitHub for `github:osmlab/editor-layer-index` show's mostly just iD and Rapid, but that package name qualifier should continue to work regardless of what's in package.json.

I also searched for `osmlab/editor-layer-index`. It's used by https://github.com/ngageoint/hootenanny-ui but from everything I could see it shouldn't matter what name is in ELI's `package.json`.

PR #2918, opened 2026-03-14
Add Maple Ridge imagery for 1978, 2016, 2018, 2020, and 2023

maxwellward:
Hello! This PR adds the imagery available from the City of Maple Ridge for 2016, 2018, 2020, and 2023, along with a historical entry from 1978.

I have marked the 2023 imagery as "best" (it is also better than Esri/Bing). I wasn't too sure if I should mark the older pre-2023 imagery as historic, as the docs say I should "when there exists a more recent photo source", but some other examples I looked at (such as Toronto) still mark the older "modern" imagery as "photo".

2016, 2018, and 2020 share a coordinate boundary, while 2023 and 1978 have their own unique coordinates as their coverage is smaller. 

I have tested all of these sources as custom layers in the iD editor.

Please let me know if I've done anything wrong or can improve anything! :)

### DIFF
--- a/sources/north-america/ca/bc/MapleRidge_1978.geojson
+++ b/sources/north-america/ca/bc/MapleRidge_1978.geojson
@@ -13,7 +13,7 @@
         "license_url": "https://osmfoundation.org/wiki/OGL_Canada_and_local_variants",
         "permission_osm": "implicit",
         "country_code": "CA",
-        "end_date": 1978,
+        "end_date": "1978",
         "type": "wms",
         "available_projections": [
             "EPSG:3857",
--- a/sources/north-america/ca/bc/MapleRidge_1978.geojson
+++ b/sources/north-america/ca/bc/MapleRidge_1978.geojson
@@ -1,0 +1,78 @@
+{
+    "type": "Feature",
+    "properties": {
+        "id": "Maple-Ridge-1978",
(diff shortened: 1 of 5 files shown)


sources/north-america/ca/bc/MapleRidge_1978.geojson
```diff
@@ -0,0 +1,78 @@
...
+            "text": "Contains information licensed under the Open Government Licence – Maple Ridge",
+            "required": true
+        },
+        "name": "City of Maple Ridge Orthoimagery (1978)",
+        "url": "https://geoservices.mapleridge.ca/image/rest/services/Orthoimagery/1978_Mosaic/ImageServer/exportImage?f=image&bbox={bbox}&bboxSR={wkid}&imageSR={wkid}&size={width},{height}&format=jpgpng",
+        "max_zoom": 24,
+        "license_url": "https://osmfoundation.org/wiki/OGL_Canada_and_local_variants",
+        "permission_osm": "implicit",
+        "country_code": "CA",
+        "end_date": 1978,
```
andrewharvey: per the CI check, this should be a string

maxwellward: Oops! Missed that in the schema. Fixed, thanks.

PR #2934, opened 2026-04-03
Add Massachusetts MassGIS 2025 aerial imagery

petercooperjr:


petercooperjr:
The license and icon URLs seem to work from a normal browser for me, even though the automated checks can't get to them. I suspect there may be some overly-zealous firewall trying to block non-browser traffic.

andrewharvey:
Looks like we are missing a start/end date on the 2023 imagery, I won't block this PR because of that, but you might want to follow up with another PR.

petercooperjr:
> Looks like we are missing a start/end date on the 2023 imagery

I *think* it's there, especially as that's what I made a copy of in order to add the 2025 imagery. In `MassGIS_2023_Aerial.geojson` I see

        "start_date": "2023-03-20",
        "end_date": "2023-04-28",

Is that not working right?

andrewharvey:
> Is that not working right?

sorry I was just looking at the diff and missed it.

PR #2941, opened 2026-04-08
Add 2003-2025 Coquitlam BC imagery

maxwellward:
Hello! This PR adds imagery sources for Coquitlam BC all the way from 2003 to 2025.

I have included every year (other than the 1963 imagery) as I'm not too sure about how far back we generally want to go with historical imagery. Should we only include the last few years, or do we want to include as much as is available? 

I am happy to remove any number of years from this PR if this is too many.

Also note that the 2022 imagery uses `Imagery_SC` in the URL as it is the same imagery as the standard 2022 URL, just with higher quality and the same license.

Thanks!


arch0345:
Thanks for finding this imagery source! Before this gets merged, can you confirm with the LWG that Coquitlam's OGL is compatible? Ideally this license should be explicitly listed as compatible here: https://osmfoundation.org/wiki/OGL_Canada_and_local_variants

maxwellward:
Hello, thanks for your comment. I've confirmed that the Coquitlam OGL is not substantially different from the Ottawa OGL in accordance with the linked wiki page. I have also sent the information to the LWG for addition to the Canada wiki page.

maxwellward:
I haven't yet heard back from the LWG about this, but from everything I can tell it's compatible. Are we able to proceed without final confirmation from the LWG given the licenses are effectively identical?

arch0345:
I did note that the license states that it's essentially the same as OGL BC, which is already listed as approved on that wiki page, so there shouldn't be any issues. I'll go ahead and merge this

PR #2942, opened 2026-04-10
MLIT Plateau new imagery (JP)

j9d3it:
Imagery has been added for MLIT Plateau but I'm not sure how to change the multipolygon to match it, it's too complicated for me. Also, since it's not hosted on AWS anymore do we need the privacy policy?


arch0345:
Do you know if this old source will be deprecated soon or if it'll stay up? I noticed in some areas, the old source is higher resolution so it might be worth keeping it around. In other areas, it's the other way around.

|Old (AWS) | New (mlit.go.jp) |
| -------- | -------- |
| image | image |
| image | image |

j9d3it:
> Do you know if this old source will be deprecated soon or if it'll stay up? I noticed in some areas, the old source is higher resolution so it might be worth keeping it around. In other areas, it's the other way around.
> 
> Old (AWS)	New (mlit.go.jp)
> image	image
> image	image

The old source will stay online I think. So it's probably better to make a new one instead.

j9d3it:
Just looking at the imagery again, it's pretty blurry everywhere so I'm not going to bother adding it.

PR #2951, opened 2026-04-17
Update City of Toronto (Ontario, Canada) imagery

tallcoleman:
Myself and a couple other mappers noticed that the City of Toronto Orthoimagery had stopped working after their latest update. Switching the source to the TMS URL seems to work (or at least works if you add the URL as a custom source in the iD editor).


arch0345:

Issue #2952, opened 2026-04-17
Oregon DOGAMI Lidar WMS URL is incorrect


BushmanK:
For DOGAMI Lidar data, the WMS URL template in the list looks like this, with `LAYERS=0`:

`https://gis.dogami.oregon.gov/arcgis/services/lidar/DIGITAL_TERRAIN_MODEL_MOSAIC_HS/ImageServer/WMSServer?LAYERS=0&STYLES=&CRS={proj}&BBOX={bbox}&FORMAT=image/jpeg&WIDTH={width}&HEIGHT={height}&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap`

However, requesting a tile like this [https://gis.dogami.oregon.gov/arcgis/services/lidar/DIGITAL_TERRAIN_MODEL_MOSAIC_HS/ImageServer/WMSServer?LAYERS=0&STYLES=&CRS=EPSG:3857&BBOX=-13646914.15407913,5679424.0748545835,-13646761.280022582,5679576.948911131&FORMAT=image/jpeg&WIDTH=256&HEIGHT=256&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap](https://gis.dogami.oregon.gov/arcgis/services/lidar/DIGITAL_TERRAIN_MODEL_MOSAIC_HS/ImageServer/WMSServer?LAYERS=0&STYLES=&CRS=EPSG:3857&BBOX=-13646914.15407913,5679424.0748545835,-13646761.280022582,5679576.948911131&FORMAT=image/jpeg&WIDTH=256&HEIGHT=256&VERSION=1.3.0&SERVICE=WMS&REQUEST=GetMap) results in `<ServiceException code="InvalidLayers"> Parameter 'Layers' contains invalid values. </ServiceException>` rather than an image.

The [WMS Capabilities request](https://gis.dogami.oregon.gov/arcgis/services/lidar/DIGITAL_TERRAIN_MODEL_MOSAIC_HS/ImageServer/WMSServer?request=GetCapabilities&service=WMS) to gis.dogami.oregon.gov returns this:
```
<Layer queryable="1">
<Name>DIGITAL_TERRAIN_MODEL_MOSAIC_HS</Name>
<Title>DIGITAL_TERRAIN_MODEL_MOSAIC_HS</Title>
```
Changing the `LAYERS` parameter value from `0` to `DIGITAL_TERRAIN_MODEL_MOSAIC_HS` returns a correct image:
[...]

PR #2953, opened 2026-04-18
Fixed DOGAMI-Lidar.geojson

BushmanK:
Fixed DOGAMI-Lidar.geojson to restore functionality due to the changed layer name. See #2952


arch0345:

PR #2975, opened 2026-06-10
Fix/remove broken imagery source URLs

KTibow:
## Summary

This updates or removes imagery sources whose current endpoints no longer provide usable public tiles.

Changed sources:

- 🔴 `sources/asia/ph/PangasinnBulacan-PhilippinesHiRes.geojson`: remove the Pangasinan/Bulacan high-resolution source because the `gravitystorm.dev.openstreetmap.org/imagery/philippines` tile directory is gone.
- 🔴 `sources/europe/fr/Enedis.geojson`: remove the Enedis source because the OSM France TMS endpoint returns HTML errors (`403`/`502`) instead of imagery, including when tested as an iD custom background.
- 🟨 `sources/europe/hu/EUFAR-Balaton-orthophotos.geojson`: replace the dead OpenStreetMap Hungary Balaton tile endpoint with the current JOSM `balaton_0` endpoint, update the dead BLI attribution URL, and explicitly mark that no tile-service privacy policy is known.
- 🟨 `sources/north-america/us/fl/Nassau_Ortho_2023.geojson`: update the Nassau County 2023 host from `maps.nassauflpa.com` to `maps.ncpafl.com`.
- 🔴 `sources/north-america/us/fl/Osceola_Ortho_2022.geojson`: remove the Osceola County 2022 source because the old service is gone and the apparent replacement blocks GitHub Actions tile validation.
- 🔴 `sources/north-america/us/fl/Osceola_Ortho_2023.geojson`: remove the Osceola County 2023 source because the old service is gone and the apparent replacement blocks GitHub Actions tile validation.
- 🟨 `sources/north-america/us/fl/Saint_Lucie_Ortho_2024.geojson`: update the service name from `2024_Aerial_Imagery` to `2024_Aerial_Imagery_FDOR`.
- 🟨 `sources/north-america/us/oh/Mercer_OH_2021.geojson`: replace the dead Mercer County GIS WMTS URL with the public ArcGIS Online `Mercer2021Aerial` tile service, and update the dead license URL to the county disclaimer page.
- 🟨 `sources/north-america/us/oh/Miami_OH_2023.geojson`: replace the token-gated ArcGIS Online organization URL with a public ArcGIS Online tile service, update the county attribution URL, and remove the broken county icon URL.

### 🔴 Pangasinan/Bulacan

- Removed source id: `Pangasinan_Bulacan_HiRes`.
[...]

PR #2974, opened 2026-06-09
Fix overdeclared imagery max zooms

KTibow:
## Summary

- lowers `max_zoom` for imagery sources whose currently declared maximum zoom returns no usable imagery
- keeps this as source metadata changes only; no upstream brute-force probing code is added
- includes the original USGS Imagery fix and related strict-validation geometry correction
- incorporates a follow-up audit fix for sources where broader sampling found valid higher-zoom tiles

## What Changed

Representative image spot-check table; for each row, the first column was selected from a sampled in-coverage tile that decoded as non-flat imagery, while the `new max + 1` and `old max` tile URLs failed during probing.

<details>
<summary>Image spot-check table for max_zoom changes</summary>

| New max | New max + 1 | Old max |
|---|---|---|
| **osmim-imagicode-S2B_R005_S7X_20200303T061749**<br><sub>antarctica/osmim-imagicode-S2B_R005_S7X_20200303T061749.geojson</sub><br>z12<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R005_S7X_20200303T061749&z=12&x=2412&y=859) | z13<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R005_S7X_20200303T061749&z=13&x=4824&y=1718) | z13<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R005_S7X_20200303T061749&z=13&x=4824&y=1718) |
| **osmim-imagicode-S2B_R049_SXX_20220316T080929**<br><sub>antarctica/osmim-imagicode-S2B_R049_SXX_20220316T080929.geojson</sub><br>z12<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R049_SXX_20220316T080929&z=12&x=2139&y=900) | z13<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R049_SXX_20220316T080929&z=13&x=4279&y=1801) | z13<br>[](https://imagico.de/map/osmim_tiles.php?layer=S2B_R049_SXX_20220316T080929&z=13&x=4279&y=1801) |
[...]

Issue #1455, opened 2022-02-11
Ensure same polygon orientation for all geometries


rbuffat:
See https://github.com/osmlab/editor-layer-index/pull/1452#issuecomment-1036615123

- [x] Check geometry orientation for new PR in strict_check.py
- [x] Update all existing sources to have a consistent orientation

PR #2977, opened 2026-06-14
Correct GeoJSON ring orientation

KTibow:
Correct all remaining source geometries whose ring orientation does not match GeoJSON RFC7946.

This applies the same orientation convention enforced by `scripts/strict_check.py`: exterior rings counterclockwise and interior rings clockwise.

Why this is needed:

- Issue #1455 explicitly called for consistent polygon orientation across sources.
- PR #1464 added the strict-check validation and emitted corrected geometries for non-RFC7946 rings.
- Commit 9c87fcadc7217979ea526ae539a1af288140ad45 previously updated existing sources to match RFC7946 orientation.
- Some later direct/source additions and updates can still bypass this until those files are changed in a PR, because `.github/workflows/pull_request.yml` runs `scripts/strict_check.py` only on changed files while `.github/workflows/push.yml` runs the basic check only.

This PR is geometry-only: properties, URLs, attribution, dates, and other source metadata are unchanged.

This is blocking follow-up maintenance PRs #2974 and #2975: those PRs touch existing source files, which causes `strict_check.py` to run on them and surface latent RFC7946 ring-orientation failures unrelated to the intended metadata/URL changes.

Validation performed:

- Reoriented all `sources/**/*.geojson` polygons and multipolygons using Shapely `orient(..., sign=1.0)`, matching the helper used by `strict_check.py`.
- Verified no source geometries still differ from that RFC7946-oriented output.
- Verified changed files have no non-geometry JSON changes.
- Ran `git diff --check`.

Disclosure: Generated by GPT 5.5 with human oversight


KTibow:
Tests are failing but I would rather keep this scoped to GeoJSON rather than adding `"https://www.unioncountync.gov/favicon.ico"` and `"privacy_policy_url": false`.

grischard:
Beautiful

PR #2985, opened 2026-06-19
Fix the generated link to test the layers in iD

bigon:
See: https://github.com/openstreetmap/iD/blob/develop/API.md


danieldegroot2:
cc @tyrasd

hlfan (approved):
How was this previous version even passing through the info correctly?

PR #2998, opened 2026-07-07
Add Orthophotos 25cm 2025 imagery (Portugal)

mikedld:
Marking `best` per multiple reports of good alignment. Not sure if 2018/2021 orthos need their `best` removed now.


matheusgomesms:
Obrigado!

PR #3024, opened 2026-08-03
Update Henrico, VA, USA aerial imagery

mhpob:
This PR updates the Henrico, Virginia, USA aerial imagery with new layer IDs after an update by the County's GIS program. It also adds the newest 2026 aerials, marking them as "best".


mhpob:
I believe the PR check is incorrect: It is failing saying that "https://henrico.gov/transparency/terms-of-use/" is unreachable. This website is, at least, reachable by me.

arch0345:
Website probably blocks requests from python

PR #3032, opened 2026-08-15
Use icon for OpenStreetMap-GPS direction key

hlfan:
The current version doesn't work:

And this method also allows the attribution translations to work:


andrewharvey (approved):
This is great!

PR #3034, opened 2026-08-16
Update geometry for Canada Ontario Ortho 2023-2027

galen8183:
Update the mosaic layer geometry to include the now present SWOOP (South Western
Ontario Orthophotography Project) imagery.
New geometry is less detailed because I based this on the existing 2018-2022 layer.


galen8183:
Huh, seems like GEO-IDS no longer (or just doesn't currently?) includes the newest CIR imagery, this was definitely present when the layer was added here...

andrewharvey:
> Huh, seems like GEO-IDS no longer (or just doesn't currently?) includes the newest CIR imagery, this was definitely present when the layer was added here...

I'm not sure.

From your point of view should this PR be merged as is?

galen8183:
Considering the layer is still not present a month later, I think it should be removed here.

PR #3037, opened 2026-08-21
Kendall county il imagery

sar-hernandez:
Added metadata for aerial imagery including properties such as attribution, name and geometry.


arch0345 (commented):
Thought I was writing a commit message instead of a comment, disregard

arch0345:
Thanks for finding this imagery! Before this can be merged, a license url is needed that either shows that this imagery is openly licensed (ie CC0) or that OSM contributors have explicit permission to use this imagery for mapping

sar-hernandez:
hi @arch0345 ! thank you so much for looking into my request! about the license url; i work on behalf of the county and our vector data is CC0, but we do not have a specified license for our imagery which is why i didn't put a license url. how do I go about doing this?

sar-hernandez:
we also don't have a privacy policy! @arch0345

arch0345:
> hi @arch0345 ! thank you so much for looking into my request! about the license url; i work on behalf of the county and our vector data is CC0, but we do not have a specified license for our imagery which is why i didn't put a license url. how do I go about doing this?

That's awesome, I recommend adding a disclaimer to your website (either https://www.kendallcountyil.gov/departments/geographic-information-systems or somewhere on https://geodata.kendallcountyil.gov) that clarifies that OSM contributors have permission to use aerial imagery from Kendall County for mapping and setting that as the license url.

> we also don't have a privacy policy! @arch0345

That's ok! We set `privacy_policy_url` as a false boolean for cases like this

arch0345:
Unfortunately some of these layers are out of scope for this index due to it's age, however these would still be very useful for our sister project OpenHistoricalMap. I've opened a PR to add everything here so that the work you've done didn't go to waste https://github.com/OpenHistoricalMap/ohm-editor-layer-index/pull/20

arch0345:

PR #3042, opened 2026-08-26
fix(scripts): unpack test_url() result so attribution and privacy-policy URL checks actually fire

avalyset:
`test_url()` returns a `(bool, int)` tuple, but both of its call sites test the return value directly:

```python
if not test_url(url, headers):
```

A non-empty tuple is always truthy, so `not test_url(...)` is always `False` and neither branch can be reached. The attribution-URL check (line 944) and the privacy-policy-URL check (line 986) have therefore never reported anything.

This unpacks both call sites explicitly and includes the status code in the message, matching the wording the `license_url` check just above already uses.

### Demonstration

Run against `sources/europe/de/Saxony-DOP20-latest.geojson`, whose attribution URL returns HTTP 404, on current `gh-pages`:

**Before**

```
INFO:root:Finished processing sources/europe/de/Saxony-DOP20-latest.geojson
```

**After**

```
ERROR:root:sources/europe/de/Saxony-DOP20-latest.geojson: could not retrieve attribution url https://geoportal.sachsen.de/cps/metadaten_seite.html?id=26df8686-08cd-4dc2-b459-4d51b9badfe8: HTTP code: 404
```

The full output is otherwise identical — one line differs. A source with a working attribution URL (`Saxony-DOP20-historic-2018-2020.geojson`, on GeoMIS) produces no new message, so this does not introduce false positives.

### Context

#3006 reported four Saxony sources with dead attribution URLs. The same dead pattern turned out to be on twelve, all returning 404, and none of them were ever flagged. That is what prompted this. The source-side migration is in #3041; this PR is only the tooling fix.

### DIFF
--- a/scripts/strict_check.py
+++ b/scripts/strict_check.py
@@ -941,11 +941,12 @@ for filename in arguments.path:
             if "url" in source["properties"]["attribution"]:
                 url = source["properties"]["attribution"]["url"]
 
-                if not test_url(url, headers):
+                reachable, status_code = test_url(url, headers)
+                if not reachable:
                     messages.append(
                         Message(
                             level=MessageLevel.ERROR,
-                            message=f"{filename}: could not retrieve attribution url {url}.",
+                            message=f"{filename}: could not retrieve attribution url {url}: HTTP code: {status_code}",
                         )
                     )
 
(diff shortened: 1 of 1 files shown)


avalyset:
The failing check here is the workflow's hardcoded `sources/europe/ch/*` subset,
which its own comment describes as expected to fail. It runs against Switzerland
regardless of what a PR touches. This PR's diff is a single file,
`scripts/strict_check.py`.

The failures are real and pre-existing. Running the old and new script against the
same two CH files locally: 2 errors before, 4 after. The two new ones come from
line 986, the privacy-policy branch. Line 944 is the attribution branch, dead in
the same way, and it fires in the full run (`could not retrieve attribution url`
on KantonBasel-Landschaft12cm-2015). Both tested a `(bool, int)` tuple with
`if not test_url(...)`, which is always truthy. That's why the dead attribution
URLs in #3041 sat unreported.

The workflow comment asks for no Python errors, so one thing to flag: the
`RemoteDisconnected` traceback against geo.bl.ch is caught inside `test_url`,
logged, and the run continues. It's pre-existing. `test_url` is called identically
in both versions, since `if not test_url(...)` calls the function and discards the
result. What changes is what happens to the return value, not whether the call is
made.

Full run (32957328130) surfaces 65 errors across 32 CH sources. Nothing in
`sources/europe/no/`.

PR #3047, opened 2026-09-03
FlyNorth images to use in Colombia

angoca:
I added new orthophotos to the index that can be used in Colombia.


arch0345:

PR #3052, opened 2026-09-04
Add Port Coquitlam 2025 imagery

maxwellward:
This layer includes cadastral boundaries and addresses as well.


andrewharvey (changes requested):
Thanks for the PR.

https://osmfoundation.org/wiki/OGL_Canada_and_local_variants lists this specific OGL license so the licensing looks OK.

Given https://maps.portcoquitlam.ca/server/rest/services/Basemap_Ortho2025_Legal/MapServer breaks down into the ortho mosaic and the cadastre, should this source be split up into one for the ortho, and then an overlay for the cadastre, and another overlay for the addresses?

That way mappers can opt to use the cadastre or addresses with other imagery providers if they choose.

maxwellward:
Good call! I wasn't aware overlays came from the ELI too, good to learn something new :) I think I did it right with the overlay key, but please do correct me if I'm wrong.

andrewharvey (approved):
Looks good.

PR #3054, opened 2026-09-05
Latvia - Remove LVM layers (unavailable)

### LINKED ISSUE #3051: Remove Latvias LVM GEO imagery
LVM had a ransomware incident that brought down the imagery back in june and they arent planning on bringing the public WMS services back.

Can read some of the background info here:
https://osmlatvija.github.io/zulip-archive/stream/358602-general/topic/LVM.20Jun.202026.20outage.html#613920084

I think ESRI is a good candidate for the best/default imagery to replace it, Bing maybe but they only have high res in Rīga and a few other cities.

danieldegroot2:
Remove lvm layers. (unavailable)

Fixes #3051 .


danieldegroot2:
Redundant as of https://github.com/osmlab/editor-layer-index/issues/3051#issuecomment-5551494078

PR #3056, opened 2026-09-07
Update Fairfax imagery URL

laurelmay:
It looks like rather than providing the current year at a non-stable
URL, it is now provided at a stable URL; however, with only the current
year available. It also seems that only zoom levels 10 through 20 are
actually available (though, I am not sure whether I verified that
correctly).

### DIFF
--- a/sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
+++ b/sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
@@ -1,23 +1,23 @@
 {
     "type": "Feature",
     "properties": {
-        "id": "Fairfax_VA_2025",
+        "id": "Fairfax_VA_Most_Recent",
         "attribution": {
             "url": "https://www.fairfaxcounty.gov/maps/aerial-photography",
             "text": "Government of Fairfax County, Virginia",
             "required": true
         },
-        "name": "Fairfax County Orthoimagery (2025)",
-        "url": "https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/2025AerialPhotographyCached/ImageServer/tile/{zoom}/{y}/{x}",
-        "min_zoom": 9,
-        "max_zoom": 23,
+        "name": "Fairfax County Orthoimagery (Current)",
+        "url": "https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/MostRecentAerialPhotographyCached/ImageServer/tile/{zoom}/{y}/{x}",
+        "min_zoom": 10,
+        "max_zoom": 20,
         "license_url": "https://www.fairfaxcounty.gov/maps/disclaimer",
         "country_code": "US",
         "type": "tms",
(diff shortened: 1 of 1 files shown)


sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
```diff
@@ -1,23 +1,23 @@
...
         "attribution": {
             "url": "https://www.fairfaxcounty.gov/maps/aerial-photography",
             "text": "Government of Fairfax County, Virginia",
             "required": true
         },
-        "name": "Fairfax County Orthoimagery (2025)",
-        "url": "https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/2025AerialPhotographyCached/ImageServer/tile/{zoom}/{y}/{x}",
-        "min_zoom": 9,
-        "max_zoom": 23,
+        "name": "Fairfax County Orthoimagery (Current)",
```
cicku: Suggestion: "Latest"

sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
```diff
@@ -1,23 +1,23 @@
...
+        "url": "https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/MostRecentAerialPhotographyCached/ImageServer/tile/{zoom}/{y}/{x}",
+        "min_zoom": 10,
+        "max_zoom": 20,
         "license_url": "https://www.fairfaxcounty.gov/maps/disclaimer",
         "country_code": "US",
         "type": "tms",
-        "start_date": "2025",
-        "end_date": "2025",
+        "start_date": "2026",
```
cicku, line 17: Are you sure the imagery here was captured in earlier this year?

cicku (changes requested):

sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
laurelmay: Based on https://www.fairfaxcounty.gov/gisimagery/rest/services/AerialPhotography/MostRecentAerialPhotographyCached/ImageServer and the typical practice from how Fairfax County has published these in the past, I believe so. Generally based on knowledge of the local area, that seems to be the case.

Other sources:
- https://www.fairfaxcounty.gov/maps/2026-aerial-photography-now-available
- https://www.fairfaxcounty.gov/informationtechnology/sites/informationtechnology/files/assets/itplan/2026-adopted/FY-2026-IT-Plan.pdf indicates images are taken between Dec-Mar

sources/north-america/us/va/Fairfax_VA_Most_Recent.geojson
laurelmay:
```suggestion
        "name": "Fairfax County Orthoimagery (Latest)",
```

cicku:
Thanks for the update, closing.

PR #3062, opened 2026-09-19
Add additional years of Fairfax County imagery

laurelmay:
This adds the last few years of imagery for Fairfax County.

If using the `MostRecentAerialPhotographyCached` still is preferable,
let me know how that should be handled to work with the changes for
grouping years from openstreetmap/iD#12561.


cicku:
Only 3~5 past years are needed. Fairfax has been developed rapidly in the past 30 years - dated imageries can lead to vandalism.

laurelmay:
> Only 3~5 past years are needed. Fairfax has been developed rapidly in the past 30 years - dated imageries can lead to vandalism.

Thanks for the feedback. I kept 2022-2026.

